PR #69, opened 2022-05-13
Add profiling code and caching for metrics

Dref360:
## Description:

Add code to profile different Modules. Work similarly as startup tasks, but run our profiling tool.

By doing so, I was able to find that loading metrics would cause a large bottleneck. So I added them to the artifact manager.

## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/model_performance/metrics.py
+++ b/azimuth/modules/model_performance/metrics.py
@@ -30,7 +30,6 @@ from azimuth.types.outcomes import ALL_OUTCOMES
 from azimuth.types.tag import ALL_DATA_ACTION_FILTERS, ALL_SMART_TAG_FILTERS
 from azimuth.utils.ml.ece import compute_ece_from_bins
 from azimuth.utils.ml.model_performance import sorted_by_utterance_count_with_last
-from azimuth.utils.object_loader import load_custom_object
 from azimuth.utils.validation import assert_not_none
 
 MAX_PRED = 3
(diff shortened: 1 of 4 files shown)


azimuth/modules/model_performance/metrics.py
```diff
@@ -78,8 +77,9 @@ def compute_on_dataset_split(self) -> List[MetricsModuleResponse]:  # type: igno
         metric_values = {}
         dm = self.get_dataset_split_manager()
         for metric_name, metric_obj_def in self.config.metrics.items():
-            met: Metric = load_custom_object(
-                metric_obj_def,
+            met: Metric = self.artifact_manager.get_metric(
```
gabegma, line 80: For other get functions from `ArtifactManager`, we have a helper function in `Module` - should we do the same here?

gabegma (approved):
This is great, I love it!

azimuth/modules/model_performance/metrics.py
Dref360: We don't have such a function for `get_tokenizer`. I'm not against adding it, but I thought that we added `get_dataset_split` and `get_model` because we use them everywhere vs tokenizers and metrics that we use only sparsely.

gabegma: That's true! Works with me.

PR #72, opened 2022-05-17
Use typing to force parsing all query parameters

JosephMarinier:
## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/utils/helpers.ts
+++ b/webapp/src/utils/helpers.ts
@@ -1,3 +1,4 @@
+import _ from "lodash";
 import { toast } from "react-toastify";
 import {
   QueryFilterState,
@@ -5,26 +6,6 @@ import {
   QueryPipelineState,
   QueryState,
 } from "types/models";
-import {
-  PAGE,
-  SORT,
-  DESCENDING,
-  PREDICTIONS,
-  LABELS,
-  SMART_TAGS,
-  DATA_ACTIONS,
-  CONFIDENCE_MIN,
-  CONFIDENCE_MAX,
-  OUTCOMES,
-  UTTERANCE,
-  PIPELINE_INDEX,
-} from "utils/const";
-import {
-  DataAction,
-  Outcome,
-  SmartTag,
-  UtterancesSortableColumn,
-} from "types/api";
 
 export const classNames = (...args: any[]) => args.filter(Boolean).join(" ");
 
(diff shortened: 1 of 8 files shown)


webapp/src/utils/helpers.ts
```diff
@@ -36,36 +17,45 @@ export const raiseSuccessToast = (message: string) => {
...
+  s?.split(",") as T[] | undefined;
 
-export const convertSearchParamsToPipelineState = (
-  q: URLSearchParams
-): QueryPipelineState => ({
-  pipelineIndex: str2num(q.get(PIPELINE_INDEX)),
-});
+const convertSearchParams = <T>(
+  q: URLSearchParams,
+  conversions: Required<{ [Key in keyof T]: (key: string | null) => T[Key] }>
```
JosephMarinier, line 31: That's the important bit: The `Required` ensures that a conversion function is defined for all query parameters defined in the different `Query...State` types, even if they have a `?:`.

webapp/src/utils/helpers.ts
```diff
@@ -36,36 +17,45 @@ export const raiseSuccessToast = (message: string) => {
...
+  (s as T) || undefined;
 
-export const convertSearchParamsToPaginationState = (
-  q: URLSearchParams
-): QueryPaginationState => ({
-  page: str2num(q.get(PAGE)),
-  sort: (q.get(SORT) || undefined) as UtterancesSortableColumn | undefined,
-  descending: q.get(DESCENDING) !== null || undefined,
-});
+const convertStringArray = <T extends string>(s: string | null) =>
```
JosephMarinier, line 26: The `T` is to seamlessly support string literals.

webapp/src/utils/helpers.ts
```diff
@@ -36,36 +17,45 @@ export const raiseSuccessToast = (message: string) => {
...
+      predictions: convertStringArray,
+      smartTags: convertStringArray,
+      dataActions: convertStringArray,
+      outcomes: convertStringArray,
+      utterance: convertString,
+    }),
+    pagination: convertSearchParams<QueryPaginationState>(q, {
+      page: convertNumber,
+      sort: convertString,
+      descending: (s) => s !== null || undefined,
```
christyler3030, line 52: Is it necessary to support `undefined` here?

christyler3030 (approved):
This seems very smart. Typescript rules.

webapp/src/utils/helpers.ts
JosephMarinier: There would be other ways, but this one is very simple. Since the default is `false` in the back end, we only support `true | undefined` (It's typed `descending?: true`). If you land on the page, you don't have any query parameters, it's clean, and you get all the default values, in this case `false`. If you click an arrow to change that, you `?descending=true`, and if you click again, you get back to no query parameters (it is omitted if `undefined`).
I know it's not your question, but side note so you understand what was my motivation: if we were to support `true | false | undefined` (`descending?: boolean`), you could end up with `?descending=false`, triggering a new call to the back end, even though you will get the same result as the cached initial call with no query parameters, since `false` is the default.
Makes sense?

christyler3030: Yes makes sense. Ship it!

PR #73, opened 2022-05-18
Nandhini/toggle post processing

nandhinibsn:
## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**


nandhinibsn:
differences on enable/disable of Postprocessing in confidence histogram, metrics, topwords.
Disabled postprocessing:
image
Enabled postprocessing:
image

differences on enable/disable of Postprocessing in confusion matrix

Disabled postprocessing:
image
Enabled postprocessing:
image
differences on enable/disable of Postprocessing in utterance table
Disabled postprocessing:
image

Enabled postprocessing:
image

JosephMarinier:
You would need to `Edit` the PR to request against `dev`, and not `main`.

azimuth/routers/v1/model_performance/outcome_count.py
```diff
@@ -78,10 +78,12 @@ def get_outcome_count_per_filter(
     task_manager: TaskManager = Depends(get_task_manager),
     dataset_split_manager: DatasetSplitManager = Depends(get_dataset_split_manager),
     pipeline_index: int = Depends(require_pipeline_index),
+    without_postprocessing: bool = False,
```
JosephMarinier, line 81: Oups, you missed this one, that's why the distribution bars in the control panel don't change, as the `withoutPostprocessing` query parameter is ignored.
```suggestion
    without_postprocessing: bool = Query(
        False, title="Without Postprocessing", alias="withoutPostprocessing"
    ),
```

JosephMarinier: Ok, I did it in 7049af2e5bbc0d197431c467d7ef69ec6eaf847f

webapp/src/components/Controls/Controls.tsx
```diff
@@ -70,11 +73,13 @@ const Controls: React.FC<Props> = ({
           datasetSplitName,
           ...filters,
           ...pipeline,
+          ...withoutPostprocessing,
         })
       : getUtteranceCountPerFilterEndpoint.useQuery({
           jobId,
           datasetSplitName,
           ...filters,
+          ...withoutPostprocessing,
```
JosephMarinier, line 82: This route doesn't support `withoutPostprocessing`, you can remove it:
```suggestion
```

webapp/src/components/Controls/Controls.tsx
```diff
@@ -112,11 +117,12 @@ const Controls: React.FC<Props> = ({
         ...filters,
         ...pagination,
         ...pipeline,
+        ...withoutPostprocessing,
```
JosephMarinier, line 120: The same should be added inside `handleClearFilters`, to not loose the state of the switch.

webapp/src/components/Controls/SwitchToggler.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const handleSwitchChange = (checked: boolean) => {
+    onChange(checked);
+  };
+  return (
+    <FormGroup aria-label="position" row>
+      <FormControlLabel
+        control={
+          <Switch
+            checked={enable ?? false}
+            onChange={(_, checked) => handleSwitchChange(checked)}
```
JosephMarinier, line 28: Is there something missing or you can call `onChange` directly?
```suggestion
            onChange={(_, checked) => onChange(checked)}
```

webapp/src/pages/UtteranceDetail.tsx
```diff
@@ -27,6 +27,7 @@ import {
   getUtterancesEndpoint,
 } from "services/api";
 import { DatasetSplitName, Outcome } from "types/api";
+import { QueryPostProcessingState } from "types/models";
```
JosephMarinier, line 30: On Teams, we concluded that we should not be changing the utterance detail page. I think you should revert all your changes to that page. In any case, this is not used and results in a warning.
```suggestion
```

nandhinibsn: thats right!.. I guess my rebase added this again. will remove all changes from utterance detail

webapp/src/components/Controls/SwitchToggler.tsx
nandhinibsn: can directly pass to onchange. I did not think about it. thats a good approach

webapp/src/components/Controls/Controls.tsx
nandhinibsn: right!

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -97,6 +98,7 @@ type Props = {
   filters: QueryFilterState;
   pagination: QueryPaginationState;
   pipeline: QueryPipelineState;
+  withoutPostprocessing: QueryPostProcessingState;
```
JosephMarinier, line 101: I think we should name the "family" of query parameters `postprocessing`, matching its type, similarly to how `pagination` is of type `QueryPipelineState`.
Also, for the type, to match the rest of our code base, I think we should stay consistent with `Postprocessing` in one word (no capital in the middle).
```suggestion
  postprocessing: QueryPostprocessingState;
```

webapp/src/components/Controls/Controls.tsx
nandhinibsn: would it be good to place the toggle below the filter section? as displaying toggle above clear filters convey it is out of filter
image

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: More details: That would avoid confusions like you had yesterday about `withoutPostprocessing.withoutPostprocessing` looking ridiculous, now being `postprocessing.withoutPostprocessing`.
Also, we can imagine other postprocessing-related query parameters joining this family, for example a `postprocessing.postprocessingType` that switches between different types. A `withoutPostprocessing.postprocessingType` wouldn't make sense.

webapp/src/components/Controls/Controls.tsx
nandhinibsn: same like rebase overwritten my commit.

webapp/src/components/Controls/Controls.tsx
JosephMarinier: I think it should not be affected by the `Clear filters` button, as it doesn't sound like a "filter" to me, and therefore it is fine over the `Filters` section where you had it. @gabegma, would you like to confirm?

azimuth/routers/v1/model_performance/outcome_count.py
nandhinibsn: oops! I added myself in the code..

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: makes sense

webapp/src/components/Controls/Controls.tsx
JosephMarinier: I see you just pushed the change you proposed, but I don't think it's good.

webapp/src/components/Controls/Controls.tsx
```diff
@@ -101,7 +106,11 @@ const Controls: React.FC<Props> = ({
 
   const handleClearFilters = () => {
     history.push(
-      `${baseUrl}${constructSearchString({ ...pagination, ...pipeline })}`
+      `${baseUrl}${constructSearchString({
+        ...pagination,
+        ...pipeline,
+        withoutPostprocessing: undefined,
```
JosephMarinier, line 112: This line doesn't do anything. Anyway, what I proposed was
```suggestion
        withoutPostprocessing,
```

nandhinibsn: It leaves me an error if I just give it like this     history.push(
      `${baseUrl}${constructSearchString({
        ...pagination,
        ...pipeline,
        withoutPostprocessing,
      })}`
    );

JosephMarinier: Sorryyy! I was confused between the object `withoutPostprocessing` and the boolean `withoutPostprocessing`. Another evidence that we should rename the object! 😉 
I meant:
```suggestion
        ...withoutPostprocessing,
```

nandhinibsn: ye true!.. working on that already

nandhinibsn: I tried options to just add "withoutPostprocessing" to the above line of code but It stops me with an error "TS18004: No value exists in scope for the shorthand property 'withoutPostprocessing'. Either declare one or provide an initializer." any help/suggestion on this would be very much helpful

webapp/src/components/Controls/Controls.tsx
JosephMarinier: I suggested it should NOT be set to `undefined` on clear filters [as I suggestion to add `withoutPostprocessing` (not `withoutPostprocessing: undefined`) in the handle clear filters].

webapp/src/components/Controls/Controls.tsx
JosephMarinier: Now that you renamed `withoutPostprocessing` to `postprocessing`, it should be
```suggestion
        ...postprocessing,
```

webapp/src/components/Controls/SwitchToggler.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <FormGroup aria-label="position" row>
+      <FormControlLabel
+        control={
+          <Switch
+            checked={enable ?? false}
+            onChange={(_, checked) => onChange(checked)}
+          />
+        }
+        label={label}
+        labelPlacement={labelPlacement ?? undefined}
```
JosephMarinier, line 29: The ` ?? undefined` won't do anything.
```suggestion
        labelPlacement={labelPlacement}
```

webapp/src/components/Controls/SwitchToggler.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const SwitchToggler: React.FC<Props> = ({
+  label,
+  labelPlacement,
+  enable,
+  onChange,
+}) => {
+  return (
+    <FormGroup aria-label="position" row>
```
JosephMarinier, line 20: I don't think you need a `FormGroup` here, as it contains only one element.
```suggestion
```
Also, general comment for next time: the `aria-label` is an accessibility feature, to help people that can't see the screen. The `position` doesn't mean anything here.

webapp/src/components/Controls/Controls.tsx
nandhinibsn: still clear filters doest reset with it

webapp/src/components/Controls/SwitchToggler.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { FormGroup, FormControlLabel, Switch } from "@mui/material";
+
+type LabelPlacement = "top" | "bottom" | "start" | "end";
```
JosephMarinier, line 4: Just so you know, this type already exists, it's `FormControlLabelProps["labelPlacement"]`. However, the default value is `"end"`, and we are happy with it, so I would suggest removing all this `labelPlacement` logic, and relying on the default MUI behavior.

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Thank you for renaming the filter family.
Now, what about the second part of the suggestion?
> Also, for the type, to match the rest of our code base, I think we should stay consistent with `Postprocessing` in one word (no capital in the middle).

webapp/src/components/Controls/SwitchToggler.tsx
nandhinibsn: Okay.. Make sense..

webapp/src/components/Controls/Controls.tsx
JosephMarinier: I tried and it works as expected.

webapp/src/components/Controls/SwitchToggler.tsx
JosephMarinier: What about?
> removing all this labelPlacement logic, and relying on the default MUI behavior.

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Plus there are 2 or 3 other places where we see `PostProcessing`, that should change to `Postprocessing`.

webapp/src/components/Controls/Controls.tsx
gabegma: Sorry I did not see the mention - I agree with Joseph - it should be **above** the filters as it is not doing any filtering. It should not be affected by clear filters.

webapp/src/components/Controls/Controls.tsx
nandhinibsn: added the necessary changes

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Thank you for renaming this one. I think some more are left though.

webapp/src/components/Controls/Controls.tsx
```diff
@@ -129,6 +139,16 @@ const Controls: React.FC<Props> = ({
   const handleDatasetSplitChange = (name: DatasetSplitName) =>
     history.push(`/${jobId}/dataset_splits/${name}/${mainView}${searchString}`);
 
+  const handlePostProcessingToggle = (enable: boolean) =>
```
JosephMarinier, line 142: Here, one capital `P` left. Also, I would change `Toggle` to `Change` to make it uniform with the other handlers. It would also avoid any possibility of confusion with a "toggle" button, which is a different component, but that's not the main reason.
```suggestion
  const handlePostprocessingChange = (enable: boolean) =>
```

webapp/src/components/Controls/Controls.tsx
```diff
@@ -207,6 +227,13 @@ const Controls: React.FC<Props> = ({
               onChange={handleDatasetSplitChange}
             />
           </Box>
+          <Box margin={1}>
+            <SwitchToggler
+              label="Without PostProcessing"
+              enable={postprocessing.withoutPostprocessing}
+              onChange={handlePostProcessingToggle}
+            />
```
JosephMarinier, line 235: Final touch! The `SwitchToggler` component ended up being so simple that I doubt it is useful. I think I would define the `Switch` directly here. What do you think?
```suggestion
            <FormControlLabel
              control={
                <Switch
                  checked={postprocessing.withoutPostprocessing ?? false}
                  onChange={(_, checked) => handlePostprocessingChange(checked)}
                />
              }
              label={"Without PostProcessing"}
            />
```

webapp/src/components/Controls/SwitchToggler.tsx
nandhinibsn: I agree to leave the default MUI behaviour but my idea is to make the label placement dynamic.. I will let the defaultvalue to be set

webapp/src/components/Controls/Controls.tsx
nandhinibsn: yes purpose is to make it reusable  but its the only place now its used now

JosephMarinier (approved):
🚢 🇮🇹

PR #74, opened 2022-05-20
Polish postprocessing switch

JosephMarinier:
## Description:

* [Fix filter family deactivated style](https://github.com/ServiceNow/azimuth/pull/74/commits/0b115af27b6174f2f0a1bafecad27938b8941c80)
* [Rename enable to checked](https://github.com/ServiceNow/azimuth/pull/74/commits/4dddfefd8848a8b8e80a6bd7631170630d945c6c)
* [Polish spacing](https://github.com/ServiceNow/azimuth/pull/74/commits/54d1d388267a81230cc025fafbf94d992665248a)
* [Change switch label](https://github.com/ServiceNow/azimuth/pull/74/commits/7d966f094727ac81a39ba78c7a712b78d8ab3010) from "Without PostProcessing" to "Exclude post-processing"
* [Add tooltip](https://github.com/ServiceNow/azimuth/pull/74/commits/a4286b16ad19c4be8ea354db6c4a78df5e99de55)

| !hover | hover |
| --- | --- |
| image | image |

## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Controls/Controls.tsx
+++ b/webapp/src/components/Controls/Controls.tsx
@@ -8,6 +8,7 @@ import {
   InputAdornment,
   OutlinedInput,
   Switch,
+  Tooltip,
   Typography,
   useTheme,
 } from "@mui/material";
@@ -229,15 +230,19 @@ const Controls: React.FC<Props> = ({
             />
           </Box>
           <Box margin={1}>
-            <FormControlLabel
-              control={
-                <Switch
-                  checked={postprocessing.withoutPostprocessing ?? false}
-                  onChange={(_, checked) => handlePostprocessingChange(checked)}
-                />
-              }
-              label="Exclude post-processing"
-            />
+            <Tooltip title="Include or exclude post-processing in predictions and any derived output. This only affects the Exploration Space, and won't affect the smart tags.">
+              <FormControlLabel
+                control={
+                  <Switch
+                    checked={postprocessing.withoutPostprocessing ?? false}
+                    onChange={(_, checked) =>
+                      handlePostprocessingChange(checked)
+                    }
+                  />
+                }
+                label="Exclude post-processing"
+              />
+            </Tooltip>
           </Box>
           <Box display="flex" justifyContent="space-between" marginX={1}>
             <Box display="flex" alignItems="center" gap={1} whiteSpace="nowrap">
--- a/webapp/src/components/Controls/Controls.tsx
+++ b/webapp/src/components/Controls/Controls.tsx
@@ -140,13 +140,13 @@ const Controls: React.FC<Props> = ({
   const handleDatasetSplitChange = (name: DatasetSplitName) =>
     history.push(`/${jobId}/dataset_splits/${name}/${mainView}${searchString}`);
 
-  const handlePostprocessingChange = (enable: boolean) =>
+  const handlePostprocessingChange = (checked: boolean) =>
     history.push(
       `${baseUrl}${constructSearchString({
         ...filters,
         ...pagination,
         ...pipeline,
-        withoutPostprocessing: enable || undefined,
+        withoutPostprocessing: checked || undefined,
       })}`
     );
 
--- a/webapp/src/components/Controls/Controls.tsx
+++ b/webapp/src/components/Controls/Controls.tsx
@@ -230,7 +230,7 @@ const Controls: React.FC<Props> = ({
             />
           </Box>
           <Box margin={1}>
-            <Tooltip title="Include or exclude post-processing in predictions and any derived output. This only affects the Exploration Space, and won't affect the smart tags.">
+            <Tooltip title="Exclude post-processing in predictions and any derived output. This only affects the Exploration Space, and won't affect the smart tags.">
               <FormControlLabel
                 control={
                   <Switch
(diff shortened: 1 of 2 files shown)


gabegma (commented):
Looks great!! Can we add in the hover something like "This only affects the Exploration Space, and won't affect the smart tags."

webapp/src/components/Controls/Controls.tsx
```diff
@@ -228,24 +228,25 @@ const Controls: React.FC<Props> = ({
...
+              gridTemplateColumns="1fr auto 1fr"
+              alignItems="center"
+            >
+              <Typography align="right">Model Output</Typography>
+              <Switch
+                checked={postprocessing.withoutPostprocessing === undefined}
+                onChange={(_, checked) => handlePostprocessingChange(checked)}
+                sx={(theme) => ({
+                  // Force thumb colored all the time, not only when checked
+                  "& .MuiSwitch-thumb": { color: theme.palette.primary.main },
```
christyler3030, line 243: This made me do a bit of reading about Material Design and I learned that generally toggle switches are not intended to be used to switch between two options, rather they are for activating/deactivating a _single_ option. (see here: https://material.io/components/selection-controls#usage scroll to switches)

We can of course do whatever we like but let's make sure we are doing it consciously. Is it worth considering a different control for this interaction (e.g. maybe a toggle switch)

christyler3030 (approved):
Approved with design related comment.

webapp/src/components/Controls/Controls.tsx
JosephMarinier: Totally agree, I saw that. By the way, they are also linking to [this](https://uxplanet.org/checkbox-vs-toggle-switch-7fc6e83f10b8) interesting distinction between checkbox and toggle switch, which we don't respect in the settings page. One day!

But yeah, I just have a bad reason: being different than the dataset split toggle button group above it.

The other option to keep the switch looking like it "activates" the post-processing in the output (by inverting the current switch). We simply had a hard time phrasing it in a clear way. But the clarity could come from the tooltip. What do you think?

image

JosephMarinier: For example, "Include post-processing" (with the switch flipped and a clear tooltip)? @gabegma, what do you think?
image

nandhinibsn: I agree with both design look appropriate but second option looks more reasonable. I guess "Include post processing" is more explainable. with that name, we need to reverse the default logic right?

JosephMarinier: Right, I flipped the switch by inverting the boolean.

JosephMarinier: ...which results in the post-processing being on by default, when one lands on the page, as desired.

I now lean towards this simple solution, as seen in my latest screenshot. I tried implementing a little logic for disabling the switch (graying it out) when no pipeline is selected, as the switch becomes pointless, and my little hack of the styling (plus the two "labels" not being true `FormControlLabel`s as there can't my more than 1) became a nightmare.

@gabegma, what do you think? Is that last screenshot OK with a good tooltip?

gabegma: I think I would find it more natural to have the switch be 'exclude post-processing' and to have it disabled by default. But I'm ok with either if someone has a strong opinion.

JosephMarinier: Ho that I am fine! So in other words, I don't flip the switch, I only change "Without PostProcessing" for "Exclude post-processing"?

For the record, my motivation for flipping the switch was

1. Avoid the "double-negation" of turning off the exclusion of post-processing; and
2. Have the switch mirror the "chronological" order, left to right: model output before pipeline output (but that's less apparent anyway with only one label on one side).

gabegma: Yeah I get the double-negation and it kind of annoys me too, but given that it should be the 'exception' to exclude post-processing, I find it odd to talk about "including" it, when it should be included by default. I could be convinced otherwise. Curious to know what our designers would say too here.

webapp/src/components/Controls/Controls.tsx
```diff
@@ -229,23 +230,21 @@ const Controls: React.FC<Props> = ({
...
-            <FormControlLabel
-              control={
-                <Switch
-                  checked={postprocessing.withoutPostprocessing ?? false}
-                  onChange={(_, checked) => handlePostprocessingChange(checked)}
-                />
-              }
-              label="Without PostProcessing"
-            />
+            <Tooltip title="Include or exclude post-processing in predictions and any derived output. This only affects the Exploration Space, and won't affect the smart tags.">
```
christyler3030, line 233: Nitpick: I think maybe the text shouldn't say "Include" since this the action one takes with this switch is only "Exclude"

I think the text should read:
"Exclude post-processing in predictions and any derived output. This only affects the Exploration Space, and won't affect the smart tags.

christyler3030 (approved):
Re-approved with one tiny suggestion for text.

webapp/src/components/Controls/Controls.tsx
JosephMarinier: Makes sense to me! Thank you!
```suggestion
            <Tooltip title="Exclude post-processing in predictions and any derived output. This only affects the Exploration Space, and won't affect the smart tags.">
```

PR #86, opened 2022-06-01
Toggle normalized confusion matrix

### LINKED ISSUE #115: Show not-normalized confusion matrix
Add a toggle beside the confusion matrix

nandhinibsn:
## Description: Added a new feature to normalize/not normalize confusion matrix data

Fixes #115 

## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/ConfusionMatrix.tsx
+++ b/webapp/src/components/ConfusionMatrix.tsx
@@ -109,7 +109,7 @@ const ConfusionMatrix: React.FC<Props> = ({
   const {
     data: { confusionMatrix: data, normalized } = {
       confusionMatrix: [],
-      normalized: true,
+      normalized: confusionMatrix.normalized,
     },
   } = getConfusionMatrixEndpoint.useQuery({
     jobId,


gabegma (commented):
I can't review the code - but it should merged on dev, not on main :)

webapp/src/components/ConfusionMatrix.tsx
```diff
@@ -145,174 +169,187 @@ const ConfusionMatrix: React.FC<Props> = ({
             fontSize="0.75em"
             sx={{ verticalAlign: "0.25em" }}
           >
-            %
+            {normalizedState.normalized ?? `%`}
```
JosephMarinier, line 172: There is no point in rendering this `Typography` if it is empty. Also, we should not do `(value * 100).toFixed(0)` if `value` is not normalized (if it's not a percentage).
I now realize that switching normalization on and off will probably glitch if we use `normalizedState.normalized` that we requested and the data from the response... The view will change before we receive the data, then the data will change. We might temporarily see things like `12300%` where we expect `123`. I can think of a couple of solutions to that, but I think the simplest and more robust is to add the `normalized` field to the API response as you suggested, and rely on it to render the matrix (but not the switch).

webapp/src/utils/helpers.ts
```diff
@@ -58,6 +59,9 @@ export const parseSearchString = (searchString: string) => {
     postprocessing: convertSearchParams<QueryPostprocessingState>(q, {
       withoutPostprocessing: (s) => s !== null || undefined,
     }),
+    normalizedState: convertSearchParams<QueryNormalizedState>(q, {
```
JosephMarinier, line 62: I think I would call that the "confusion matrix state", as it could welcome more options. It also allows to remove the `State` from the variable name, making it uniform with the others, without causing a confusing `normalized.normalized`. Instead, `confusionMatrix.normalized` is quite clear.
Makes sense?
```suggestion
    confusionMatrix: convertSearchParams<QueryConfusionMatrixState>(q, {
```

webapp/src/components/ConfusionMatrix.tsx
```diff
@@ -145,174 +169,187 @@ const ConfusionMatrix: React.FC<Props> = ({
...
+        <FormControlLabel
+          control={
+            <Switch
+              checked={normalizedState.normalized ?? true}
+              onChange={(_, checked) => handleNormalizedStateChange(checked)}
+            />
+          }
+          label="Normalize"
+        />
+      </Box>
```
JosephMarinier, line 191: As discussed in Teams, we would like the toggle to stick right on top of the matrix. You can achieve that by simply moving this `Box` inside the `Box` opening on the following line. Also, you will be able to remove the React fragment (`<>...</>`) around it and unindent the whole return statement.

webapp/src/components/ConfusionMatrix.tsx
nandhinibsn: okay ... definitely I would like to explore on the API side solution and get back to you. pleas give me sometime today..

Also, why cant we conditionally render and evaluate the data in displaying for normalized/not normalized? May be I am not clear enought to understand when you say "The view will change before we receive the data, then the data will change. "

webapp/src/utils/helpers.ts
nandhinibsn: I always guess the name inappropriate :) I should be improving that part

webapp/src/components/ConfusionMatrix.tsx
nandhinibsn: Agreed!. I explored this option of keeping it closer to matrix but felt "placing it above right" is more common design. but I personally liked keeping it closer too..

webapp/src/utils/helpers.ts
JosephMarinier: Haha! It was not "inappropriate". Don't worry, that's why we do reviews.

nandhinibsn:
If normalized:
image
If not normalized:
image

nandhinibsn:
now, I read "normalized" from the API. Quick question, do we keep track of the history for "normalized" param?

azimuth/types/model_performance.py
```diff
@@ -87,7 +87,7 @@ class MetricsPerFilterAPIResponse(MetricsPerFilterModuleResponse):
 
 class ConfusionMatrixResponse(ModuleResponse):
     confusion_matrix: Array[float] = Field(..., title="Confusion Matrix normalized")
-
+    normalized: bool = Field(..., title="normalized state of Confusion Matrix")
```
JosephMarinier, line 90: Black (in the tests) fails because you removed that empty line.
```suggestion
    normalized: bool = Field(..., title="Normalized state of Confusion Matrix")

```

azimuth/routers/v1/model_performance/confusion_matrix.py
```diff
@@ -50,12 +50,12 @@ def get_confusion_matrix(
...
         SupportedModule.ConfusionMatrix,
         dataset_split_name,
         task_manager=task_manager,
         mod_options=mod_options,
         last_update=dataset_split_manager.last_update,
-    )[0]
+    )[0].confusion_matrix
 
-    return task_result
+    return ConfusionMatrixResponse(confusion_matrix=task_result, normalized=normalized)
```
JosephMarinier, line 61: I think this router shouldn't change since the `ConfusionMatrixResponse` module response is supposed to already have `normalized`, as per how you defined the type. So, I think `normalized` should be added there:
https://github.com/ServiceNow/azimuth/blob/f69febe750281b59831f286b4aabf3204c8bb69f/azimuth/modules/model_performance/confusion_matrix.py#L41

azimuth/types/model_performance.py
nandhinibsn: yes important! changes can be made

azimuth/routers/v1/model_performance/confusion_matrix.py
nandhinibsn: Actually, the task result is returning this -> confusion_matrix=array([[],[]]) so I had to write that way. will figure out more cleaner way of sending back the response data

webapp/src/utils/helpers.ts
JosephMarinier: I see you added `State` to the field name `confusionMatrix`. Could you remove it to keep it uniform with the fields?

azimuth/types/model_performance.py
JosephMarinier: The pre-commits should catch that.

azimuth/routers/v1/model_performance/confusion_matrix.py
JosephMarinier: No, it returns a `List[ConfusionMatrixResponse]` (hence the `[0]`) here. Mypy (included in the pre-commits, but you can also run `poetry run mypy azimuth`) should point you to where the `normalized` field is currently missing in the module.

nandhinibsn: okay.. yes thats right. I could notice on running pre-commit and as well as poetry run. But I added the "normalized" param already to the ConfusionMatrixResponse Model. strange why its missing there now

webapp/src/utils/helpers.ts
nandhinibsn: Right, the reason I added "state" to the suggested name because of the variable names conflict on the file
Here is the conflict
  const { data: { confusionMatrix, normalized } = { confusionMatrix: [] } } =
    getConfusionMatrixEndpoint.useQuery({
      jobId,
      datasetSplitName,
      ...filters,
      ...pipeline,
      ...postprocessing,
      ...confusionMatrixState,
    });

JosephMarinier: I would prefer renaming that variable so it only impact one file. Something like:
```diff
-const { data: { confusionMatrix, normalized } = { confusionMatrix: [] } } =
```

nandhinibsn: yes thats true. I agree

azimuth/routers/v1/model_performance/confusion_matrix.py
```diff
@@ -50,12 +50,12 @@ def get_confusion_matrix(
...
+    task_results: ConfusionMatrixResponse = get_standard_task_result(
         SupportedModule.ConfusionMatrix,
         dataset_split_name,
         task_manager=task_manager,
         mod_options=mod_options,
         last_update=dataset_split_manager.last_update,
     )[0]
 
-    return task_result
+    return task_results
```
JosephMarinier, line 61: Can you please revert those changes?

tests/test_modules/test_model_performance/test_confusion_matrix.py
```diff
@@ -56,3 +57,17 @@ def test_confusion_matrix(tiny_text_config):
     )
     [json_output_not_normalized] = mod_not_normalized.compute_on_dataset_split()
     assert json_output_not_normalized.confusion_matrix.sum() == dm.num_rows
+    assert [json_output_not_normalized] == [
+        ConfusionMatrixResponse(confusion_matrix=json_output_not_normalized, normalized=False)
+    ]
```
JosephMarinier, line 62: What are you trying to test here? It doesn't seem to work.

tests/test_modules/test_model_performance/test_confusion_matrix.py
```diff
@@ -56,3 +57,17 @@ def test_confusion_matrix(tiny_text_config):
...
+
+    # When normalized, we get the predictions.
+    mod_not_normalized = ConfusionMatrixModule(
```
JosephMarinier, line 67: I guess you didn't meant to call this "not normalized".
```suggestion
    mod_normalized = ConfusionMatrixModule(
```

nandhinibsn: that was test script to check the normalized attribute value.

azimuth/routers/v1/model_performance/confusion_matrix.py
nandhinibsn: I did... :)

webapp/src/components/ConfusionMatrix.tsx
```diff
@@ -83,28 +94,45 @@ const ConfusionMatrix: React.FC<Props> = ({
...
-      jobId,
-      datasetSplitName,
-      ...filters,
-      ...pipeline,
-      ...postprocessing,
-    });
+  const {
+    data: { confusionMatrix: data, normalized } = {
+      confusionMatrix: [],
+      normalized: true,
```
JosephMarinier, line 112: Interesting... We might want to better handle the loading and error states of this component in the future, but for now I think we should at least use our best guess as the default value: the query parameter.
```suggestion
      normalized: confusionMatrix.normalized,
```

JosephMarinier: Yes, it sets the default value of `normalized` if the API response did not came back yet.

webapp/src/components/ConfusionMatrix.tsx
JosephMarinier: Done in f1e586a0badc8ff89ccbffd65601c5f5f0ddd142

webapp/src/components/ConfusionMatrix.tsx
JosephMarinier: Done in e819d2ccc535a4828cce530e4bab399a9724df11

JosephMarinier (approved):

PR #108, opened 2022-06-03
Fix issue where dependency caching was not done properly in the backend

Dref360:
## Description:

Fix issue where dependency caching was not done properly in the backend

## Checklist:

it.

* [ ] **FRONTEND TYPES.** Regenerate the front-ent types if you played with types and routes.
  our users.
* [ ] **DEV CHANGES.**

### DIFF
--- a/.github/workflows/pythonci.yml
+++ b/.github/workflows/pythonci.yml
@@ -29,7 +29,6 @@ jobs:
       - run: poetry --version
       - name: Install dependencies
         run: |
-          poetry config virtualenvs.create false
           poetry install --no-interaction --no-ansi
       - name: Test with pytest/flake8/mypy
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonci.yml
```diff
@@ -29,7 +29,6 @@ jobs:
       - run: poetry --version
       - name: Install dependencies
         run: |
-          poetry config virtualenvs.create false
```
JosephMarinier, line 32: Can you explain how does that fix the issue?

Dref360: This line would install dependencies outside of poetry, ie in the system python folder.
So the cache step didn't trigger as ~/.cache/poetry was always empty. Now it trigger correctly based on the MD5 of poetry.lock

JosephMarinier (approved):

PR #145, opened 2022-06-14
Smart tag families

JosephMarinier:
## Description:

Create smart tag families to group similar smart tags.

Screen Shot 2022-06-14 at 8 16 10 AM

Screen Shot 2022-06-14 at 8 24 32 AM

Screen Shot 2022-06-14 at 8 16 47 AM

## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**


webapp/src/components/Controls/FilterSelector.tsx
```diff
@@ -151,10 +150,6 @@ const FilterSelector = <FilterValue extends string>({
               )}
               className={classes.checkbox}
               color="primary"
-              disabled={
-                operator === "AND" &&
-                (isFetching || filter.utteranceCount === 0)
-              }
```
JosephMarinier, line 157: The `"AND"` feature is no more! The operator is now uniform, always `OR` per family of filters. For example:
`(label A OR label B) AND (prediction A or prediction B) AND (extreme_length short_sentence OR extreme_length long_sentence) AND (uncertain high_epistemic_uncertainty)`

azimuth/types/model_performance.py
```diff
@@ -30,18 +34,46 @@ class UtteranceCountPerFilterValue(AliasModel):
...
+        Generic[T],
+        **{k.value: (List[T], Field(..., title=k.value)) for k in SMART_TAG_FAMILY_DATASET}
     )
-    label: List[UtteranceCountPerFilterValue] = Field(..., title="Label")
-    smart_tag: List[UtteranceCountPerFilterValue] = Field(..., title="Smart tag")
-    data_action: List[UtteranceCountPerFilterValue] = Field(..., title="Data action tag")
-    outcome: Optional[List[UtteranceCountPerFilterValue]] = Field(
-        ..., title="Outcome", nullable=True
+    ValuePerPipelineSmartTag = AliasModel.with_fields(
+        "ValuePerDatasetSmartTag",
```
JosephMarinier, line 56: Should that be?
```suggestion
        "ValuePerPipelineSmartTag",
```

azimuth/types/model_performance.py
```diff
@@ -30,18 +34,46 @@ class UtteranceCountPerFilterValue(AliasModel):
...
+
+    ValuePerDatasetSmartTag = GenericAliasModel
+    ValuePerPipelineSmartTag = GenericAliasModelPipeline
+
+else:
+    ValuePerDatasetSmartTag = AliasModel.with_fields(
+        "ValuePerDatasetSmartTag",
+        GenericModel,
+        Generic[T],
+        **{k.value: (List[T], Field(..., title=k.value)) for k in SMART_TAG_FAMILY_DATASET}
```
JosephMarinier, line 53: If the `title` is the same as the field name, is it really useful? Can we just omit the `Field` altogether? It's not important, I'm just curious.
```suggestion
        **{k.value: List[T] for k in SMART_TAG_FAMILY_DATASET}
```

azimuth/types/tag.py
```diff
@@ -56,6 +56,26 @@ class SmartTag(str, Enum):
...
+    behavioral_testing = "behavioral_testing"
+    uncertain = "uncertain"
+
+
+SMART_TAG_FAMILY_DATASET = [
+    SmartTagFamily.extreme_length,
+    SmartTagFamily.partial_syntax,
+    SmartTagFamily.dissimilar,
+]
+SMART_TAG_FAMILY_PIPELINE = [
```
JosephMarinier, line 73: Any objection to using plural "families" and putting dataset/pipeline as an adjective at the beginning, similarly to the types like `ValuePerPipelineSmartTag`?
```suggestion
DATASET_SMART_TAG_FAMILIES = [
    SmartTagFamily.extreme_length,
    SmartTagFamily.partial_syntax,
    SmartTagFamily.dissimilar,
]
PIPELINE_SMART_TAG_FAMILIES = [
```

azimuth/utils/pydantic_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Args:
+    __model_name: name of the created model
+    __config__: config class to use for the new model
+    __base__: base class for the new model to inherit from
+    __module__: module of the created model
+    __validators__: a dict of method names and @validator class methods
+    field_definitions: fields of the model (or extra fields if a base is supplied)
+        in the format `<name>=(<type>, <default default>)` or `<name>=<default value>, e.g.
+        `foobar=(str, ...)` or `foobar=123`, or, for complex use-cases, in the format
```
JosephMarinier, line 38: I guess that answers my question https://github.com/ServiceNow/azimuth/pull/145/files#r899309117: No, this doesn't support supplying the type but not the default value.

azimuth/utils/pydantic_utils.py
```diff
@@ -0,0 +1,86 @@
...
+                f'fields may not start with an underscore, ignoring "{f_name}"', RuntimeWarning
+            )
+        if isinstance(f_def, tuple):
+            try:
+                f_annotation, f_value = f_def
+            except ValueError as e:
+                raise ConfigError(
+                    "field definitions should either be a tuple of (<type>, <default>) or just a "
+                    "default value, unfortunately this means tuples as "
+                    "default values are not allowed"
```
JosephMarinier, line 68: But you could specify a type and a tuple as a default value, right? `potato: (Tuple, (1, 2))`? So that sentence is not completely true.

azimuth/types/model_performance.py
```diff
@@ -30,18 +34,46 @@ class UtteranceCountPerFilterValue(AliasModel):
...
+if typing.TYPE_CHECKING:
+
+    class GenericAliasModel(AliasModel, Generic[T]):
+        pass
+
+    class GenericAliasModelPipeline(AliasModel, Generic[T]):
+        pass
+
+    ValuePerDatasetSmartTag = GenericAliasModel
+    ValuePerPipelineSmartTag = GenericAliasModelPipeline
```
JosephMarinier, line 46: Was there a problem when using the same class twice?
```suggestion
    ValuePerPipelineSmartTag = GenericAliasModel
```

azimuth/types/model_performance.py
```diff
@@ -30,18 +34,46 @@ class UtteranceCountPerFilterValue(AliasModel):
...
+
+else:
+    ValuePerDatasetSmartTag = AliasModel.with_fields(
```
JosephMarinier, line 49: But bro, that's pretty cool stuff! I'm happy that the dataset vs pipeline separation can now be defined in a single constant. 👍

azimuth/routers/v1/utterances.py
```diff
@@ -212,23 +212,32 @@ def get_utterances(
...
-        else set(ALL_SMART_TAGS).difference(ALL_PREDICTION_TAGS)
-    )
+    if pipeline_index is not None:
+        available_families = list(SMART_TAGS_FAMILY_MAPPING.keys())
+    else:
+        available_families = [
+            SmartTagFamily.extreme_length,
+            SmartTagFamily.partial_syntax,
+            SmartTagFamily.dissimilar,
+        ]
```
JosephMarinier, line 222: Can we reuse our new constants here?
```suggestion
    available_families = SMART_TAG_FAMILY_DATASET
    if pipeline_index is not None:
        available_families += SMART_TAG_FAMILY_PIPELINE
```

azimuth/types/model_performance.py
Dref360: Yes, Utterance inherits from both so they have to be different types.

webapp/src/styles/theme.tsx
```diff
@@ -151,6 +151,13 @@ const customTheme: Theme = createTheme({
         },
       },
     },
+    MuiTooltip: {
```
christyler3030, line 154: Just checking: are we sure this should be at the theme level?

webapp/src/utils/const.ts
```diff
@@ -37,3 +46,35 @@ export const OUTCOME_PRETTY_NAMES = {
...
+
+export const SMART_TAG_FAMILIES = [
+  ...DATASET_SMART_TAG_FAMILIES,
+  "almostCorrect",
+  "behavioralTesting",
+  "uncertain",
+] as const;
+
+export const SMART_TAG_FAMILY_ICONS = {
+  extremeLength: ExtremeLength,
```
christyler3030, line 65: Nitpick: do you want to type these to be sure they have the right keys? Same for below with pretty names

webapp/src/components/Controls/Controls.tsx
```diff
@@ -336,17 +341,28 @@ const Controls: React.FC<Props> = ({
...
-                label="Smart Tags"
-                maxCount={maxCount}
-                operator="AND"
-                searchValue={searchValue}
-                selectedOptions={selectedSmartTags}
-                handleValueChange={handleFilterSelectorChange("smartTags")}
-                filters={countPerFilter?.countPerFilter.smartTag}
-                isFetching={isFetchingCountPerFilter}
-              />
+              {SMART_TAG_FAMILIES.map((filterName) => (
```
christyler3030, line 344: Shouldn't this depend on whether there is a pipeline?

azimuth/types/utterance.py
```diff
@@ -25,13 +29,12 @@ class ModelSaliency(AliasModel):
...
+class Utterance(ValuePerDatasetSmartTag[str], ValuePerPipelineSmartTag[str]):
     index: int = Field(..., title="Index", description="Row index computed by Azimuth..")
     model_prediction: Optional[ModelPrediction] = Field(
         ..., title="Model prediction", nullable=True
     )
     model_saliency: Optional[ModelSaliency] = Field(..., title="Model saliency", nullable=True)
-    smart_tags: List[SmartTag] = Field(..., title="Smart tags")
     data_action: DataAction = Field(..., title="Data action tag")
     label: str = Field(..., title="Label")
     utterance: str = Field(..., title="Utterance")
```
christyler3030, line 40: Not related to this PR but: Given that this type is already an Utterance i think this property might be better off as `text` or something

webapp/src/utils/const.ts
JosephMarinier: Haha, I think it's the 3rd time you ask me this question in this particular file. I'm starting to think I am wrong. Basically, I like it with `as const` as we can hover `SMART_TAG_FAMILY_ICONS` in our IDE and see the individual values. Then, the different places were we use the constant will take care of checkign the type. For example, if we loop over `SMART_TAG_FAMILIES` and try to index `SMART_TAG_FAMILY_ICONS` with it, the keys will have to match for typescript to compile.
It's even more true with arrays like `SMART_TAG_FAMILIES `, as the order is also constant in the type and shown in the IDE tooltip. Neat!
Finally, for this particular case of the smart tag families, there is currently no generated type with those values, so I'd have to "hardcode" the list anyway.
Does that make sense to you?

webapp/src/pages/UtteranceDetail.tsx
```diff
@@ -194,15 +203,17 @@ const UtteranceDetail = () => {
...
-              variant="outlined"
-              size="small"
-              key={tag}
-              label={tag}
-            />
-          ))}
+          {smartTagFamilies.map(
+            (family) =>
+              utterance[family].length > 0 && (
+                <SmartTagFamilyBadge
```
JosephMarinier, line 209: @gabegma 

> As such, what I had in mind is to have no blue circles, and to display the name of the smart tag value instead of the name of the smart tag family. Hence, an icon would be there multiple times if more than one value is available within a family. We should discuss this, I'm curious what you think and our designers. Also, this could totally be done in a subsequent PR.
> 
> Screen Shot 2022-06-20 at 10 06 18 AM

That's what our designers mocked up. I was concerned that if we implement some more dynamic tags (tags that would depends on the number of classes for example), the list might become very long. I agree we should discuss it. For now, I simply resued my new component from the utterances table, with the addition of the family name.

azimuth/types/utterance.py
```diff
@@ -25,13 +29,12 @@ class ModelSaliency(AliasModel):
...
 
 
-class Utterance(AliasModel):
+class Utterance(ValuePerDatasetSmartTag[str], ValuePerPipelineSmartTag[str]):
     index: int = Field(..., title="Index", description="Row index computed by Azimuth..")
     model_prediction: Optional[ModelPrediction] = Field(
         ..., title="Model prediction", nullable=True
     )
     model_saliency: Optional[ModelSaliency] = Field(..., title="Model saliency", nullable=True)
-    smart_tags: List[SmartTag] = Field(..., title="Smart tags")
```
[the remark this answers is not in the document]
JosephMarinier: I don't think it makes a real difference in the front end. I personally changed the responses of metrics/outcome/utterance counts per filter to be flat (not nested under `smart_tags`) to greatly simply the generic types. I don't really mind about this utterance response, but I thought it was fine like that, more uniform with the responses of counts per filter. Is it OK with you?

azimuth/types/utterance.py
JosephMarinier: Since the column name (in the database as well as in the front end) is "utterance", I think it should remain. Could we instead change the type's name to something like `UtteranceDetails`?

JosephMarinier: Of course. But I like it! 👍 Thanks for pointing it out!

azimuth/types/utterance.py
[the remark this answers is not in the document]
christyler3030: If you have a strong preference for this then go ahead but one of the reasons I would rather group them together is for places like here:
https://github.com/ServiceNow/azimuth/blob/4060e1b4a897a6268a659bb132bc073fecf07ed3/webapp/src/components/SmartTagFamilyBadge.tsx#L14

A component that shows a list of smart tags shouldn't need to know about the rest of the utterance.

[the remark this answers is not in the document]
JosephMarinier: No, as I said, I don't really mind about this utterance response, and I like that argument! I'll try the change. Thanks!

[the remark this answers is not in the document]
JosephMarinier: Actually, `SmartTagFamilyBadge` doesn't need to know about all the smart tag families, but only one family, so it solves the problem: d293ef8c014c4d892c3af4b194bbf3c8b23d3f02

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -41,13 +53,15 @@ const ColumnMenu = ({ hideMenu, currentColumn, open }: GridColumnMenuProps) => (
   </GridColumnMenuContainer>
 );
 
-const OPTIONS = ["label", "prediction", "smartTag"] as const;
+const BASIC_FILTER_OPTIONS = ["label", "prediction"] as const;
```
JosephMarinier, line 56: Sorry @christyler3030, I added a commit, since it was ready: d4fb590afb68491aa548a84ee442ba85342a1e13
Here I separate `BASIC_FILTER_OPTIONS` and `SMART_TAG_FAMILIES`, which are rendered a bit differently, with the family icon.

christyler3030 (approved):

PR #146, opened 2022-06-14
Description in UI

nandhinibsn:
## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**


nandhinibsn:
Added a test cases that covered all the section descriptions and documentation link each one points to. 
image

webapp/src/components/Analysis/PreviewCard.tsx
```diff
@@ -1,25 +1,43 @@
...
+import Description from "components/Description";
 import React from "react";
-import { Link } from "react-router-dom";
+import { Link as RouterLink } from "react-router-dom";
 
 type Props = {
   title: string;
+  subtitle?: string;
   to?: string;
+  href?: string;
```
JosephMarinier, line 10: Any reason for making `subtitle` and `href` optional? I think they should be required.
Also, I think we could avoid repeating and passing the props of `Description` by passing a `Description` element directly. Something like
```suggestion
  to?: string;
  description: React.ReactElement<typeof Description>;
```
and
```diff
-const PreviewCard: React.FC<Props> = ({
-  title,
-  subtitle,
-  to,
-  href,
-  children,
-}) => (
+const PreviewCard: React.FC<Props> = ({ title, to, description, children }) => (
````
and
```diff
-        {subtitle && href && <Description text={subtitle} link={href} />}
+        {description}
````
and
```diff
-          subtitle="Compare the class distribution of your training and evaluation sets. "
-          href="/dataset-warnings/"
           to={`/${jobId}/dataset_class_distribution_analysis${searchString}`}
+          description={
+            <Description
+              text="Compare the class distribution of your training and evaluation sets. "
+              link="/dataset-warnings/"
+            />
+          }
```

webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
```diff
@@ -4,6 +4,22 @@ import MetricsDataGrid from "components/DatasetWarnings/MetricsDataGrid";
 import { ResponsivePlotWrapper, WarningPlot } from "components/PlotWrapper";
 import React from "react";
 import { DatasetWarningGroup } from "types/api";
+import Description from "components/Description";
+
+interface DescriptionMap {
+  [key: string]: string;
+}
+
+const WARNING_GROUP_DESCRIPTION: DescriptionMap = {
```
JosephMarinier, line 13: I think you should use `Record<string, string>` instead of creating this vague `DescriptionMap`.
```suggestion
const WARNING_GROUP_DESCRIPTION: Record<string, string> = {
```

webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
```diff
@@ -4,6 +4,22 @@ import MetricsDataGrid from "components/DatasetWarnings/MetricsDataGrid";
...
+  GeneralWarnings: "See class distributions differences. ",
+  SyntacticWarnings:
+    "Review the arrangement of utterances between your training and testing data. ",
```
JosephMarinier, line 16: I would add the spaces in the keys here
```suggestion
  "General Warnings": "See class distributions differences. ",
  "Syntactic Warnings":
    "Review the arrangement of utterances between your training and testing data. ",
```
to avoid the `.replace(/\s/g, "")`
```diff
                 <Description
-                  text={
-                    WARNING_GROUP_DESCRIPTION[
-                      warningGroup.name.replace(/\s/g, "")
-                    ]
-                  }
-                  link={
-                    WARNING_GROUP_DOC_LINK[warningGroup.name.replace(/\s/g, "")]
-                  }
+                  text={WARNING_GROUP_DESCRIPTION[warningGroup.name]}
+                  link={WARNING_GROUP_DOC_LINK[warningGroup.name]}
                 />
```

webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
```diff
@@ -4,6 +4,22 @@ import MetricsDataGrid from "components/DatasetWarnings/MetricsDataGrid";
...
+} as const;
```
JosephMarinier, line 17: Since you specify a type `: DescriptionMap` (or `: Record<string, string>`), the `as const` doesn't do anything.
```suggestion
};
```

nandhinibsn: okay I understood about const assertion in a wrong way. but this article gave a good explanation on the usage https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case.

webapp/src/components/Analysis/PreviewCard.tsx
nandhinibsn: I agree this is a good way to avoid passing props everytime..

JosephMarinier: > so is this is the way to pass props to components where there are only optional param is present?

I'm not sure I get the question. A `?` makes a prop optional. Here, I think all `PreviewCard`s should require a description, so it should not be optional. Does that answer your question?

webapp/src/components/Description.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+type Props = {
+  text?: string;
+  link?: string;
+};
+
+const Description: React.FC<Props> = ({ text, link }) => {
+  return (
+    <Typography variant="body2">
+      {text}
```
JosephMarinier, line 14: I think it's too error-prone to expect the `text`s defined everywhere to all finish with a whitespace. Maybe we should create that space using a `<Box gap={1}`? It might also give more flexibility for aligning the icon on the text.
```suggestion
    <Box display="flex" gap={1}>
      <Typography variant="body2">{text}</Typography>
```

webapp/src/components/Analysis/PreviewCard.tsx
nandhinibsn: yes Thats the explanation I asked for.. thanks

webapp/src/components/Description.tsx
```diff
@@ -0,0 +1,31 @@
...
+        >
+          <Link href={`${DOCS_URL + link}`} color="secondary" target="_blank">
+            <LinkIcon sx={{ position: "relative", top: 5, marginRight: 0.2 }} />
+            Learn more
+          </Link>
+        </Typography>
+      )}
+    </Box>
+  );
+};
```
JosephMarinier, line 30: I see a couple of potential improvements:
* Avoid the gap id `text` is not defined by wrapping it in a `{text && ...}`
* I think we can remove a bunch of stuff that doesn't do anything:
  - `alignContent="center"`
  - `display="inline"`
  - `align="center"`
  - `position: "relative"`
  - ``{`${DOCS_URL + link}`}`` that is the same as `{DOCS_URL + link}`
* I think we can simplify the `bottom: 5` and `top: 5` thing with a single `marginY: "-5px"`.

All together, is results in the following:
```suggestion
const Description: React.FC<Props> = ({ text, link }) => (
  <Box display="flex" gap={1}>
    {text && <Typography variant="body2">{text}</Typography>}
    {link && (
      <Link
        href={DOCS_URL + link}
        color="secondary"
        target="_blank"
        variant="body2"
      >
        <LinkIcon sx={{ marginRight: 0.5, marginY: "-5px" }} />
        Learn more
      </Link>
    )}
  </Box>
);

```

webapp/src/components/Icons/Doc.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <circle cx="26" cy="34" r="9" fill="white" stroke="#0047AB" />
+    <path
+      d="M22.1109 37.182C21.9157 37.3772 21.9157 37.6938 22.1109 37.8891C22.3062 38.0844 22.6228 38.0844 22.818 37.8891L22.1109 37.182ZM30.0355 30.4645C30.0355 30.1883 29.8117 29.9645 29.5355 29.9645L25.0355 29.9645C24.7594 29.9645 24.5355 30.1883 24.5355 30.4645C24.5355 30.7406 24.7594 30.9645 25.0355 30.9645L29.0355 30.9645L29.0355 34.9645C29.0355 35.2406 29.2594 35.4645 29.5355 35.4645C29.8117 35.4645 30.0355 35.2406 30.0355 34.9645L30.0355 30.4645ZM22.818 37.8891L29.8891 30.818L29.182 30.1109L22.1109 37.182L22.818 37.8891Z"
+      fill="#0047AB"
+    />
+  </svg>,
+  "Doc"
+);
+
+export default DocIcon;
```
JosephMarinier, line 37: No need for this file anymore?

webapp/src/components/Controls/Controls.tsx
```diff
@@ -207,6 +208,9 @@ const Controls: React.FC<Props> = ({
           <Box display="flex" alignItems="center" gap={1} whiteSpace="nowrap">
             <TuneIcon />
             <Typography variant="subtitle2">Controls</Typography>
+            <Box paddingTop={0.5}>
+              <Description link="/exploration-space/#filters" />
+            </Box>
```
JosephMarinier, line 213: With my previous suggestion, that should align perfectly without tweaking it:
```suggestion
            <Description link="/exploration-space/#filters" />
```

webapp/src/components/ThresholdPlot.tsx
```diff
@@ -195,20 +197,24 @@ const ThresholdPlot: React.FC<Props> = ({ jobId, pipeline }) => {
...
-            <Tooltip
-              key={`y label right ${i}`}
-              placement="top"
-              title={majorGridLines[y].tooltip}
+            <Typography
+              gridColumn="end"
+              gridRow={`${1 + i} tick`}
+              lineHeight={1}
+              marginLeft={1}
+              marginTop={-1}
```
JosephMarinier, line 205: I think it was better better when the tooltip was on the whole label, not just the icon, since the icon is very small and somewhat difficult to hover. Would you mind switching it back? What do you think?

webapp/src/pages/UtteranceDetail.tsx
```diff
@@ -255,12 +260,21 @@ const UtteranceDetail = () => {
...
+              <DatasetSplitToggler
+                value={neighborsDatasetSplitName}
+                onChange={(value) =>
+                  value && setNeighborsDatasetSplitName(value)
+                }
+              />
+            </Box>
+          </>
         )}
         <div className={classes.tabContent}>
```
JosephMarinier, line 279: Details, but can we define the descriptions for the two tabs inside this `div` instead of defining one next to the `DatasetSplitToggler`?

webapp/src/components/ThresholdPlot.tsx
JosephMarinier: Also, the current setup results in an error in the console, as
> Each child in a list should have a unique "key" prop.

Switching it back will fix this, as the `Tooltip` has the `key`.

webapp/src/components/ThresholdPlot.tsx
```diff
@@ -195,20 +197,24 @@ const ThresholdPlot: React.FC<Props> = ({ jobId, pipeline }) => {
...
             >
-              <Typography
-                gridColumn="end"
-                gridRow={`${1 + i} tick`}
-                lineHeight={0}
-                marginLeft={1}
-              >
+              <Box display="flex" alignItems="center">
```
JosephMarinier, line 207: This `Box` results in another error in the console, as
> <div> cannot appear as a descendant of <p>

I think you can simply remove the `Box` and instead adjust the `Typography`'s `marginTop={-1}` to `marginTop={-1.5}`.

webapp/src/components/Description.tsx
JosephMarinier: Good question. I have seen even MUI examples with some negative margins, so I think it's completely fine.

webapp/src/components/Icons/Doc.tsx
nandhinibsn: its not required in any other place but its a cool icon for future use.. At the same time, its not to good to keep this to hold extra memory for it as well

webapp/src/components/Controls/Controls.tsx
nandhinibsn: ohh yes right..

webapp/src/components/Description.tsx
JosephMarinier: There are still two sub-points that are not resolved:
> * I think we can remove a bunch of stuff that doesn't do anything:
>   - `alignContent="center"`
>   - ``{`${DOCS_URL + link}`}`` that is the same as `{DOCS_URL + link}`

webapp/src/pages/UtteranceDetail.tsx
```diff
@@ -84,6 +85,20 @@ const UtteranceDetail = () => {
...
+      "Shown here are the result of the perturbation tests that were automatically run to test the model's robustness to minor variations.",
+  };
+
+  const UTTERANCE_DETAILS_TAB_DOC: Record<string, string> = {
+    similarity:
+      "/exploration-space/utterance-details/#semantically-similar-utterances",
+    perturbedUtterances:
+      "/exploration-space/utterance-details/#behavioral-tests",
+  };
+
```
JosephMarinier, line 101: Can you please define those constants outside of the `UtteranceDetail` function? Right bellow the imports? That way, they don't get redefined on every render.

webapp/src/pages/UtteranceDetail.tsx
JosephMarinier: Fantastic! Thank you!

webapp/src/pages/UtteranceDetail.tsx
```diff
@@ -254,6 +273,12 @@ const UtteranceDetail = () => {
             />
           </Tabs>
         </Box>
+        {view && (
+          <Description
+            text={UTTERANCE_DETAILS_TAB_DESCRIPTION[view]}
+            link={UTTERANCE_DETAILS_TAB_DOC[view]}
+          />
+        )}
```
JosephMarinier, line 281: The `view` is always defined, no need to check for it.
```suggestion
        <Description
          text={UTTERANCE_DETAILS_TAB_DESCRIPTION[view]}
          link={UTTERANCE_DETAILS_TAB_DOC[view]}
        />
```

webapp/src/components/ThresholdPlot.tsx
JosephMarinier: I found a better way! By wrapping the `Typography` inside a `<Box height={0}`, I avoid hardcoding any non-zero vertical dimension (like `marginTop: -1`, `marginY: -4px` and `lineHeight: 1`). I committed ca64900888bc307e69185abf00e1e1224dda8dc8 with my change. Makes sense?

webapp/src/components/Description.tsx
```diff
@@ -0,0 +1,30 @@
+import LinkIcon from "@mui/icons-material/InsertLink";
```
JosephMarinier, line 1: As discussed offline, to be more semantically rigorous, you can use the identical `Link` icon (instead of `InsertLink`). That way, if ever one day Material MUI changes the `InsertLink` to add a + sign or something, we won't be affected.
```suggestion
import LinkIcon from "@mui/icons-material/Link";
```

webapp/src/components/Description.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+      link="https://servicenow.github.io/azimuth/"
+    />
+  );
+
+  await screen.findByText("test description");
+
+  const links: HTMLAnchorElement[] = screen.getAllByRole("link");
+  expect(links[0].href).toContain("https://servicenow.github.io/azimuth/");
+  await screen.findByText("Learn more");
+  await screen.getAllByTestId("InsertLinkIcon");
```
JosephMarinier, line 24: This test will need to be adapted. Also, I don't think the `All` is useful.
```suggestion
  await screen.getByTestId("LinkIcon");
```

webapp/src/components/Description.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+      text="test description"
+      link="https://servicenow.github.io/azimuth/"
+    />
+  );
+
+  await screen.findByText("test description");
+
+  const links: HTMLAnchorElement[] = screen.getAllByRole("link");
+  expect(links[0].href).toContain("https://servicenow.github.io/azimuth/");
+  await screen.findByText("Learn more");
```
JosephMarinier, line 23: I don't think any of the asynchronous `await` + `find...` are useful in this file. You should be able to use `get...` everywhere.
```suggestion
  screen.getByText("Learn more");
```

JosephMarinier (approved):
After the small comments in the test, that should be good to go!

webapp/src/components/Description.test.tsx
JosephMarinier: Still have the `All` to remove.

webapp/src/components/Description.test.tsx
JosephMarinier: Still should change `find` to `get`.

webapp/src/components/Description.test.tsx
JosephMarinier: Done in f1bc6e32c2297b2b1c88e923c0ccc02994215ad6

nandhinibsn:
Great thanks for reviewing it @JosephMarinier . so are we good to merge the changes or do we need Gab's review once for the text content?

JosephMarinier:
I think we can merge, and people will have the opportunity to check the test contents before the next release.

Obtenir Outlook pour Android<https://aka.ms/AAb9ysg>
________________________________

PR #159, opened 2022-06-30
Add visual bars to Performance Analysis table

nandhinibsn:
## Description: Added a visual bar display for the following columns in performance analysis table and semantic similar utterance table
Performance analysis table columns:
1. Outcome columns
2. custom metric columns
3. ECE
Semantic similar utterance column:
1. similarity

## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/VisualBar.tsx
+++ b/webapp/src/components/VisualBar.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Box, BoxProps } from "@mui/material";
+import { Box, BoxProps, Typography } from "@mui/material";
 import { motion } from "framer-motion";
 
 type Props = {
@@ -8,17 +8,12 @@ type Props = {
   color: BoxProps["bgcolor"];
 };
 
-const VisualBar: React.FC<Props> = ({ formattedValue, value, color }) => {
-  return (
-    <Box
-      display="grid"
-      gridTemplateColumns="46px auto"
-      width="100%"
-      alignItems="center"
-      textAlign="right"
-      gap={1}
-    >
+const VisualBar: React.FC<Props> = ({ formattedValue, value, color }) => (
+  <Box width="100%" display="flex" gap={1}>
+    <Typography align="right" variant="inherit" width={46}>
       {formattedValue}
+    </Typography>
+    <Box flex="1">
       <Box
         component={motion.div}
         key={formattedValue}
@@ -32,7 +27,7 @@ const VisualBar: React.FC<Props> = ({ formattedValue, value, color }) => {
         bgcolor={color}
       />
     </Box>
-  );
-};
+  </Box>
+);
 
 export default React.memo(VisualBar);


nandhinibsn:
Screenshot of performance Analysis table with visual bars
image
Added visual bars to similarity column
image

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +181,99 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
       field: outcome,
       headerName: OUTCOME_PRETTY_NAMES[outcome],
+      minWidth: 105,
       renderHeader: () => OutcomeIcon({ outcome }),
-      valueGetter: ({ row }) => row.outcomeCount[outcome] / row.utteranceCount,
-      valueFormatter: percentageFormatter,
+      renderCell: ({ row }: GridCellParams<undefined, Row>) =>
+        isNaN(row.outcomeCount[outcome] / row.utteranceCount) ? (
+          "--%"
+        ) : (
```
JosephMarinier, line 195: Can we put that inside the `Box` to align the `NaN`s with the other percentages?
177602054-a3452ef9-dbcf-434b-8558-497b54e2a502

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +181,99 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
+              1
+            )}
+            <Box
+              component={motion.div}
+              key={outcome}
+              overflow="auto"
+              height="90%"
+              animate={{
+                width: `${
+                  (100 * row.outcomeCount[outcome]) / row.utteranceCount || 1
```
JosephMarinier, line 213: I don't think the width should be 1% when the value is 0%. Any reason for that?
```suggestion
                  (100 * row.outcomeCount[outcome]) / row.utteranceCount
```

JosephMarinier: Interesting. Let's keep it, then, and review that with the team. I find it wrong, but it's a very tiny imprecision so it might not matter.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -130,6 +129,17 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
     );
   };
 
+  const VisualBarPercentage = (value: number, bgColor: string) =>
+    isNaN(value) ? (
+      "--%"
+    ) : (
+      <VisualBar
+        value={formatRatioAsPercentageString(value as number, 1)}
+        width={100 * value || 1}
```
JosephMarinier, line 138: I think the `100 * value || 1` should happen inside `VisualBar`, so that this logic is not duplicated. Also, it's currently not obvious that `VisualBar` expects a number from 1 to 100, so I think it would be clearer:
```suggestion
        width={value}
```

nandhinibsn:
Recent screenshot:

image

webapp/src/components/Metrics/PerformanceAnalysis.tsx
nandhinibsn: I think we can show no bar for 0% and see what the user think of it. what say?

webapp/src/components/Metrics/PerformanceAnalysis.tsx
nandhinibsn: I agree that part. I removed all that logic of checking NAN and left it to "formatRatioAsPercentageString" function to handle that. my recent commit had that change.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +191,40 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
       ...NUMBER_COL_DEF,
       field: "ece",
       headerName: "ECE",
-      valueFormatter: twoDigitFormatter,
+      minWidth: 105,
+      renderCell: ({ value }: GridCellParams) => (
+        <VisualBar
+          value={(value as number).toFixed(2)}
+          width={isNaN(value) ? 0 : value}
+          bgColor="#0b012e"
```
JosephMarinier, line 225: This one is very close to
```suggestion
          bgColor={theme.palette.primary.dark}
```

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +191,40 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
     })),
     ...customMetricNames.map<Column<Row>>((metricName) => ({
       ...NUMBER_COL_DEF,
       field: metricName,
       headerName: metricName,
-      valueGetter: ({ row }) => row.customMetrics[metricName],
-      valueFormatter: percentageFormatter,
+      minWidth: 105,
+      renderCell: ({ row }: GridCellParams<undefined, Row>) =>
+        VisualBarPercentage(row.customMetrics[metricName], "#d5d1e3"),
```
JosephMarinier, line 214: But this one I'm not sure how/where to define it. It's like our primary color, but a different shade. Maybe using `alpha(theme.palette.primary.light, <something>)` for now?

JosephMarinier (commented):
I feel like those colors should come from our theme.

webapp/src/components/Utterance/SimilarUtterances.tsx
```diff
@@ -76,7 +77,13 @@ const SimilarUtterances: React.FC<Props> = ({
       headerName: "Similarity",
       description:
         "Cosine similarity (1 indicates the utterance is identical while 0 indicates it is unrelated)", // tooltip
-      valueFormatter: (params) => (params.value as number).toFixed(2),
+      renderCell: ({ value }: GridCellParams) => (
+        <VisualBar
+          formattedValue={(value as number).toFixed(2)}
+          width={isNaN(value) || value < 0 ? 0 : value}
```
JosephMarinier, line 83: Two things:
- You wouldn't need the `isNaN(value)` if you simply do `value > 0 ? value : 0`, since `NaN > 0` returns false.
- I would still put that logic inside `VisualBar`.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +191,40 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
       ...NUMBER_COL_DEF,
       field: outcome,
       headerName: OUTCOME_PRETTY_NAMES[outcome],
+      minWidth: 105,
       renderHeader: () => OutcomeIcon({ outcome }),
-      valueGetter: ({ row }) => row.outcomeCount[outcome] / row.utteranceCount,
-      valueFormatter: percentageFormatter,
+      renderCell: ({ row }: GridCellParams<undefined, Row>) =>
+        VisualBarPercentage(
+          row.outcomeCount[outcome] / row.utteranceCount,
```
JosephMarinier, line 204: You need to keep the `valueGetter` to keep the sorting working. That should then allow you to use `value` directly in `renderCell`. Something like
```suggestion
      valueGetter: ({ row }) => row.outcomeCount[outcome] / row.utteranceCount,
      renderCell: ({ value }: GridCellParams<number>) =>
        VisualBarPercentage(
          value,
```

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -130,6 +129,14 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
     );
   };
 
+  const VisualBarPercentage = (value: number, bgColor: string) => (
+    <VisualBar
+      formattedValue={formatRatioAsPercentageString(value as number, 1)}
```
JosephMarinier, line 134: No need for "dangerous" `as number`
```suggestion
      formattedValue={formatRatioAsPercentageString(value, 1)}
```

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -130,6 +129,14 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
     );
   };
 
+  const VisualBarPercentage = (value: number, bgColor: string) => (
```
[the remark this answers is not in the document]
JosephMarinier: And it's a normal function, so it should begin with a lowercase letter.

webapp/src/components/Utterance/SimilarUtterances.tsx
nandhinibsn: yep... applied that change

webapp/src/components/Metrics/PerformanceAnalysis.tsx
JosephMarinier: I think `0.2` is close enough.

JosephMarinier: I cleaned up my code to show you how I calculated it:
```ts
  const average = (arr: number[]) => arr.reduce((a, b) => a + b) / arr.length;
  const channels: [number, number][] = [
    [1, 3],
    [3, 5],
    [5, 7],
  ];
  const desiredColor = "#d5d1e3";
  const alphaPerChannel = channels.map(
    (range) =>
      (255 - parseInt(desiredColor.substring(...range), 16)) /
      (255 - parseInt(theme.palette.primary.light.substring(...range), 16))
  );
  console.log(average(alphaPerChannel), alphaPerChannel);
```
which prints:
```
0.20518460670744934 [0.2153846153846154, 0.233502538071066, 0.16666666666666666]
```

nandhinibsn: 0.2 is more like a grey  but screenshot has purple shade. I added some changes to the theme. Let me know what do you think?

webapp/src/components/Metrics/PerformanceAnalysis.tsx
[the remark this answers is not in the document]
JosephMarinier: Only the React function components start with an uppercase letter. Basically, it's think that you can use `<LikeThis />`. In fact, React fails if we define a function component that starts with a lowercase letter. I guess that comes from the old days when we were defining React components as classes, which must start with an uppercase letter.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
JosephMarinier: Yes, you can see the blue (3rd) channel is a bit off from the others in my little array. I don't know where the purple of the mockup comes from. Maybe we should revise the theme to align with that. For this PR, let's just get something that works.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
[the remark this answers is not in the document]
nandhinibsn: okay understood!.. thanks.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
JosephMarinier: I like the idea! 👍 One small comment though: https://github.com/ServiceNow/azimuth/pull/159/files#r916100361

webapp/src/components/Metrics/PerformanceAnalysis.tsx
JosephMarinier: I don't have a strong preference either way.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +191,45 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
     })),
     {
       ...NUMBER_COL_DEF,
       field: "ece",
       headerName: "ECE",
-      valueFormatter: twoDigitFormatter,
+      minWidth: 105,
+      renderCell: ({ value }: GridCellParams) => (
+        <VisualBar
+          formattedValue={(value as number).toFixed(2)}
```
JosephMarinier, line 228: It's best to define the more specific type `GridCellParams<number>` than forcing the type with `as number`.
```suggestion
      renderCell: ({ value }: GridCellParams<number>) => (
        <VisualBar
          formattedValue={value.toFixed(2)}
```

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +191,45 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
       field: "utteranceCount",
       headerName: "Utterance Count",
-      minWidth: 160,
+      minWidth: 150,
     },
     ...ALL_OUTCOMES.map<Column<Row>>((outcome) => ({
       ...NUMBER_COL_DEF,
       field: outcome,
       headerName: OUTCOME_PRETTY_NAMES[outcome],
+      minWidth: 105,
```
JosephMarinier, line 200: Instead of overwritting `minWidth` 3 times with the same value, I would suggest changing it in `NUMBER_COL_DEF`.

webapp/src/components/Utterance/SimilarUtterances.tsx
```diff
@@ -76,7 +77,15 @@ const SimilarUtterances: React.FC<Props> = ({
...
       description:
         "Cosine similarity (1 indicates the utterance is identical while 0 indicates it is unrelated)", // tooltip
-      valueFormatter: (params) => (params.value as number).toFixed(2),
+      renderCell: ({ value }: GridCellParams) => (
+        <VisualBar
+          formattedValue={(value as number).toFixed(2)}
+          width={value}
+          bgColor={
+            value > 0.5 ? theme.palette.info.light : theme.palette.primary.dark
+          }
```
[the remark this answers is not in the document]
nandhinibsn: okay thats a great news then I can change it to primary light.

webapp/src/components/VisualBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+        component={motion.div}
+        key={formattedValue}
+        overflow="auto"
+        height="90%"
+        animate={{
+          width: `${width > 0 ? 100 * width : 0}%`,
+        }}
+        initial={false}
+        transition={{ type: "tween" }}
+        bgcolor={bgColor}
```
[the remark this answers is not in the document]
nandhinibsn: I have no particular preference but would like to stick to name color

webapp/src/components/Metrics/PerformanceAnalysis.tsx
nandhinibsn: yes thats right.. sorry I missed this part

webapp/src/styles/theme.tsx
```diff
@@ -23,7 +23,7 @@ const customTheme: Theme = createTheme({
       default: "#fafafa", // TODO awaiting design input.
     },
     primary: {
-      light: "#3C3A57",
+      light: "#d5d1e3",
```
JosephMarinier, line 26: That's a tiny detail, but would you mind using uppercase like the other colors defined here?
```suggestion
      light: "#D5D1E3",
```

nandhinibsn: okay sure going forward

webapp/src/components/VisualBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const VisualBar: React.FC<Props> = ({ formattedValue, value, color }) => {
+  return (
+    <Box
+      display="grid"
+      gridAutoColumns={50}
+      gridAutoFlow="column"
+      alignItems="center"
+    >
+      {formattedValue}
```
JosephMarinier, line 19: Can we justify the text to the right? Like it used to be. That way, all the percentage signs would line up.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -23,12 +24,14 @@ import SeeMoreLess, {
   useMoreLess,
 } from "components/SeeMoreLess";
 import { Table, Column } from "components/Table";
+import VisualBar from "components/VisualBar";
 import React from "react";
 import { getMetricsPerFilterEndpoint } from "services/api";
-import { DatasetSplitName, MetricsPerFilterValue } from "types/api";
+import { DatasetSplitName, MetricsPerFilterValue, Outcome } from "types/api";
```
JosephMarinier, line 30: That new import is not used, resulting in a warning.
```suggestion
import { DatasetSplitName, MetricsPerFilterValue } from "types/api";
```

webapp/src/components/VisualBar.tsx
```diff
@@ -10,12 +10,7 @@ type Props = {
 
 const VisualBar: React.FC<Props> = ({ formattedValue, value, color }) => {
   return (
-    <Box
-      display="grid"
-      gridAutoColumns={50}
-      gridAutoFlow="column"
-      alignItems="center"
-    >
+    <Box display="grid" gridAutoColumns={50} gridAutoFlow="column">
```
JosephMarinier, line 13: I think you were correct wiht the `alignItems="center"`, so that the bar is centered in the row.

webapp/src/components/VisualBar.tsx
JosephMarinier: Don't forget this! I think it would require something like `textAlign="end" gap={1}` on the outer `Box`.

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -184,28 +181,99 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
       renderHeader: () => OutcomeIcon({ outcome }),
-      valueGetter: ({ row }) => row.outcomeCount[outcome] / row.utteranceCount,
-      valueFormatter: percentageFormatter,
+      renderCell: ({ row }: GridCellParams<undefined, Row>) =>
+        isNaN(row.outcomeCount[outcome] / row.utteranceCount) ? (
+          "--%"
+        ) : (
+          <Box
+            display="grid"
+            gridAutoColumns={50}
```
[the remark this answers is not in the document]
JosephMarinier: Actually, you can get this result with very minimal changes by setting an `auto` width to the second column of the grid. Paired with a `width="100%"`, the bars should fill the remaining space automatically. Here I am also including my suggestion to justify the text right. Note that I used 46 pixels instead of 50 to compensate the 8 pixel gap that I would add. Let me know what you think!
```diff
     <Box
       display="grid"
-      gridAutoColumns={50}
-      gridAutoFlow="column"
+      gridTemplateColumns="46px auto"
+      width="100%"
+      alignItems="center"
+      textAlign="right"
+      gap={1}
```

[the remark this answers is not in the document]
JosephMarinier: Cool!
Two small things:
1. As suggested, you can remove the `gridAutoFlow="column"` as our template defines all the columns.
1. It wouldn't make a difference now, but I think `textAlign="right"` would be slithly better than `textAlign="end"`, as `end` would give a different result if one day we support languages that read from right to left. It's a small detail, but it doesn't hurt to make it right.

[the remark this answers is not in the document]
nandhinibsn: Make sense!.. right

JosephMarinier (approved):
👍

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -199,16 +189,26 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
...
       valueGetter: ({ row }) => row.outcomeCount[outcome] / row.utteranceCount,
-      renderCell: ({ value }: GridCellParams<number>) =>
-        visualBarPercentage(value, theme.palette[OUTCOME_COLOR[outcome]].main),
+      renderCell: ({ value }: GridCellParams<number>) => (
+        <VisualBar
+          formattedValue={formatRatioAsPercentageString(value, 1)}
+          value={value}
+          color={(theme) => theme.palette[OUTCOME_COLOR[outcome]].main}
+        />
+      ),
```
JosephMarinier, line 198: @nandhinibsn

> can we still not go with function? any reason to call the component directly on each function as format is similar for outcome and custom metric columns?

No hard requirement. As I said in the commit message, I simply did that "to make definitions more straightforward and avoid some typing". I was too lazy to import `BoxProps` to use `BoxProps["color"]` in the function. Also, the function was only used twice, and didn't support the third case of the ECE column. Now it's more uniform and you don't need to scroll back an forth to understand what is going on. But that's a detail. Would you prefer that I put it back? I don't mind.

nandhinibsn:
Screenshot captured after final changes done.
image

JosephMarinier:
Beauuutiful!

webapp/src/components/Metrics/PerformanceAnalysis.tsx
nandhinibsn: okay then it make sense and I agree in terms of readability and to reduce one import which is boxprops. Let it be this way then

gabegma:
It looks SO good!! I tagged you Nandhini in our Mattermost thread.

nandhinibsn:
thank you @JosephMarinier for the Approval and your time to contribute some final change and reviews. I will merge this once the checks are completed.

nandhinibsn:
@gabegma  yes sorry I forgot to notice that. I was looking for updates in the teams and forgot about mattermost..

PR #160, opened 2022-07-07
Update python dependencies

gabegma:
## Description:
- Fix some dependencies to versions that we can use.
- Move some dependencies to dev.

## Checklist:

it.

  our users.
* [x] **DEV CHANGES.**


pyproject.toml
```diff
@@ -90,6 +71,10 @@ seaborn = "^0.11.1"
 hdbscan = "^0.8.27"
 pre-commit = "^2.15.0"
 isort = "^5.9.3"
+types-simplejson = "^3.17.0"
+types-filelock = "^0.1.5"
+types-orjson = "^3.6.0"
```
Dref360, line 76: `fastapi-profiler` is not a dev dep: https://github.com/ServiceNow/azimuth/blob/main/azimuth/app.py#L10

We probably should remove it from `app.py` altogether as we do not use it often.

gabegma: Yeah, I switched it back. Ok, I can remove it - thanks! 

However, I'm having issues with the router test on this branch. At first, they were all falling, but then I tried a fix in 297238f, but locally they hang most of the time now. On Github, they passed though. (I still need to fix one test related to something else - I haven't looked into it yet).

gabegma:
@JosephMarinier @Dref360 can you both review this makes sense to you?

pyproject.toml
```diff
@@ -20,76 +20,60 @@ scikit-learn = "^1.0.1"
...
-lab = "^6.2"
 ipywidgets = ">=7.6"
-notebook = "^6.4.12"
 black = ">=20.8b1"
 mypy = "^0.910"
 bandit = "^1.7.0"
 matplotlib = "^3.4.1"
 umap-learn = "^0.5.1"
-seaborn = "^0.11.1"
+seaborn = "0.11.1"
```
JosephMarinier, line 70: Is there a reason why we keep the `^` for other dependencies but we remove it for this one?

pyproject.toml
```diff
@@ -20,76 +20,60 @@ scikit-learn = "^1.0.1"
 faiss-cpu = "=1.6.5"
 tensorflow = "~2.8.0"
 tensorflow-text = "~2.8.0"
-### tensorflow-estimator: poetry issues when versions like "x.y.0rc1" are released
```
JosephMarinier, line 23: Is this comment not relevant anymore?

pyproject.toml
gabegma: Yes, otherwise poetry was using `0.11.2`, and I need `0.11.1`.

tests/test_routers/conftest.py
```diff
@@ -13,6 +14,9 @@
 from azimuth.config import AzimuthConfig
 from tests.utils import DATASET_CFG, SIMPLE_PERTURBATION_TESTING_CONFIG
 
+# Fixes issue RuntimeError: This event loop is already running.
+nest_asyncio.apply()  # From https://github.com/encode/starlette/issues/440
```
Dref360, line 18: This happenned when bumping FastAPI or Dask?

gabegma: Good question - I can try again without it - now that Dask is no longer updated.

gabegma: good catch - is no longer needed!

gabegma:
@JosephMarinier - are you ok with me merging as is? Did you want me to change something based on your comments?

pyproject.toml
JosephMarinier: @gabegma
> are you ok with me merging as is?

Yes, it looks good to me.
I am wondering why we don't freeze all dependencies (remove all the `^`s and `~`s), if we don't want them to update. But that's another story.

JosephMarinier (approved):

pyproject.toml
gabegma: Yeah, I think it could be a good avenue as well and it would make it easier in a way because we wouldn't need to go check the `.lock` to see what version was actually installed. However, it would maybe prevent dependabot to make useful suggestions of what should be bumped for security reasons?  And make it slightly more annoying when adding dependencies because we'll need to deal with conflicts. We can think about it and leave it for another PR.

JosephMarinier: Exactly. Yes, Dependabot won't be able to make PRs.

PR #174, opened 2022-07-15
Singular filters

JosephMarinier:
## Description:

The "problem" is that our metrics per filter are returned for singular `label` and `prediction`. To tie the two together in the Performance Analysis table, we have to implement a dodgy logic for when to add a `s` and when not (for example: not for smart tags families). This could lead to unnoticed errors if one day we add a filter that pluralizes as `y` :arrow_right: `ies` for example.

Here's why putting all filters singular makes sense to me (as opposed to putting everything plural): Even though the value of the query parameter is an array of multiple **labels**, what it means as a filter is that the **label** or each utterance (and each utterance has only one label) is one of those values. I would read `label=A,B` as "label is A or B".

* Rename filters to singular
* Refactor controls now that filters and counts per filter align

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/model_performance/metrics.py
+++ b/azimuth/modules/model_performance/metrics.py
@@ -284,13 +284,13 @@ class MetricsPerFilterModule(AggregationModule[AzimuthConfig]):
         """
         filter_copy = initial_filter.copy(deep=True)
         if label is not None:
-            filter_copy.labels.append(label)
+            filter_copy.label.append(label)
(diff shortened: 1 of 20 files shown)


nandhinibsn:
Good approach to change the name for label and prediction. I guess we can avoid giving plurals in the name going forward. whats your view point?

christyler3030 (approved):

PR #177, opened 2022-07-18
Use snake_case query parameters

### LINKED ISSUE #176: Make all query parameters snake_case
Make all query parameters snake_case instead of camelCase, to avoid bugs like #175 (and multiple more in the past).

- Remove all query parameters' camelCase `alias`es in the back end.
- Convert to and from camelCase in front end, using `snakeToCamelCase()` (and probably a new `camelToSnakeCase()`).

JosephMarinier:
Resolve #176

## Description:

* Remove all query parameters' camelCase aliases
  - Use regular expression to replace all `,\n? +alias="\w+[A-Z]\w+"` with ``.
  - Not having to define `alias`es to our query parameters will avoid bugs when we forgot to define one, like #175 last week and multiple other bugs in the past.
* Convert query parameters to and from snake_case in front end

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/utils/helpers.ts
+++ b/webapp/src/utils/helpers.ts
@@ -19,6 +19,9 @@ export const raiseSuccessToast = (message: string) => {
   toast.success(message);
 };
 
+const camelToSnakeCase = (s: string) =>
+  s.replace(/[A-Z]/g, (c) => `_${c.toLowerCase()}`);
+
 const convertNumber = (s: string | null) =>
   s === null ? undefined : Number(s);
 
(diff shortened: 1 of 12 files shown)


webapp/src/utils/helpers.ts
```diff
@@ -19,6 +19,9 @@ export const raiseSuccessToast = (message: string) => {
   toast.success(message);
 };
 
+const camelToSnakeCase = (s: string) =>
```
christyler3030, line 22: Detail: This doesn't actually enforce camel case do you could end up with something like  `ThisName` being converted to `This_name`

JosephMarinier: I will merge as is and revisit that later if you think I should, to avoid conflicts with the backend changes.

christyler3030: Ya I don't have any recommendations here I just wanted to point it out. I think if we wanted to make it slightly safer I wouldn't try to type it I would just handle the case where the first letter was capitalized.

JosephMarinier: Ok, but handle it how? What do you expect from `camelToSnakeCase("ThisName")`? Throwing an error saying the input is not camelCase?

christyler3030: I would say just don't do the `replace` to avoid inserting an extra underscore. (But also really not important now I think)

PR #186, opened 2022-07-25
Azimuth test

### LINKED ISSUE #179: Add front-end tests for the performance analysis module

nandhinibsn:
Resolve #179 

## Description:
This PR will have test coverage set up for the following sections under Dashboard page.

Dashboard section:
1. Dataset Class distribution analysis
2. Performance Analysis
3. Behavioural testing
4. Behavioural testing summary

## Checklist:

  our users.
* [ ] **DEV CHANGES.**


webapp/package.json
```diff
@@ -11,6 +11,7 @@
     "@mui/system": "^5.5.1",
     "@mui/x-data-grid": "^5.6.1",
     "@reduxjs/toolkit": "^1.8.1",
+    "@testing-library/react-hooks": "^8.0.1",
```
JosephMarinier, line 14: This should be in `devDependencies` instead, right?

webapp/package.json
```diff
@@ -28,7 +29,8 @@
   "scripts": {
     "start": "PORT=3005 react-app-rewired start",
     "build": "react-app-rewired build",
-    "test": "react-app-rewired test",
+    "testAll": "react-app-rewired test --coverage --watchAll=false",
```
JosephMarinier, line 32: I'm not convinced we need to add a separate "script" just to add an argument. You can instead call the existing `yarn test` with the argument, like `yarn test --watchAll=false`. If you find this option useful (what does it do?), you could document it in the README or something.

JosephMarinier:
Can you remove webapp/package-lock.json please? It should not exist since we use `yarn`.

webapp/yarn.lock
```diff
@@ -4,7 +4,7 @@
 
 "3d-view@^2.0.0":
   version "2.0.1"
-  resolved "https://registry.yarnpkg.com/3d-view/-/3d-view-2.0.1.tgz#2e174571c48215736b376bb66938a3513dad2179"
+  resolved "https://registry.npmjs.org/3d-view/-/3d-view-2.0.1.tgz"
```
JosephMarinier, line 7: Those changes in webapp/yarn.lock seem wrong. Can you revert them and simply run `yarn` again (after moving the new dependency in `devDependencies`)?

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+describe("DatasetInfo API", () => {
+  const handlers = [getDatasetInfo];
+  const server = setupServer(...handlers);
+
+  beforeAll(() => server.listen());
+  afterEach(() => server.resetHandlers());
+  afterAll(() => server.close());
+
+  test("shoud have 2 toggle buttons displayed and enabled with icons", () => {
```
JosephMarinier, line 27: Missing a `l`. Also, that's a details, but you are using the perfect sentence construction to use the alias `it()` instead of `test()`. By the way, I also really like that, as the output is super clear in case of failure. 👍 
```suggestion
  it("should have 2 toggle buttons displayed and enabled with icons", () => {
```
```
DatasetInfo API > should have 2 toggle buttons displayed and enabled with icons
```

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+} from "mocks/api/mockDatasetInfoAPI";
+
+const onChangeMock = (value: DatasetSplitName) => jest.fn();
+const renderDataSplitToggler = (value: DatasetSplitName) =>
+  renderWithRouterAndRedux(
+    <DatasetSplitToggler value={value} onChange={onChangeMock} />,
+    { route: "/local", path: "/:jobId" }
+  );
+
+describe("DatasetInfo API", () => {
```
JosephMarinier, line 19:
```suggestion
describe("DatasetSplitToggler", () => {
```

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  test("should the toggle button Training set displayed with ModelTrainingIcon", () => {
+    renderDataSplitToggler("eval");
+    expect(
+      within(screen.getByText("Training Set")).findByRole("ModelTrainingIcon")
+    );
+  });
+  test("should be able to test onchange updates the value property", () => {
+    renderDataSplitToggler("train");
+    const buttonEl = screen.getByRole("button", { name: /Training/i });
+    expect(buttonEl).toBeInTheDocument();
```
JosephMarinier, line 53: The `getByRole()` would have thrown if it was not in the document. In fact, I haven't yet seen a situation where `toBeInTheDocument()` was relevant. It must be for special cases.
```suggestion
```

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    expect(screen.queryByText("Training Set")).toBeInTheDocument();
+    expect(screen.queryByTestId("ModelTrainingIcon")).toBeInTheDocument();
+    expect(
+      screen.getAllByRole("button", { name: /Evaluation/i })[0]
+    ).toHaveAttribute("aria-pressed", "true");
+  });
+
+  test("should the toggle button Evaluation set displayed with SpeedIcon", () => {
+    renderDataSplitToggler("eval");
+    expect(within(screen.getByText("Evaluation Set")).findByRole("SpeedIcon"));
```
JosephMarinier, line 41: `findBy` returns a promise, which is not `await`ed here, so this test would pass even if we changed it to
```suggestion
    expect(within(screen.getByText("Evaluation Set")).findByRole("potato"));
```
However, using `getBy`, this would expectedly fail
```suggestion
    expect(within(screen.getByText("Evaluation Set")).getByRole("SpeedIcon"));
```
That is because the `SpeedIcon` is not a role. I think you can use `getByTestId` instead:
```suggestion
    expect(within(screen.getByText("Evaluation Set")).getByTestId("SpeedIcon"));
```
Same thing for next test.

General comment: When writing a test, we have to make sure it actually tests something. Here, as I showed for example, I would switch `SpeedIcon` with `potato` to make sure the test fails. In more complex situations, I would consciously change the component we are testing to make it fail (for example here changing the icon in `DataSplitToggler`). If the test still passes, it's that it doesn't do its job.

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    renderDataSplitToggler("train");
+    const buttonEl = screen.getByRole("button", { name: /Training/i });
+    expect(buttonEl).toBeInTheDocument();
+    expect(buttonEl).toBeDisabled();
+    fireEvent.click(buttonEl, {
+      target: {
+        value: "train",
+      },
+    });
+    expect(screen.getAllByRole("button", { name: /Training/i })[0]).toHaveClass(
```
JosephMarinier, line 60: Any reason for using regular expressions in this test and not in the others?
```suggestion
    expect(screen.getByRole("button", { name: "Training Set" })).toHaveClass(
```

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    expect(screen.queryByText("Evaluation Set")).toBeInTheDocument();
+    expect(screen.queryByTestId("SpeedIcon")).toBeInTheDocument();
+    expect(screen.queryByText("Training Set")).toBeInTheDocument();
+    expect(screen.queryByTestId("ModelTrainingIcon")).toBeInTheDocument();
+    expect(
+      screen.getAllByRole("button", { name: /Evaluation/i })[0]
+    ).toHaveAttribute("aria-pressed", "true");
+  });
+
+  test("should the toggle button Evaluation set displayed with SpeedIcon", () => {
```
JosephMarinier, line 39: Just to make a bit more sense
```suggestion
  test("evaluation set should be displayed with SpeedIcon", () => {
```

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  test("should the toggle button Training set displayed with ModelTrainingIcon", () => {
+    renderDataSplitToggler("eval");
+    expect(
+      within(screen.getByText("Training Set")).findByRole("ModelTrainingIcon")
+    );
+  });
+  test("should be able to test onchange updates the value property", () => {
+    renderDataSplitToggler("train");
```
JosephMarinier, line 51: Do you mean to test that `onChange updates the value property`? But you mocked `onChange` to do nothing. And you initialized the `DataSplitToggler` with `"train"`, so the training set was already selected, and the click doesn't change anything.
```suggestion
  test("onChange updates the value property", () => {
    renderDataSplitToggler("eval");
```
On a general note, to test that something changes, you should test the initial state before changing it. Here, something like `expect(buttonEl).not.toHaveClass("Mui-selected");`. You would have noticed that the test doesn't work.

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const buttonEl = screen.getByRole("button", { name: /Training/i });
+    expect(buttonEl).toBeInTheDocument();
+    expect(buttonEl).toBeDisabled();
```
JosephMarinier, line 54: When the button is disabled (before it receives the dataset info), the button is not clickable, so it's an odd thing to `expect` here.

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const server = setupServer(...handlers);
+
+  beforeAll(() => server.listen());
+  afterEach(() => server.resetHandlers());
+  afterAll(() => server.close());
+
+  test("shoud have training set toggle button disabled", () => {
+    renderDataSplitToggler("eval");
+    expect(screen.queryAllByRole("button")).toHaveLength(2);
+    expect(screen.queryByText("Training Set")).toBeDisabled();
```
JosephMarinier, line 77: The evaluation set is also disabled, both are, since we don't wait for dataset info response.
Related, the console shows a bunch of warnings about not intercepting `dataset_info`.
```
      [MSW] Warning: captured a request without a matching request handler:

        • GET http://localhost/api/local/dataset_info

      If you still wish to intercept this unhandled request, please create a request handler for it.
      Read more: https://mswjs.io/docs/getting-started/mocks
```

webapp/package.json
nandhinibsn: Ohh got it why it went up there. I did direct yarn install so it might have added under dependencies. I am also not sure we need this library. I will check and remove this if its not used.

webapp/src/components/Controls/DatasetSplitToggler.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    { route: "/local", path: "/:jobId" }
+  );
+
+describe("DatasetInfo API", () => {
+  const handlers = [getDatasetInfo];
+  const server = setupServer(...handlers);
+
+  beforeAll(() => server.listen());
+  afterEach(() => server.resetHandlers());
+  afterAll(() => server.close());
```
JosephMarinier, line 25: We could pass `availableDatasetSplits` to `DatasetSplitToggler` so it doesn't need to fetch `datasetInfo` at all. All 3 callers, or their callers, already have `availableDatasetSplits` at hand. That way, we wouldn't need to mock the API to unit-test `DatasetSplitToggler`. I created a draft PR https://github.com/ServiceNow/azimuth/pull/209 with this suggestion, if you wish to rebase over it and simplify the tests.

With testing in mind, I think this is a good approach. Separating data (API fetches) and presentation. `DatasetSplitToggler` becomes a pure presentation component, which is a lot easier to test. I think it could apply to a lot more components in our app. Ideally, only top-level or page-level components would be responsible for fetching data. Right @christyler3030? 😉 

There are caveats to only fetching the API from page-lever, as calling the API conditionally bring a whole lot of typing problems. So, for example, I think it's fine that the different boxes in the `Dashboard` are responsible for fetching the API instead of the `Dashboard` itself, the `Dashboard` being responsible for rendering the boxes conditionally based on what features are available (dataset vs pipeline and more).

webapp/package.json
nandhinibsn: You are right. we don't require a new argument here. It was for my convenience I added to test with/without coverage while I am working on it. this will be removed and kept one argument with coverage feature on when we go ahead merge with main

webapp/yarn.lock
nandhinibsn: this went with mistake..  can be removed in next commit

nandhinibsn:
I am closing this pull request to create pull request separately for each modules

PR #190, opened 2022-07-27
Add progress info during startup

### LINKED ISSUE #135: Show progress bar during startup

Dref360:
Resolve #135 

## Description:

Add progress info during startup.
I would need someone help to get a SVG of the progress icon. I got one from Material UI, but I would like to see 3 dots waving.

Looks like:
image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/StatusCheck.tsx
+++ b/webapp/src/components/StatusCheck.tsx
@@ -1,13 +1,42 @@
-import { Box, Typography } from "@mui/material";
+import { Box, capitalize, Stack, Typography } from "@mui/material";
 import noData from "assets/launch.svg";
 import React from "react";
 import { useParams } from "react-router-dom";
 import { getStatusEndpoint } from "services/api";
 import Loading from "./Loading";
+import Grid from "@mui/material/Grid";
+import Paper from "@mui/material/Paper";
+import CheckIcon from "./Icons/Check";
+import ErrorIcon from "@mui/icons-material/Error";
+import PendingOutlinedIcon from "@mui/icons-material/PendingOutlined";
 
 type Props = {
   children: React.ReactNode;
 };
+
+interface StringByString {
+  [key: string]: any;
+}
+
+const STATUS_ICON_MAPPING: StringByString = {
+  finished: <CheckIcon />,
+  not_started: <CheckIcon />,
+  pending: <PendingOutlinedIcon />,
+  error: <ErrorIcon />,
+  lost: <ErrorIcon />,
+};
+
+const render_cell = (task_name: string, task_status: string) => (
+  <Grid item xs={1.5}>
+    <Paper variant="outlined">
+      <Box display="flex" alignItems="center" gap={1}>
+        <Typography>
+          {`${task_name.split("_").map(capitalize).join(" ")} `} {STATUS_ICON_MAPPING[task_status]}
+        </Typography>
+      </Box>
+    </Paper>
+  </Grid>
+);
 
 const StatusCheck: React.FC<Props> = ({ children }) => {
   const { jobId } = useParams<{ jobId: string }>();
@@ -31,20 +60,25 @@ const StatusCheck: React.FC<Props> = ({ children }) => {
         justifyContent="center"
         alignItems="center"
         width="100%"
-        height="75vh"
+        height="100vh"
       >
         <Box
           display="flex"
           flexDirection="column"
           alignItems="center"
           gap={4}
-          width={600}
+          width={700}
         >
           <img src={noData} alt="Startup tasks still in progress" width={400} />
           <Typography variant="h2" align="center">
             The startup tasks are still in progress. Grab a coffee and we will
             auto-refresh for you.
           </Typography>
+          <Grid container spacing={2} columnSpacing={3} columns={3}>
+            {Object.entries(status.startupTasksStatus).map(
+              ([task_name, task_status]) => render_cell(task_name, task_status)
+            )}
+          </Grid>
         </Box>
       </Box>
     );
--- a/webapp/src/components/StatusCheck.tsx
+++ b/webapp/src/components/StatusCheck.tsx
@@ -31,7 +31,8 @@ const render_cell = (task_name: string, task_status: string) => (
     <Paper variant="outlined">
       <Box display="flex" alignItems="center" gap={1}>
         <Typography>
-          {`${task_name.split("_").map(capitalize).join(" ")} `} {STATUS_ICON_MAPPING[task_status]}
+          {`${task_name.split("_").map(capitalize).join(" ")}  `}
+          {STATUS_ICON_MAPPING[task_status]}
         </Typography>
       </Box>
     </Paper>


webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,44 @@
...
 
 type Props = {
   children: React.ReactNode;
 };
 
+interface StringByString {
+  [key: string]: any;
+}
+
+const STATUS_ICON_MAPPING: StringByString = {
```
JosephMarinier, line 21: The utility type [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) helps typing an object (dictionary) like that. Also, `React.ReactElement` would be a lot more robust that `any`. Finally, in the front end code we don't normally name things "mapping". For example, `SMART_TAG_FAMILY_ICONS` is just `ICONS` and not `ICON_MAPPING`. That's a detail, but I would make it uniform by renaming to `STATUS_ICONS`.
```suggestion
const STATUS_ICONS: Record<string, React.ReactElement> = {
```

webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,44 @@
...
+  finished: <CheckIcon />,
+  not_started: <CheckIcon />,
```
JosephMarinier, line 23: In what context do we get `not_started`? Why a check icon?

webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,44 @@
...
+};
+
+const render_cell = (task_name: string, task_status: string) => (
+  <Grid item xs={1.5}>
+    <Paper variant="outlined">
+      <Box display="flex" alignItems="center" gap={1}>
+        <Typography>
+          {`${task_name.split("_").map(capitalize).join(" ")}  `}
+          {STATUS_ICON_MAPPING[task_status]}
+        </Typography>
```
JosephMarinier, line 36: I think I would capitalize only the first word, and not all of them. Some letters are obviously wrong, even as title, like `By` in `Metrics By Filter Train`. Also, the icon should be outside of the `Typography` for the `Box`'s `alignItems="center"` and `gap={1}` to work. I would put the icon to the left of the `Typography` so that all icons are aligned.
```suggestion
        {STATUS_ICON_MAPPING[task_status]}
        <Typography>{capitalize(task_name).replace(/_/g, " ")}</Typography>
```

webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,44 @@
...
 };
 
+interface StringByString {
+  [key: string]: any;
+}
+
+const STATUS_ICON_MAPPING: StringByString = {
+  finished: <CheckIcon />,
+  not_started: <CheckIcon />,
+  pending: <PendingOutlinedIcon />,
```
JosephMarinier, line 24: How about we use a spinner (`CircularProgress` from `"@mui/material"`). With a specific size and margin, it will look just like an icon.
```suggestion
  pending: <CircularProgress size={16} sx={{ margin: "2px" }} />,
```

webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,44 @@
...
+  error: <ErrorIcon />,
+  lost: <ErrorIcon />,
```
JosephMarinier, line 26: We can give icons a color. For the check, `color="success"`, and here:
```suggestion
  error: <ErrorIcon color="error" />,
  lost: <ErrorIcon color="error" />,
```

webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,44 @@
...
+};
+
+const render_cell = (task_name: string, task_status: string) => (
+  <Grid item xs={1.5}>
+    <Paper variant="outlined">
```
JosephMarinier, line 31: I think we can remove the `Paper`, for a cleaner look. Together with my other suggestions, that results in

image

webapp/src/components/StatusCheck.tsx
```diff
@@ -31,20 +61,25 @@ const StatusCheck: React.FC<Props> = ({ children }) => {
...
           <img src={noData} alt="Startup tasks still in progress" width={400} />
           <Typography variant="h2" align="center">
             The startup tasks are still in progress. Grab a coffee and we will
             auto-refresh for you.
           </Typography>
+          <Grid container spacing={2} columnSpacing={3} columns={3}>
+            {Object.entries(status.startupTasksStatus).map(
+              ([task_name, task_status]) => render_cell(task_name, task_status)
+            )}
+          </Grid>
```
JosephMarinier, line 82: Implementation detail. The `Grid` component is a bit overkill here, as it is meant to create a
> responsive layout grid [that] adapts to screen size and orientation

For example, `container columns={3}` and `item xs={1.5}` is a pretty convoluted way to say "2 columns". I don't think we will need to adapt the number of columns to the width of the page. So I propose replacing it by a "raw" `Box` with `display="grid"` and a couple of properties. We can completely remove the `<Grid item xs={1.5}>` from the cell too. At that point, `render_cell` is so small that I think we should just define the cels inline here. By the way, in JavaScript, we usually use camelCase variable names. All together, it results in something like this:
```suggestion
          <Box
            display="grid"
            rowGap={2}
            columnGap={4}
            gridTemplateColumns="repeat(2, 1fr)"
            width="100%"
          >
            {Object.entries(status.startupTasksStatus).map(([task, status]) => (
              <Box display="flex" alignItems="center" gap={1}>
                {STATUS_ICON_MAPPING[status]}
                <Typography>{capitalize(task).replace(/_/g, " ")}</Typography>
              </Box>
            ))}
          </Box>
```

webapp/src/components/StatusCheck.tsx
JosephMarinier: See comment https://github.com/ServiceNow/azimuth/pull/190/files#r942810300 first as it cover this too.

webapp/src/components/StatusCheck.tsx
Dref360: `not_started` is when startup tasks were already cached. They are not started, but the result is available. This is an artefact of the Dask scheduler which uses states named as such.

JosephMarinier: Ah-ha! Cool, thank you! Thank you! Maybe we should add a small comment here.

JosephMarinier, line 23: It's a detail, but I would suggest using the `Done` icon from MUI. It is the same as `Check`, but it seems semantically perfect in this context to use this icon when the task is "done".
```ts
import DoneIcon from "@mui/icons-material/Done";
```
and something like
```suggestion
  finished: <DoneIcon color="success" />,
  not_started: <DoneIcon color="success" />, // The task result was already cached, so the task didn't need to be started. 
```

Dref360:
Thank you for your comments! I made the changes.
I  was also thinking of changing the name of the startup task to remove the pipeline index if it`s None or 0 wyt?

`f"{startup.name}_{k.value}_{pipeline_index or ''}"` gives the following:


image

webapp/src/components/StatusCheck.tsx
JosephMarinier: Any reason for not using the size and margin I suggested? I thought it was perfect.

JosephMarinier:
> [...] remove the pipeline index if it`s None or 0 wyt?
> 
> `f"{startup.name}_{k.value}_{pipeline_index or ''}"`

Hooo, that's an easy solution! Maybe I would just avoid ending with a `_`.

Alternatively, if we brainstorm... Can we remove the 0 only if there is only one pipeline configured? And/or do we have the pipeline name we can use instead of the index?

webapp/src/components/StatusCheck.tsx
```diff
@@ -1,14 +1,24 @@
...
+import ErrorIcon from "@mui/icons-material/Error";
 
 type Props = {
   children: React.ReactNode;
 };
 
+const STATUS_ICONS: Record<string, React.ReactElement> = {
+  finished: <DoneIcon color="success" />,
+  not_started: <DoneIcon color="success" />, // Happens when the task was already computed.
+  pending: <CircularProgress />,
```
Dref360, line 17:
```suggestion
  pending: <CircularProgress size={16} sx={{ margin: "2px" }} />,
```

JosephMarinier:
> Hooo, that's an easy solution! Maybe I would just avoid ending with a `_`.

So, something like this?
```py
                    f"{startup.name}_{k.value}{pipeline_index and f'_{pipeline_index}' or ''}": v
```
or
```py
                    f"{startup.name}_{k.value}_{pipeline_index or ''}".rstrip("_"): v
```

JosephMarinier (approved):
Awesome job!
I just committed a couple of details to polish it up.

PR #191, opened 2022-07-28
Polish Performance Analysis table headers

JosephMarinier:
## Description:

Polish Performance Analysis table headers
* Align headers left
* Rename `Utterance Count` to `Total` to save 50 pixels
* Refactor metrics descriptions
  - Now expecting a string and supporting newlines to split into paragraphs. As a result, this now handles the newlines in the custom metrics descriptions.
* Add tooltips to metrics in Performance Analysis table headers
  - This also makes the memoized `columns` independent of `data`, so changing dataset split or pipeline doesn't require a `columns` redefinition.
* Avoid hardcoded number of bins for ECE
* Fix metrics width so that it doesn't vary when the numbers change

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Metrics/Metric.tsx
+++ b/webapp/src/components/Metrics/Metric.tsx
@@ -41,17 +41,14 @@ const Metric: React.FC<Props> = ({
   const classes = useStyles();
   const theme = useTheme();
 
+  const tooltip = description
+    .trim()
+    .split("\n")
+    .map((paragraph) => <Typography variant="inherit">{paragraph}</Typography>);
+
   return (
     <Box display="flex" justifyContent="flex-end">
-      <Tooltip
-        title={description
-          .trim()
-          .split("\n")
-          .map((paragraph) => (
-            <Typography variant="inherit">{paragraph}</Typography>
-          ))}
-        classes={{ popper: classes.popper }}
-      >
+      <Tooltip title={tooltip} classes={{ popper: classes.popper }}>
         <Box
           height="100%"
           paddingX={2.5}
--- a/webapp/src/components/Metrics/Metric.tsx
+++ b/webapp/src/components/Metrics/Metric.tsx
@@ -9,7 +9,6 @@ import {
   Skeleton,
   Tooltip,
   tooltipClasses,
-  TooltipProps,
   Typography,
   useTheme,
 } from "@mui/material";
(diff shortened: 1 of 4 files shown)


webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -137,11 +137,10 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
       return gridStringOrNumberComparator(v1, v2, param1, param2);
     };
 
-    const NUMBER_COL_DEF = {
+    const METRIC_COLUMN = {
       flex: 1,
       minWidth: 120,
       maxWidth: 220,
-      type: "number",
```
JosephMarinier, line 144: That was adding a special formatter with a comma in the `utteranceCount` column (e.g. `1,000`). I propose to go with `1000` since:
* the rest of the the app doesn't use thousand separators;
* the comma is confusing to most of the world as it is usually used as the decimal separator; and
* the numbers should not get bigger than that (commas would be useful if say we get to 1,000,000 utterances, but we'll get bigger problems before then).

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -137,11 +137,10 @@ const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
       return gridStringOrNumberComparator(v1, v2, param1, param2);
     };
 
-    const NUMBER_COL_DEF = {
+    const METRIC_COLUMN = {
```
JosephMarinier, line 140: Just renaming to avoid confusion with `type: "number"` and since the metrics columns are not just numbers, but visual bars too.

webapp/src/components/Metrics/Metric.tsx
```diff
@@ -44,7 +43,15 @@ const Metric: React.FC<Props> = ({
 
   return (
     <Box display="flex" justifyContent="flex-end">
-      <Tooltip title={description} classes={{ popper: classes.popper }}>
+      <Tooltip
+        title={description
```
christyler3030, line 47: Nitpick: maybe move logic out of JSX?

JosephMarinier: I like it! Thanks!

christyler3030 (approved):

PR #203, opened 2022-08-03
Fix filtering aggregation modules without post-processing

### LINKED ISSUE #195: Utterance count is incorrect when post-processing is excluded
**Describe the bug**

When `exclude post-processing` is toggled on, the utterance count (parenthetical next to the Filters title) shows the count with `exclude post-processing` toggled off (i.e., it does not take this toggle into consideration).

**To Reproduce**

Select a post-processing-related filter (and possibly another to limit the number of utterances), toggle `exclude post-processing`, and observe that the count doesn't change and the number of utterances in the Utterances Table does not match the count. I would suggest selecting the rejection class as the prediction, and the outcome `Incorrect & Rejected`, since this should change with post-processing; this works nicely with CLINC-150.

**Expected behavior**

The utterance count should be correct!

**Vesion (please complete the following information):**
 - OS: MacOS Monterey 12.4
 - Azimuth version: 2.2.3

**Additional context**

NA

JosephMarinier:
Resolve #195

## Description:

Easy to reproduce with clinc_dummy by selecting the outcome=IncorrectAndRejected filter or the prediction=NO_INTENT filter. In those two examples, there should not be any `CorrectAndPredicted` in the distribution bars, nor in the confusion matrix. Filtering of all aggregation modules ignore `without_postprocessing`. It doesn't show in the confidence histogram since the "filtering"/graying out happens in the front end, but if I use the back end the same thing happens. It doesn't affect the utterance table as its filtering is not done in an aggregation module.
Image Pasted at 2022-8-3 06-08
Image Pasted at 2022-8-3 06-09

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/aggregation_module.py
+++ b/azimuth/modules/base_classes/aggregation_module.py
@@ -56,7 +56,12 @@ class FilterableModule(AggregationModule[ConfigScope], ExpirableMixin, ABC):
 
         """
         ds = super().get_dataset_split(name)
-        return filter_dataset_split(ds, filters=self.mod_options.filters, config=self.config)
+        return filter_dataset_split(
+            dataset_split=ds,
+            filters=self.mod_options.filters,
+            config=self.config,
+            without_postprocessing=self.mod_options.without_postprocessing,
+        )
 
     def _get_predictions_from_ds(self) -> List[int]:
         """Get predicted classes according to the module options (with or without postprocessing).
--- a/tests/test_modules/test_model_performance/test_metrics.py
+++ b/tests/test_modules/test_model_performance/test_metrics.py
@@ -231,6 +231,40 @@ def test_outcome_count_per_filter(tiny_text_config):
     assert res_post.count_per_filter.extreme_length != res.count_per_filter.extreme_length
 
 
+def test_outcome_count_per_filter_without_postprocessing(tiny_text_config):
+    save_predictions(tiny_text_config)
+    save_outcomes(tiny_text_config)
+
+    [res] = OutcomeCountPerFilterModule(
+        DatasetSplitName.eval,
+        config=tiny_text_config,
+        mod_options=ModuleOptions(
+            filters=DatasetFilters(outcome=[OutcomeName.IncorrectAndRejected]), pipeline_index=0
+        ),
+    ).compute_on_dataset_split()
+    for per_outcome in res.count_per_filter.outcome:
+        if per_outcome.filter_value != OutcomeName.IncorrectAndRejected:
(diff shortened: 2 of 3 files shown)


azimuth/modules/base_classes/aggregation_module.py
```diff
@@ -56,7 +56,12 @@ def get_dataset_split(self, name: DatasetSplitName = None) -> Dataset:
 
         """
         ds = super().get_dataset_split(name)
-        return filter_dataset_split(ds, filters=self.mod_options.filters, config=self.config)
+        return filter_dataset_split(
+            dataset_split=ds,
+            filters=self.mod_options.filters,
+            config=self.config,
+            without_postprocessing=self.mod_options.without_postprocessing,
```
JosephMarinier, line 63: That's what was missing

tests/test_modules/test_model_performance/test_metrics.py
```diff
@@ -231,6 +231,40 @@ def test_outcome_count_per_filter(tiny_text_config):
...
+        config=tiny_text_config,
+        mod_options=ModuleOptions(
+            filters=DatasetFilters(outcome=[OutcomeName.IncorrectAndRejected]),
+            pipeline_index=0,
+            without_postprocessing=True,
+        ),
+    ).compute_on_dataset_split()
+    for per_outcome in res_without_postprocessing.count_per_filter.outcome:
+        if per_outcome.filter_value != OutcomeName.IncorrectAndRejected:
+            assert per_outcome.utterance_count == 0, f"expected no {per_outcome.filter_value}"
```
JosephMarinier, line 265: Before the fix, this fails with `AssertionError: expected no CorrectAndPredicted`

gabegma: I'm afraid with time we will forget why we made this test - could we add a comment as to why we are only assessing other outcomes names than IncorrectAndRejected here? Something like: 
`# without_postprocessing, IncorrectAndRejected utterances from above will get other outcomes value. We verify that the results are still 0 for these outcomes.`

gabegma (approved):
Awesome work, thank you for adding a test. I only have a tiny comment as it took me a while to understand what we were testing.

tests/test_modules/test_model_performance/test_metrics.py
JosephMarinier: Cool! I had given it a try but I struggled finding how to explain it without writing a novel. Your suggestion helped me a lot. Thank you! Here is what I wrote: 8f1782820093a1319cb9f839fdd79b88b071a178

PR #209, opened 2022-08-05
Separate data and presentation

JosephMarinier:
Related to #179

## Description:

* Pass `AvailableDatasetSplits` to `DatasetSplitToggler` so it doesn't need to fetch `datasetInfo`. All 3 callers, or their callers, already had it at hand. That way, we don't need to mock the API to unit-test `DatasetSplitToggler`.
  * I'm inviting @nandhinibsn to review and rebase her work on top if this, as well as @christyler3030 as I know he'll love it. 😆 
* Clean up

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Controls/DatasetSplitToggler.tsx
+++ b/webapp/src/components/Controls/DatasetSplitToggler.tsx
@@ -15,36 +15,28 @@ const DatasetSplitToggler: React.FC<Props> = ({
   availableDatasetSplits,
   value,
   onChange,
-}) => {
-  const handleSelectionChange = (newValue: DatasetSplitName | null) => {
-    if (newValue !== null) {
-      onChange(newValue);
-    }
-  };
-
-  return (
-    <ToggleButtonGroup
-      color="secondary"
-      exclusive
-      fullWidth
-      size="small"
-      value={value}
-      onChange={(_, value) => handleSelectionChange(value)}
-    >
-      {DATASET_SPLIT_NAMES.map((datasetSplitName) => (
-        <ToggleButton
-          key={datasetSplitName}
-          sx={{ gap: 1, whiteSpace: "nowrap" }}
-          value={datasetSplitName}
-          disabled={!availableDatasetSplits?.[datasetSplitName]}
-        >
-          {datasetSplitName === "eval" && <SpeedIcon />}
-          {datasetSplitName === "train" && <ModelTrainingIcon />}
-          {DATASET_SPLIT_PRETTY_NAMES[datasetSplitName]} Set
-        </ToggleButton>
-      ))}
-    </ToggleButtonGroup>
-  );
-};
+}) => (
+  <ToggleButtonGroup
+    color="secondary"
+    exclusive
+    fullWidth
+    size="small"
+    value={value}
+    onChange={(_, value: DatasetSplitName | null) => value && onChange(value)}
+  >
+    {DATASET_SPLIT_NAMES.map((datasetSplitName) => (
+      <ToggleButton
+        key={datasetSplitName}
+        sx={{ gap: 1, whiteSpace: "nowrap" }}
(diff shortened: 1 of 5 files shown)


webapp/src/components/Controls/DatasetSplitToggler.tsx
```diff
@@ -1,50 +1,42 @@
...
+    size="small"
+    value={value}
+    onChange={(_, value: DatasetSplitName | null) => value && onChange(value)}
+  >
+    {DATASET_SPLIT_NAMES.map((datasetSplitName) => (
+      <ToggleButton
+        key={datasetSplitName}
+        sx={{ gap: 1, whiteSpace: "nowrap" }}
+        value={datasetSplitName}
+        disabled={!availableDatasetSplits?.[datasetSplitName]}
```
christyler3030, line 32: `!availableDatasetSplits.includes(datasetSplitName)` ?

webapp/src/components/Metrics/PerformanceAnalysis.tsx
```diff
@@ -76,11 +80,16 @@ type FilterByViewOption =
...
   jobId: string;
   pipeline: Required<QueryPipelineState>;
+  availableDatasetSplits: AvailableDatasetSplits | undefined;
 };
 
 type Row = MetricsPerFilterValue & { id: number };
 
-const PerformanceAnalysis: React.FC<Props> = ({ jobId, pipeline }) => {
+const PerformanceAnalysis: React.FC<Props> = ({
+  jobId,
```
christyler3030, line 89: If it makes sense we could think about a similar refactor for all query hooks that use jobId i.e. do all the fetching in one place and pass in the received values.

webapp/src/pages/Dashboard.tsx
```diff
@@ -90,7 +90,11 @@ const Dashboard = () => {
             />
           }
         >
-          <PerformanceAnalysis jobId={jobId} pipeline={pipeline} />
+          <PerformanceAnalysis
+            jobId={jobId}
+            pipeline={pipeline}
+            availableDatasetSplits={data?.availableDatasetSplits}
```
christyler3030, line 96: Everywhere else this is aliased to `datasetInfo`... maybe do the same here for readability?

christyler3030 (approved):
I _do_ love it.

webapp/src/components/Controls/DatasetSplitToggler.tsx
JosephMarinier: `availableDatasetSplits` is an object `{ train: boolean, eval: boolean }`, it doesn't have an `includes()` method.

webapp/src/pages/Dashboard.tsx
JosephMarinier: There are 7 files like that where we don't rename `{ data, isFetching, ... }`. I starting do that when the page/component does only one API call so there is no ambiguity about what data it is. If there are more than one calls, I usually rename them like `{ data: datasetInfo, isFetching: isFetchingDatasetInfo, ... }` and `{ data: countPerFilter, isFetching: isFetchingCountPerFilter, ... }`, but that sometimes results in very long names. An hybrid used here and there is to rename only `data`, as it is pretty vague, I must admit. Then `isFetching` and others are pretty clear. I'm happy to do that here, while we're at it, and keep it in mind next time I touch the 6 other places we use `data`. Makes sense?

webapp/src/components/Metrics/PerformanceAnalysis.tsx
JosephMarinier: Right, I always keep that in mind. Here we have the infamous situation where `pipeline` is required, and since the query hook can not be called conditionally inside the `Dashboard` page, it has to be called here.

PR #211, opened 2022-08-06
Cache tokenizer

JosephMarinier:
## Description:

Cache Hugging Face files when using Docker Compose. Hugging Face cache includes tokenizer in `transformers`, as well as `datasets`, `metrics` and `modules`. The default value is `~/.cache/hugginface`. Instead, we set it to `/azimuth_shr/huggingface_cache`, which is mounted.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,8 @@ services:
     image: ${REGISTRY}/azimuth:${BACKEND_VERSION}_${DEVICE}
     environment:
       - "CFG_PATH=${CFG_PATH}"
+      - "HF_HOME=/azimuth_shr/huggingface_cache"
       - "SENTENCE_TRANSFORMERS_HOME=/azimuth_shr/sentence_transformers"
-      - "TRANSFORMERS_CACHE=/azimuth_shr/sentence_transformers_cache"
       - "TFHUB_CACHE_DIR=/azimuth_shr/tf_hub"
       - "TH"
       - "TEMP"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       - "CFG_PATH=${CFG_PATH}"
       - "SENTENCE_TRANSFORMERS_HOME=/azimuth_shr/sentence_transformers"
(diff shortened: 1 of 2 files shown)


gabegma (commented):
Can you explain why is this a different path than the home - aren't they the same? Can we keep the same naming convention for the env var? `SENTENCE_TRANSFORMERS_X`?

docker-compose.yml
```diff
@@ -5,6 +5,7 @@ services:
     environment:
       - "CFG_PATH=${CFG_PATH}"
       - "SENTENCE_TRANSFORMERS_HOME=/azimuth_shr/sentence_transformers"
+      - "TRANSFORMERS_CACHE=/azimuth_shr/sentence_transformers_cache"
```
JosephMarinier, line 8: @gabegma In reverse order:
> Can we keep the same naming convention for the env var? `SENTENCE_TRANSFORMERS_X`?

I have no control over the name `TRANSFORMERS_CACHE`. It is defined [here](https://github.com/huggingface/transformers/blob/47e1676255e5dd86b9541f734cd4f4bdcbb50f4a/src/transformers/utils/hub.py#L94) on the Hugging Face side, and mentioned [here](https://huggingface.co/docs/transformers/installation?highlight=transformers_cache#cache-setup) in the documentation.

As for `SENTENCE_TRANSFORMERS_HOME`, I guess I don't have control, but to be honest, I don't know where it comes from. Do you?

> Can you explain why is this a different path than the home - aren't they the same?

I guess the two could have the same path value. I see some examples like [this](https://github.com/huggingface/hub-docs/blob/7c236b177759e523f6227367b33ac3f936b19638/api-inference-community/docker_images/sentence_transformers/Dockerfile#L20-L21) in Hugging Face documentation where the two have the same path value.

gabegma: Ah if this is standard, then that sounds good! Let's keep it as is.

gabegma (approved):
LGTM!

docker-compose.yml
JosephMarinier: Would you like to re-review with my updated solution? I updated the PR description for more details.

gabegma (approved):
LGTM!

PR #213, opened 2022-08-08
Fix high_epistemic_uncertainty smart tag

### LINKED ISSUE #181: Fix high_epistemic_uncertainty smart tag

gabegma:
Resolve #181 

## Description:
The smart tag wasn't showing in the UI because the tag was not saved under the 'main' key.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/indexable_module.py
+++ b/azimuth/modules/base_classes/indexable_module.py
@@ -2,6 +2,7 @@
 # This source code is licensed under the Apache 2.0 license found in the LICENSE file
 # in the root directory of this source tree.
 import abc
+import dataclasses
 from abc import ABC, abstractmethod
 from typing import Callable, Dict, List, Optional, Set, cast
 
@@ -211,6 +212,10 @@ class ModelContractModule(DatasetResultModule[ModelContractConfig], abc.ABC):
                     for idx, r in enumerate(res_casted)
                 }
                 dm.add_tags(high_epistemic, table_key=table_key)
+                # We also save with use_bma=False so the smart tag is saved in the
+                # main prediction table.
+                non_bma_table_key = dataclasses.replace(table_key, use_bma=False)
+                dm.add_tags(high_epistemic, table_key=non_bma_table_key)
             pred_tags = self.get_pred_tags(
                 label=dm.get_dataset_split(table_key=table_key)[self.config.columns.label],
                 model_predictions=dm.get_dataset_split(table_key=table_key)[
--- a/azimuth/modules/base_classes/indexable_module.py
+++ b/azimuth/modules/base_classes/indexable_module.py
@@ -204,6 +204,8 @@ class ModelContractModule(DatasetResultModule[ModelContractConfig], abc.ABC):
             )
 
             if table_key.use_bma:
+                # Epistemic uncertainty is only computed when use_bma is set to True, hence why the
+                # smart tag is only computed on that condition.
                 high_epistemic = {
                     idx: {
                         SmartTag.high_epistemic_uncertainty: r.epistemic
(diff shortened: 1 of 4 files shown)


config/development/clinc_dummy/conf_bma.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "CLINC Dummy",
```
JosephMarinier, line 2: Can we give a different name so that it's clear which cache is which? I don't know what "bma" means, so I don't know what case to use. That?
```suggestion
  "name": "CLINC Dummy BMA",
```

azimuth/modules/base_classes/indexable_module.py
```diff
@@ -211,6 +212,10 @@ def _save_result(self, res, dm):
                     for idx, r in enumerate(res_casted)
                 }
                 dm.add_tags(high_epistemic, table_key=table_key)
+                # We also save with use_bma=False so the smart tag is saved in the
+                # main prediction table.
```
JosephMarinier, line 216: Sorry, I am not familiar with that code. What is "the main prediction table"? Not the one independant of the pipelines?

gabegma: > the one independant of the pipelines

That would be the base table, not a prediction table. We have a table key to retrieve the relevant prediction table, according to the threshold, pipeline_idx, temperature, and if use_bma is set to True. The smart tag was only saved with the table key with use_bma=True. However, when the front end queries the utterances, it will retrieve the table associated with the table key for the current threshold, selected pipeline_idx, and current temperature. That is what I call main here, and the smart tag should be available in that table too. 

The use case for the front-end to ask for the prediction table with `use_bma=True` would be to get the predictions and confidence scores using BMA, which would be the average of the 20 iterations that were done for every data point. We don't support it yet in the UI, but it could be cool to do, and the back end already does it.

Let me know if you have more questions, and if you have recommendations to reformulate my comment in the code.

config/development/clinc_dummy/conf_bma.json
gabegma: Good catch!

azimuth/modules/base_classes/indexable_module.py
JosephMarinier: Thank you for the explanation.
I don't understand why this column is different from the others. What would be the problem if instead we removed the `if table_key.use_bma:` on line 206?

gabegma: We only compute the smart tag if the users set a number of iterations > 1 in the config (which will affect `table_key.use_bma`).

JosephMarinier: And `table_key.use_bma` is `True`, what adds the other columns (other smart tags, predictions, confidences, etc.) to the "main table" (since they don't have this `non_bma_table_key` shenanigan)? Will `_save_result` be called twice, with and without `table_key.use_bma`?

gabegma: That is a good point, I did some digging to figure it out. When the number of iterations is > 1 in the config, the prediction task does get called twice, once with `use_bma=False`, and once with `use_bma=True`. I don't think it's very clean how it's done at the moment (I don't like the name of the variable `use_bma`, and I think our approach should be more similar to other modules which can be deactivated, such as behavioral testing), but for now, at least this solves the bug. I can add an issue about refactoring this in the backlog? Let me know what you think.

JosephMarinier: Ok, it is starting to make sense in my mind. Thank you very much for explaining.
If the goal is to add the smart tag only if the number of iterations is > 1, but add it to both the "main" and the BMA tables (like the others tags/columns), would that more explicit code work as a solution?
```diff
-            if table_key.use_bma:
+            if self.config.uncertainty.iterations > 1:
                 high_epistemic = {
                     idx: {
                         SmartTag.high_epistemic_uncertainty: r.epistemic
                         > self.config.uncertainty.high_epistemic_threshold
                     }
                     for idx, r in enumerate(res_casted)
                 }
                 dm.add_tags(high_epistemic, table_key=table_key)
-                # We also save with use_bma=False so the smart tag is saved in the
-                # main prediction table.
-                non_bma_table_key = dataclasses.replace(table_key, use_bma=False)
-                dm.add_tags(high_epistemic, table_key=non_bma_table_key)
```

gabegma: The idea is clever, but it wouldn't work, because the epistemic uncertainty (`r.epistemic`) is just computed in the module when `table_key.use_bma` is set to `True`. So this implementation still gives 0 smart tag, because all `r.epistemic` are set to 0 when the module is called with `table_key.use_bma` is set to `False`. So we get the same issue as before.

gabegma: I added comments :)

JosephMarinier: Thank you very much for helping me understand, as well as the next dev who will read it. 👍

JosephMarinier (approved):

PR #214, opened 2022-08-10
Add tooltip if filter selector is unavailable

### LINKED ISSUE #194: Add tooltip to control panel sections when not available (no pipeline selected)
We could reuse and generalize the component `TabPipelineRequired`.

JosephMarinier:
Resolve #194

## Description:

3 commits to isolate from noise:
* Pass `isSuccess` to differentiate between successful data and data from before an error occurred.
* Add tooltip if filter selector is unavailable 
* Clean up unnecessary awaits

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Controls/FilterSelector.test.tsx
+++ b/webapp/src/components/Controls/FilterSelector.test.tsx
@@ -10,7 +10,8 @@ test("display loading state", async () => {
       searchValue=""
       selectedOptions={[]}
       handleValueChange={() => {}}
-      isFetching={true}
+      isFetching
+      isSuccess={false}
     />
   );
 
@@ -40,13 +41,16 @@ test("display reloading state", async () => {
           filterValue: "type1",
         },
       ]}
-      isFetching={true}
+      isFetching
+      isSuccess={false}
     />
   );
 
   await screen.findByText("type");
   await screen.findByRole("progressbar");
-  expect(await screen.queryByRole("figure")).not.toBeNull();
+  expect(screen.getByRole("figure").parentElement!.parentElement).toHaveStyle({
+    opacity: 0.38,
+  });
   expect(screen.getByLabelText("collapse-type")).not.toBeDisabled();
 });
 
@@ -59,6 +63,7 @@ test("display null state", async () => {
       selectedOptions={[]}
       handleValueChange={() => {}}
       isFetching={false}
+      isSuccess
     />
   );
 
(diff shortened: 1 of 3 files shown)


webapp/src/components/Controls/FilterSelector.test.tsx
```diff
@@ -89,15 +135,19 @@ test("empty outcome count", async () => {
...
+  screen.getByText("type");
+  const items = screen.getAllByRole("listitem");
   expect(items).toHaveLength(1);
 
-  const distribution = await screen.findByRole("figure");
-  expect(distribution).toHaveStyle("width: 0%");
+  const distribution = screen.getByRole("figure");
+  expect(distribution).toHaveStyle({ width: "0%" });
+  expect(distribution.parentElement!.parentElement).not.toHaveStyle({
+    opacity: 0.38,
```
christyler3030, line 149: Where does this number come from? Do we own it?

christyler3030 (approved):

webapp/src/components/Controls/FilterSelector.test.tsx
JosephMarinier: It comes from the default Mui theme, we did not overwrite it.
https://github.com/ServiceNow/azimuth/blob/cabf4c50643086887916881001d6c11d52a370b0/webapp/src/components/Controls/FilterSelector.tsx#L178-L180
I think it's fine to hardcode the value in the test. If ever one day we update Mui and the value happens to change, I think it's fine to have to adapt the test, and make sure the new value still makes sense.
That being said, I think that applies for the "positive" tests where I look for `toHaveStyle()`, but you made me realize that this `not.toHaveStyle()` is a bit more tricky. If the value changes, we won't notice it for this test, and the test will become useless. I experimented a bit more with the library, and the following works. It really does assert the `undefined`, and it doesn't just ignore it.
```diff
-  expect(distribution.parentElement!.parentElement).not.toHaveStyle({
-    opacity: 0.38,
+  expect(distribution.parentElement!.parentElement).toHaveStyle({
+    opacity: undefined,
```

JosephMarinier: Alternatively, I could make it more explicit by keeping the `opacity: 1` instead of `opacity: undefined`. I went for `undefined` just because I thought it was not necessary to "hardcode" `1`, but it's pretty straightforward and helps readability. What do you think? I think I like that.

JosephMarinier: Here is the latest suggestion: 7cae61f0861cd5cb30fd9e12c9de60044b40d3c9

PR #215, opened 2022-08-10
Test behavioral testing preview

### LINKED ISSUE #179: Add front-end tests for the performance analysis module

nandhinibsn:
Resolve #179

## Checklist:
Files added coverage:
1. PerturbationTestingPreview
2. mockAPIPerturbedUtterances

  our users.
* [ ] **DEV CHANGES.**


webapp/package.json
```diff
@@ -53,14 +54,29 @@
...
+        "statements": 75
+      }
+    }
+  },
   "engines": {
     "node": ">=14.17.5 <18.0.0"
   },
   "license": "Apache-2.0"
-}
+}
```
JosephMarinier, line 82: I think prettier is failing in CI because of this missing newline at the end of the file.

webapp/package.json
```diff
@@ -53,14 +54,29 @@
     "@types/react-dom": "^17.0.11",
     "@types/react-plotly.js": "^2.2.4",
     "@types/react-router-dom": "^5",
+    "canvas": "^2.9.3",
     "http-proxy-middleware": "^1.0.6",
+    "jest-canvas-mock": "^2.4.0",
```
JosephMarinier, line 59: I don't see those 3 new dependencies used. Can you remove them?

webapp/src/mocks/api/mockPertubationAPI.ts
```diff
@@ -0,0 +1,471 @@
...
+          },
+        },
+      ],
+      failureRates: { eval: 0.14, train: 0.16015625 },
+    };
+    return res(ctx.json(datasetInfoResponse));
+  }
+);
+
+export const getPertubationResponseWithnoTrainFailureRate = rest.get(
```
JosephMarinier, line 236: Detail, but I think `Withno` sould be `WithNo` or `Without`.
```suggestion
export const getPertubationResponseWithNoTrainFailureRate = rest.get(
```
or
```suggestion
export const getPertubationResponseWithoutTrainFailureRate = rest.get(
```

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  beforeAll(() => server.listen());
+  afterEach(() => server.resetHandlers());
+  afterAll(() => server.close());
+
+  it("should be having two toggle buttons for the train and test", async () => {
+    renderPerturbationTestingPreview({ train: true, eval: true });
+    await waitFor(() => {
+      expect(screen.getAllByRole("button")).toHaveLength(2);
+      expect(screen.getAllByRole("button")).not.toHaveLength(1);
```
JosephMarinier, line 37: That's redundant.
```suggestion
```

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+      // styling should be applied for the selected toggle as expected
+      expect(screen.getByText(/14.0%/)).toHaveStyle(
+        "font-size: 3.75rem; color: rgb(237, 108, 2)"
+      );
+
+      // selected toggle should be highlighted with some bgcolor
+      expect(screen.getByRole("button", { pressed: true })).toHaveStyle(
+        "color: rgb(156, 39, 176)"
+      );
```
JosephMarinier, line 72: That's OK, it doesn't hurt, but it doesn't seem necessary since it is mostly testing that Mui's `<ToggleButtonGroup color="secondary" ...` is working. We can leave that for Mui to test. Again, it's fine to include it, but general comment: let's not waste too much energy on re-testing Mui.

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+  renderWithRouterAndRedux(
+    <PerturbationTestingPreview
+      jobId="local"
+      pipeline={{ pipelineIndex: 0 }}
+      availableDatasetSplits={availableDatasetSplits}
+    />,
+    { route: "/local?pipeline_index=0", path: "/:jobId?:pipeline" }
+  );
+
+describe("Perturbation Testing Preview", () => {
```
JosephMarinier, line 25: Detail, but when possible, I call the test the same as the function/component we are testing, so it's easy to search for and things like that.
```suggestion
describe("PerturbationTestingPreview", () => {
```

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+      const columheaders = screen.getAllByRole("columnheader");
+
+      // verify if all the required columns are displayed
+      columheaders.forEach((item, index) => {
+        expect(item.textContent).toEqual(table_columns[index]);
+      });
+    });
+  });
+
+  it("should be having a default sort as desc for colum header 'FR on Evaluation/Training set'", async () => {
```
JosephMarinier, line 105:
```suggestion
  it("should have default sort as desc for column header 'FR on Evaluation/Training set'", async () => {
```

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+    renderPerturbationTestingPreview({ train: true, eval: true });
+    const columheaders = await screen.findAllByRole("columnheader");
+    const fr_column = columheaders.find(
+      (column) => column.ariaColIndex === "5"
+    );
+    fr_column && waitFor(() => expect(fr_column.ariaSort).toBe("descending"));
```
JosephMarinier, line 111: With `fr_column && `, if we don't find a column with `ariaColIndex === "5"`, we silently don't test anything. I guess that's not desired. Also, I'm sure we can get the right column header directly, without having to loop. Can you try that?

webapp/package.json
nandhinibsn: its not used part of these testing but we would require for the plotly.js

webapp/src/mocks/api/mockPertubationAPI.ts
nandhinibsn: okay I would go with this "getPertubationResponseWithoutTrainFailureRate"

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
nandhinibsn: yeah its just to test with unexpected values and see how the test reacts

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
nandhinibsn: yes MUI libraries should not be taken in to consideration.. the motive here is to test the background color.. but can be removed if thats part of mui. I assumed we had to do some css for it

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
nandhinibsn: make sense! will keep the consistent name as the component

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+      expect(screen.queryByText(/26.0%/i)).toBeNull();
+    });
+  });
+
+  it("should not display value if a datasplit is not available", async () => {
+    renderPerturbationTestingPreview({ train: false, eval: true });
+    await waitFor(() => expect(screen.getByText("--%")).toBeVisible());
+  });
+
+  it("should verify the perturbation preview table displayed with expected columns, rows, formatted contents, sort model", async () => {
```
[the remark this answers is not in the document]
nandhinibsn: okay.. I will give more attention on the description of the tests

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
nandhinibsn: Actually, we pass the API data in the beginning so it should have columns and rows but null check was actually required by the syntax rule here. But I dont like to do that as well. if we have a better approach to deal this, it would be great to know.

And I tried fetching like giving columnheader[index] but test could not find the expected index/header. so had to do loop and find the index.

webapp/package.json
JosephMarinier: I say we take one PR at a time.

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+        "Modif. Type",
+        "Test Description",
+        "FR on Evaluation Set",
+      ];
+      const columheaders = screen.getAllByRole("columnheader");
+
+      // verify if all the required columns are displayed
+      columheaders.forEach((item, index) => {
+        expect(item.textContent).toEqual(table_columns[index]);
+      });
```
JosephMarinier, line 101: This does not "verify if all the required columns are displayed" (and they aren't). It verifies that all displayed columns are in the expected `table_columns` (which should be camelCase by the way). We would need to switch on which array we are iterating:
```suggestion
      table_columns.forEach((textContent, index) => {
        expect(columheaders[index].textContent).toEqual(textContent);
      });
```
Which fails since the table only has the 3 first columns.
Also, here is a way to write the test without the need for
```
-      const columheaders = screen.getAllByRole("columnheader");
-      
```
```suggestion
      table_columns.forEach((name) => {
        screen.getByRole("columnheader", { name });
      });
```

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
JosephMarinier: Well the test passes because it doesn't test anything. Two reasons:
* `fr_column` is `undefined`
* You did not `await` the promise returned by `waitFor()`, so `() => expect(fr_column.ariaSort).toBe("descending")` would never be executed anyway.

So typing gets in the way here with `fr_column` being possibly `undefined`? You can assert that it is not `undefined` with the `!` post-fix "Non-null assertion operator".
```suggestion
    const columheaders = await screen.findAllByRole("columnheader");
    const fr_column = columheaders.find(
      (column) => column.ariaColIndex === "5"
    );
    await waitFor(() => expect(fr_column!.ariaSort).toBe("descending"));
```

You could also add a `expect(fr_column).toBeInTheDocument();` to make the error message clearer. But again, I'm sure we can get the right column header directly, without having to loop.

JosephMarinier: By the way, the trick of trying the test with an unexpected value would have worked to make sure it tests something. For example, you could have changed `"descending"` to `"potato"`, and the test would have still passed, indicating it is not doing its job.
```suggestion
    const columheaders = await screen.findAllByRole("columnheader");
    const fr_column = columheaders.find(
      (column) => column.ariaColIndex === "5"
    );
    fr_column && waitFor(() => expect(fr_column.ariaSort).toBe("potato"));
```

nandhinibsn:
Apologies for the confusion.. closing this one as there is another PR created with some rename of PR name https://github.com/ServiceNow/azimuth/pull/218

webapp/package.json
nandhinibsn: resolved

webapp/src/mocks/api/mockPertubationAPI.ts
nandhinibsn: resolved

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
nandhinibsn: removed "not" conditions. where ever is redundant resolved

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
nandhinibsn: yes we will let the MUI take care of such testing.resolved

PR #228, opened 2022-08-29
Add model comparison

### LINKED ISSUE #125: Being able to compare the performance analysis table with a second model

nandhinibsn:
Resolve #125 

## Checklist:

  our users.
* [x] **DEV CHANGES.**


webapp/src/components/Analysis/PreviewCard.tsx
```diff
@@ -6,10 +6,17 @@ import { Link } from "react-router-dom";
...
   to?: string;
+  linkButton?: string;
   description: React.ReactElement<typeof Description>;
 };
 
-const PreviewCard: React.FC<Props> = ({ title, to, children, description }) => (
+const PreviewCard: React.FC<Props> = ({
+  title,
+  to,
+  linkButton,
```
JosephMarinier, line 16: Since `linkButton` is optional, it should default to something. An empty button looks very weird.
image
I suggest leaving `"View details"` as the default and removing it where it is specified in `Dashboard`. Also, can we come up with a better name? Maybe `linkButtonText` or simply `buttonText`? What do you think?
```suggestion
  buttonText = "View details",
```

webapp/src/App.tsx
```diff
@@ -127,6 +128,14 @@ export default class App extends React.Component<Props> {
...
                               </BasicLayout>
                             </Route>
+                            <Route
+                              path="/:jobId/performance_analysis_comparison"
+                              exact
+                            >
+                              <BasicLayout>
+                                <PerformanceAnalysisComparison />
+                              </BasicLayout>
+                            </Route>
```
JosephMarinier, line 138: So this page is not necessarily about comparison, right? It works with a single model too, and might be useful to get more vertical space and see more rows. Therefore, I would remove the word "Comparison" everywhere, including file names.
```suggestion
                            <Route
                              path="/:jobId/performance_analysis"
                              exact
                            >
                              <BasicLayout>
                                <PerformanceAnalysis />
                              </BasicLayout>
                            </Route>
```

webapp/src/components/Metrics/PerformanceAnalysisComparisonTable.tsx
```diff
@@ -0,0 +1,567 @@
...
+        deltaPipelinePrecision: true,
+        comparedPipelineRecall: true,
+        deltaPipelineRecall: true,
+        comparedPipelineF1: true,
+        deltaPipelineF1: true,
+        comparedPipelineece: true,
+        deltaPipelineece: true,
+      });
+    }
+  }, [pipeline, comparedPipeline]);
```
JosephMarinier, line 444: It doesn't quite work as expected since you included an extra "Pipeline" in the metrics columns and used `ece` instead of `ECE` with the `delta` column.
That being said, the custom metrics should not be hard coded since the user can customize them. They should come from `Object.keys(metricInfo ?? {})`. Ideally, the outcomes would also come from `ALL_OUTCOMES`.
Finally, if we change
```diff
-        field: `${pipeline}ece`,
+        field: `${pipeline}ECE`,
```
it becomes uniform with the other columns.
Here is how I would do it:
```suggestion
    const visibility = comparedPipeline !== undefined;
    const columnVisibilityEntries = [
      "UtteranceCount",
      ...ALL_OUTCOMES,
      ...Object.keys(metricInfo ?? {}),
      "ECE",
    ].flatMap((field) => [
      [`comparedPipeline${field}`, visibility],
      [`delta${field}`, visibility],
    ]);
    setColumnVisibilityModel(Object.fromEntries(columnVisibilityEntries));
  }, [pipeline, comparedPipeline, metricInfo]);
```

webapp/src/components/Analysis/PreviewCard.tsx
nandhinibsn: I agree with the appropriate name change. But I am curious to know where is the empty button rendered.

webapp/src/components/Metrics/PerformanceAnalysisComparisonTable.tsx
nandhinibsn: that's a bug with the column names added for delta column sets I fixed yesterday. I like this solution of doing a flatmap and concatenating the string.  thanks for this suggestion.

webapp/src/components/Analysis/PreviewCard.tsx
JosephMarinier: Nowhere. I tested it locally.

nandhinibsn: You are right, the empty link button just flashes in a second before text renders. So, it would be right to set the default value and we don't have to pass every time "View details" text.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,533 @@
...
+            {config?.pipelines?.map(
+              (pipelineData, index) =>
+                index !== pipeline.pipelineIndex && (
+                  <MenuItem key={index} value={index}>
+                    {pipelineData.name}
+                  </MenuItem>
+                )
+            )}
+          </Select>
+        </FormControl>
```
JosephMarinier, line 492: First, I think we should create a reusable component `PipelineSelect` to make sure both look the same, and avoid duplicating that code about `No pipeline`. Don't worry about the `variant="standard"`, I think it's best to use it, like everywhere else in the app.
Second, I think you are looking for a `FormControlLabel`, not an `InputLabel`.
I think the result should look like
```suggestion
        <FormControlLabel
          control={
            <PipelineSelect
              value={comparedPipeline}
              onChange={setComparedPipeline}
              pipelines={config?.pipelines?.map(
                (_, index) => index !== pipeline.pipelineIndex
              )}
            />
          }
          label="Compared with"
          labelPlacement="start"
          sx={{ gap: 1 }}
        />
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,533 @@
...
+      {
+        field: `deltaece`,
+        headerName: "Delta",
+        width: 160,
+        align: "right",
+        valueGetter: ({ row }) =>
+          row.comparedPipeline &&
+          row.comparedPipeline.ece - row.basePipeline.ece,
+        renderCell: ({ value }: GridCellParams<number>) =>
+          value && (
```
JosephMarinier, line 381: If you return `0` instead of a `VisualBar`, it gives this weird result, so I think you should always render the `VisualBar`, even if `value` is `0`.
image

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,617 @@
+import {
+  Box,
+  Paper,
+  ListSubheader,
+  MenuItem,
+  Select,
+  Typography,
+  InputLabel,
```
JosephMarinier, line 8: This is not used. You can check the terminal for warnings like that.
```suggestion
```

webapp/src/pages/Dashboard.tsx
```diff
@@ -85,6 +85,8 @@ const Dashboard = () => {
       {isPipelineSelected(pipeline) && (
         <PreviewCard
           title="Performance Analysis"
+          to={`/${jobId}/performance_analysis${searchString}`}
+          linkButtonText="Compare models"
           description={
```
JosephMarinier, line 90: Can you reuse `performanceAnalysisDescription` from its page to make sure they stay in sync?
```suggestion
          description={performanceAnalysisDescription}
```

webapp/src/pages/Dashboard.tsx
```diff
@@ -85,6 +85,8 @@ const Dashboard = () => {
       {isPipelineSelected(pipeline) && (
         <PreviewCard
           title="Performance Analysis"
+          to={`/${jobId}/performance_analysis${searchString}`}
+          linkButtonText="Compare models"
```
JosephMarinier, line 89: With `pipelines` instead of `models`?
```suggestion
          linkButtonText="Compare pipelines"
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Sorry, "always" was not precise. I was talking about when the value is `0`. However, it can also be `undefined`, when the `comparedPipeline` is not yet loaded. JavaScript crashes right now when selecting the comparePipeline for the first time.
So, the type for all those `renderCell` functions should be `GridCellParams<number | undefined>` (adding ` | undefined`), and the body of the function should be preceded by `value !== undefined &&`. Like before, but with ` !== undefined` to deal properly with the case of `0`.

JosephMarinier: By the way, I preferred this `{ value }` destructuring assignment over the too-generic variable name `params`.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: yeah right! my future did this clean-up of removing unused imports

webapp/src/pages/Dashboard.tsx
nandhinibsn: Sure!. I can do it

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: > my future did this clean-up of removing unused imports

Sorry, what do you mean?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Okay, I don't find "PipelineSelect" in the code. Do you suggest creating one component for both places? I agree its a more efficient way of having it

JosephMarinier: Yes, I said
> we should create a reusable component `PipelineSelect`

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: You just went back to `value && ` in 1a5a7d16755ae114871981dde4dee2cf6b04d880. Please see [this comment](https://github.com/ServiceNow/azimuth/pull/228#discussion_r967281480).

nandhinibsn: Yes, I added back the undefined check in the previous PR. Otherwise, it throws an error and blocks the page from rendering.

I have also replaced params with {value} in the upcoming change.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Yep, I can create one.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: > You just went back to `value && ` in [1a5a7d1](https://github.com/ServiceNow/azimuth/commit/1a5a7d16755ae114871981dde4dee2cf6b04d880). Please see [this comment](https://github.com/ServiceNow/azimuth/pull/228#discussion_r967281480).

Yes, I removed 0 check but added undefined check in the next PR

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,608 @@
...
+          }
+        ></FormControlLabel>
+      </Box>
+      <Paper
+        variant="outlined"
+        sx={{
+          marginTop: 2,
+          padding: 3,
+        }}
+      >
```
JosephMarinier, line 543: Any reason for the outlined paper here? That's not the look we are going for in the rest of the app. In fact, we specifically remove the border of the `Table` here:
https://github.com/ServiceNow/azimuth/blob/d3c1e6b3647adf5bccb2e89565c3182d1ad99ec3/webapp/src/components/Table.tsx#L96
Can you change that for a `Box`?
```suggestion
      <Box marginTop={2} padding={3}>
```

nandhinibsn: Yes, right, we don't have borders in other places. I added this because of the design here https://www.figma.com/proto/FEndxTlShYTSUjMrInKqqn/ModelExplainer-V1?page-id=518%3A12557&node-id=1833%3A32699&viewport=1484%2C9621%2C0.5&scaling=min-zoom&starting-point-node-id=1833%3A32699.

nandhinibsn: > 

Okay, so I am just curious to learn here, so is it good to use a box instead paper mui?

nandhinibsn: > 

Understood. borders can be consistent like all other tables. it's for the design I added but not really necessary. what do you think?

nandhinibsn:
Final Look what I have for model comparison. Please provide your comments on the look @gabegma @JosephMarinier @lindsaydbrin 
image
image
image
image
image

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Yes, I think `Box` is better here, as `Paper` is meant to be the background of the app, like the sheet of paper we draw the app on.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,608 @@
...
+        <Table
+          sx={{
+            "& .MuiDataGrid-cell": {
+              borderBottom: "none",
+            },
```
JosephMarinier, line 548: I found an easy solution to the column borders not being continuous when the header is an icon, with `DATA_GRID_PROPS_DEFAULT_VALUES` `from "@mui/x-data-grid"`:
```suggestion
            },
            "& .MuiDataGrid-columnHeader": {
              // Fix line-height not affecting non-text headers like icons
              height: DATA_GRID_PROPS_DEFAULT_VALUES.headerHeight,
            },
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,608 @@
...
+
+    const styleDeltaValues = (value: number) =>
+      value >= 0 ? "delta-value-positive" : "delta-value-negative";
+
+    return [
+      {
+        id: 1,
+        field: "filterValue",
+        width: 200,
+        headerClassName: "sticky", // TODO doesn't work
```
JosephMarinier, line 284: Do you have some changes to push to fix the sticky corner cell?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Any progress on this? Blockers?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Yes, I have a workaround for this. as we discussed, the Grid Header component made the content sticky on the scroll.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,608 @@
...
+          }),
+    });
+
+    const utteranceCount: Column<Row>[] =
+      selectedMetricPerFilterOption === "label"
+        ? [
+            {
+              field: "utteranceCount",
+              headerName: "Total",
+              description: "Number of Utterances in each model",
```
JosephMarinier, line 248: There aren't really utterances "in" a model. Also, we prefer using "pipeline". That being said, none of the other columns have that kind of mention. It seems straightforward. I think we can remove it.
```suggestion
              description: "Number of utterances",
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Nice!.. I made some CSS to display the borders without breaks in between. So, what is this "DATA_GRID_PROPS_DEFAULT_VALUES"? is it a const?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: True!. The design suggested displaying it when the "Label" column is chosen. It was recommended to display in the header, but considering the width of the header, I ended up adding it as a tooltip.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Can you push it?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Yes, a const `DATA_GRID_PROPS_DEFAULT_VALUES` `from "@mui/x-data-grid"`

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,608 @@
+import {
+  Box,
+  Paper,
+  ListSubheader,
+  MenuItem,
+  Select,
+  Typography,
+  FormControlLabel,
```
JosephMarinier, line 8:
```suggestion
  Box,
  FormControlLabel,
  ListSubheader,
  MenuItem,
  Paper,
  Select,
  Typography,
```

nandhinibsn: They are ordered as per the alphabetical sequence. Please check my next PR for this change.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Sure. I will push it now

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: By adding this "DATA_GRID_PROPS_DEFAULT_VALUES.headerHeight" CSS property, I could still see the breaks between vertical borders next to the headers. When I investigated, I noticed that height defaults to "26px" for icon headers, whereas it's 56px for others. I assume it's because of the autoHeight?

JosephMarinier: I tested my suggestion and it worked. Maybe you have an issue with other changes you have locally?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Yes, I created a separate component for pipeline selection and used it in PageHeader and Performance table.

JosephMarinier: Could you check my suggestion again and try to move the `MenuItem`s inside `PipelineSelect.tsx` as shown? Instead of passing `pipelineMenuItem`, I suggested passing a simple array of pipelines, that I named `pipelines`. In `PageHeader`, that would simply be `pipelines={config.pipelines}`, and in `PerformanceAnalysisTable.tsx`, it would be something like
```ts
              pipelines={config?.pipelines?.filter(
                (_, index) => index !== pipeline.pipelineIndex
              )}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: I'm sorry, I was confused here by the two nested `Paper`s. I mean to comment on the second one. I'll write new comments...

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+                />
+              }
+            ></FormControlLabel>
+          </Box>
+        )}
+      </Box>
+      <Box
+        border={(theme) => `1px solid ${theme.palette.grey[200]}`}
+        sx={{ backgroundColor: (theme) => theme.palette.background.paper }}
+      >
```
JosephMarinier, line 580: This is the `Paper` I meant to remove, so the style matches the rest of the app.
```suggestion
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+    >
+      {errorFetchingComparedPipelineData.message}
+    </Typography>
+  ) : (
+    <Box
+      border={(theme) => `1px solid ${theme.palette.grey[200]}`}
+      height="100%"
+      padding={4}
+      marginTop={2}
+    >
```
JosephMarinier, line 542: And was perfect as a `Paper`, like in `PerturbationTestingSummaryTable` (with 4 of margin on top).
```suggestion
  <Paper variant="outlined" sx={{ height: "100%", padding: 4, marginTop: 4 }}>
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: I think you solved one, but there are 6 left. Reminder:
> So, the type for all those `renderCell` functions should be `GridCellParams<number | undefined>` (adding ` | undefined`), and the body of the function should be preceded by `value !== undefined &&`.

webapp/src/components/Metrics/DeltaComputationBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+        borderLeft={(theme) =>
+          `1px solid ${alpha(theme.palette.common.black, 0.2)}`
+        }
+        sx={{
+          zIndex: 1,
+          left: "50%",
+          transform: "translateX(-0.5px)",
+        }}
+      ></Box>
+      {value ? (
```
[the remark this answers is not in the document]
gabegma: I agree for 2!

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+      longHeader: React.ReactNode = shortHeader
+    ) => ({
+      ...(comparedPipeline === undefined
+        ? {
+            headerClassName: "no-border-right",
+            renderHeader: () => shortHeader,
+          }
+        : {
+            headerClassName: "no-border-right",
+            cellClassName: "no-border-right",
```
JosephMarinier, line 234: I think that should be the only place we use `no-border-right` class. You can simply remove it from the other 7 places. That way, the column borders extend all the way to the top of the table, like in the mockups.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+    </Link>
+  );
+
+  const LabelHeader = () => (
+    <Box
+      position="relative"
+      height={LABEL_HEIGHT}
+      width={LABEL_WIDTH}
+      top={LABEL_HEIGHT}
+      padding={(theme) => theme.spacing(1)}
```
JosephMarinier, line 477: That's equivalent to
```suggestion
      padding={1}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+      )}`}
+    >
+      <GridRow {...props} />
+    </Link>
+  );
+
+  const LabelHeader = () => (
+    <Box
+      position="relative"
+      height={LABEL_HEIGHT}
```
JosephMarinier, line 474: The height should be
```suggestion
      height={DATA_GRID_PROPS_DEFAULT_VALUES.headerHeight}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+      width={LABEL_WIDTH}
+      top={LABEL_HEIGHT}
```
JosephMarinier, line 476: Instead of offsetting the content with `top`, leaving a blank space above the table, I would use a negative `marginBottom`, like that:
```suggestion
      marginBottom={-DATA_GRID_PROPS_DEFAULT_VALUES.headerHeight}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+      padding={(theme) => theme.spacing(1)}
+      zIndex={(theme) => theme.zIndex.mobileStepper}
```
JosephMarinier, line 478: What is a `mobileStepper`? I think `1` is sufficient:
```suggestion
      zIndex={1}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+      bgcolor={(theme) => theme.palette.background.paper}
```
JosephMarinier, line 479: I would add the right border to match the rest of the column:
```suggestion
      bgcolor={(theme) => theme.palette.background.paper}
      className="MuiDataGrid-withBorder"
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+      {errorFetchingComparedPipelineData.message}
+    </Typography>
+  ) : (
+    <Box
+      border={(theme) => `1px solid ${theme.palette.grey[200]}`}
+      height="100%"
+      padding={4}
+      marginTop={2}
+    >
+      <Box marginBottom={1} display="flex" justifyContent="space-between">
```
JosephMarinier, line 543: Maybe we should increase the margin a little bit, now that we have some borders.
```suggestion
      <Box marginBottom={2} display="flex" justifyContent="space-between">
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+              position: "sticky",
+              left: 0,
+              background: (theme) => theme.palette.grey[200],
+            },
+            "& .no-border-right": {
+              borderRight: "none",
+            },
+            "& .border-left": {
+              borderLeft: (theme) => `1px solid ${theme.palette.grey[300]}`,
+            },
```
JosephMarinier, line 616: I prefer not to add this border, as in the mockups.
```suggestion
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+              overscrollBehaviorX: "contain",
+            },
+            "& .overall": {
+              borderRight: (theme) => `1px solid ${theme.palette.grey[300]}`,
+              background: (theme) => theme.palette.grey[200],
+            },
+            "& .delta": {
+              background: (theme) => theme.palette.grey[100],
+            },
+            "& .content-sticky": {
```
JosephMarinier, line 600: What do you mean by the word "content"? I am wondering if it shouldn't just be `sticky`:
```suggestion
            "& .sticky": {
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+            },
+            "& .MuiDataGrid-columnHeaders": {
+              borderBottom: "none",
+            },
+            "& .MuiDataGrid-virtualScroller": {
+              // Stops accidental navigation on horizontal scroll with touch pad
+              overscrollBehaviorX: "contain",
+            },
+            "& .overall": {
+              borderRight: (theme) => `1px solid ${theme.palette.grey[300]}`,
```
JosephMarinier, line 594: I don't think we want that
```suggestion
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+            value={selectedDatasetSplit}
+            onChange={setSelectedDatasetSplit}
+          />
+        </Box>
+        {config && (
+          <Box display="flex" flexDirection="row" alignItems="center">
+            <FormControlLabel
+              label={`Compare Baseline (${
+                config?.pipelines?.[pipeline.pipelineIndex].name
+              } ) with:`}
```
JosephMarinier, line 556: Let's remove that space:
```suggestion
              }) with:`}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+          }
+          getCellClassName={({ id, field }) =>
+            `${
+              id === OVERALL_ROW_ID
+                ? "content-sticky-overall"
+                : field.startsWith("delta")
+                ? "delta"
+                : ""
+            }`
+          }
```
JosephMarinier, line 635: 4 small things to simplify this bit:
- Move the `& .overall` definition down below the `& .content-sticky` definition, so it has priority
- Add `cellClassName: "delta"` to all delta columns (3 places)
- Remove `& .content-sticky-overall`
- Use only the `getCellClassName` to add the `overall` class, as follows:
```suggestion
          getCellClassName={({ id }) =>
            id === OVERALL_ROW_ID ? "overall" : undefined
          }
```
That's 11 lines shorter and easier to follow.

webapp/src/pages/PerformanceAnalysis.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box, Typography } from "@mui/material";
+import Description from "components/Description";
+import PerformanceAnalysisComparisonTable from "components/Metrics/PerformanceAnalysisTable";
```
JosephMarinier, line 3: We normally keep the same name:
```suggestion
import PerformanceAnalysisTable from "components/Metrics/PerformanceAnalysisTable";
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+
+    return [
+      {
+        id: 1,
+        field: "filterValue",
+        width: LABEL_WIDTH,
+        headerClassName: "no-border-right",
+        cellClassName: "content-sticky",
+        hideSortIcons: true,
+        disableColumnMenu: false,
```
JosephMarinier, line 264: I don't think those have any effect
```suggestion
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+          getCellClassName={({ id, field }) =>
+            `${
+              id === OVERALL_ROW_ID
+                ? "content-sticky-overall"
+                : field.startsWith("delta")
+                ? "delta"
+                : ""
+            }`
+          }
+          autoHeight
```
JosephMarinier, line 636: That doesn't work if we have a big number of classes and the table is too tall for the screen. You can test that by forcing a small height in the `BasicLayout`, for example:
```diff
-    <Box height="100%" display="flex" flexDirection="column" overflow="auto">
+    <Box height="500px" display="flex" flexDirection="column" overflow="auto">
```
So we have to remove that:
```suggestion
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Ho actually, we'll need to make the display flex, to support a big number of classes that make the table taller than the screen:
```suggestion
    <Paper
      variant="outlined"
      sx={{
        display: "flex",
        flexDirection: "column",
        height: "100%",
        marginTop: 4,
        padding: 4,
      }}
    >
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: I tried the solution of putting menuItem inside, but it won't work when we filter the pipeline, and the order of the index gets changed. We need to preserve the right pipeline index to fetch the API result. So, I pass the menuItem dynamically if I need to send filtered pipelines with their associated index.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: I meant future PR would remove this unused import. There is a typo here. Apologize for that.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: so, having an undefined check is good, but there is a scenario of an empty object [eg: customMetric: {}) to be noted as well. In such a scenario, it shows empty rows like the one below. Unless I go check myself, I did not know why the rows are empty here. 
image

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: I don't think it's required now. so we can resolve this

JosephMarinier: I think the borders must be added back to the headers. But good if this CSS is not necessary anymore.

webapp/src/components/PipelineSelect.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import { Select, MenuItem } from "@mui/material";
```
JosephMarinier, line 2:
```suggestion
import { Select, MenuItem } from "@mui/material";
import React from "react";
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+    React.useState<FilterByViewOption>("label");
+
+  const [comparedPipeline, setComparedPipeline] = React.useState<
+    number | undefined
+  >();
+
+  const { data: config } = getConfigEndpoint.useQuery(
+    { jobId },
+    { skip: jobId === undefined }
+  );
```
JosephMarinier, line 115: The type of `jobId` is `string`, so it (supposedly) can't be `undefined` and this `skip` seems irrelevant. Am I missing a special reason for including it? If so, a comment should explain it.
```suggestion
  const { data: config } = getConfigEndpoint.useQuery({ jobId });
```

webapp/src/components/PipelineSelect.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+type Props = {
+  selectedPipeline: number | string;
```
JosephMarinier, line 5: I would let `PipelineSelect` deal with the `""` for no pipeline (an implementation detail), by supporting `undefined` and simply removing the two `?? ""` where `PipelineSelect` is called.
```suggestion
  selectedPipeline: number | undefined;
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+            <FormControlLabel
+              label={`Compare Baseline (${
+                config?.pipelines?.[pipeline.pipelineIndex].name
+              } ) with:`}
+              labelPlacement="start"
+              sx={{ gap: 1, paddingRight: 2 }}
+              control={
+                <PipelineSelect
+                  selectedPipeline={comparedPipeline ?? ""}
+                  onChange={(value) => setComparedPipeline(value)}
```
JosephMarinier, line 562:
```suggestion
                  onChange={setComparedPipeline}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: I see. Interesting. I am not sure why `customMetrics` is an empty object if there are no utterances in the class, but we can easily work around that by defaulting to `NaN` instead of `undefined` in that case:
```diff
-            valueGetter: ({ row }) => row[pipeline]?.customMetrics[metricName],
+            valueGetter: ({ row: { [pipeline]: metrics } }) =>
+              metrics ? metrics.customMetrics[metricName] ?? NaN : undefined,
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: So, is the suggestion to remove the bgcolor but have the box?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: I reverted to paper, and it's good to have flex around.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+        headerClassName: "no-border-right",
+        cellClassName: "border-left",
+        headerAlign: "center",
+        valueGetter: ({ row }) =>
+          row.comparedPipeline &&
+          row.comparedPipeline.ece - row.basePipeline.ece,
+        renderCell: ({ value }: GridCellParams<number>) => (
+          <DeltaComputationBar
+            value={value}
+            formattedValue={value && (value as number).toFixed(2)}
```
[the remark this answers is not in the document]
JosephMarinier: Good point, but you are currently ignoring the `formattedValue ` if `value` is `0`, so it doesn't matter.
https://github.com/ServiceNow/azimuth/blob/d388ed031d4f9165b330bfeab6418cff6189ecec/webapp/src/components/Metrics/DeltaComputationBar.tsx#L31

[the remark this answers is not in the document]
JosephMarinier: If you change that behavior, and want `0` to give `"0"`, then I think you should convert it to a string too, like
```suggestion
            formattedValue={value === 0 ? "0" : value.toFixed(2)}
```

webapp/src/components/Metrics/DeltaComputationBar.tsx
```diff
@@ -0,0 +1,82 @@
+import React from "react";
+import { alpha, Box, Typography } from "@mui/material";
+import { motion } from "framer-motion";
+
+type Props = {
+  value: number;
+  formattedValue: string | number;
```
JosephMarinier, line 7: It feels error-prone to accept an unformatted number as `formattedValue`. I would prefer removing it:
```suggestion
  formattedValue: string;
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Yes right! Is it not a good practice to make use of mui themes here?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Good one to make use of DataGrid default height

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Perfect!. margin-bottom is more suitable, but it also needed to add padding for the box around the table.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Same question. is it not a good approach to use theme instead of constant numbers?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Good one!.. I can use the mui class name.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Content. I meant to write the "page content" for code readability.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Right!.. Agreed. It can be removed.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Ohh!!. Yes!..

webapp/src/pages/PerformanceAnalysis.tsx
nandhinibsn: Ohh!... that's a typo. You are right! It should be changed to "PerformanceAnalysisTable."

webapp/src/components/Metrics/DeltaComputationBar.tsx
nandhinibsn: I would say number type has its purpose here because "deltaUtteranceCount" is a number. but, we already send defined type values(number | string) from render cell.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Yes!. I had the same thought. Perfect!.. I can set it as NaN for such a case.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Right!.. I preferred adding the border-left looks aesthetically good. I have also posted a screenshot asking Karine's opinion. But, the mockup does not have a border-left.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: > But if the URL seems to be this way (maybe the user edits by themselves like this http://localhost:3005/undefined/performance_analysis?pipeline_index=0 ), the page gets blocked.

In that case, `jobId` is `"undefined"`, not `undefined`, so it still wouldn't skip. Its type is `jobId: string;`, so it's not possible that `jobId === undefined`.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: It is exactly the same. For `padding` and [all those CSS properties](https://mui.com/system/spacing/#api), specifying a number calls `theme.spacing()` behind the scene (the "Theme key" specified in that table).

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: If we were writing a `mobileStepper`, I guess it would be a good idea. I don't know what a mobile stepper is so that is probably not one.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: No, the suggestion is to remove the Box.

webapp/src/components/Metrics/DeltaComputationBar.tsx
JosephMarinier: Ah, interesting. I didn't think of `deltaUtteranceCount`. Maybe it would be better to force the caller to intentionally convert to string, to make sure one didn't forget the formatting and accidentally passed a number. 🤷‍♂️ I'm not sure. You decide.
```diff
-              formattedValue={value}
+              formattedValue={String(value)}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: ohh okay! nice info :) thank you

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: So here is the info for mobileStepper I found [mobileStepper for z-index](https://mui.com/material-ui/customization/z-index/#main-content).

JosephMarinier: Here is the [MobileStepper](https://mui.com/material-ui/react-stepper/#mobile-stepper) component. We are not using it.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Also, `LABEL_HEIGHT` is quite a bad name, since the select can be label, prediction and smart tags, not just label.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,658 @@
...
+          }),
+          ...pipeline,
+        }
+      )}`}
+    >
+      <GridRow {...props} />
+    </Link>
+  );
+
+  const LabelHeader = () => (
```
JosephMarinier, line 470: Same comment for `LabelHeader` that could simply be
```suggestion
  const Header = () => (
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,646 @@
...
+      ...metricsEntries.flatMap<Column<Row>>(
+        ([metricName, { description }]) => [
+          ...pipelines.map<Column<Row>>((pipeline) => ({
+            ...METRIC_COLUMN,
+            field: `${pipeline}${metricName}`,
+            ...groupHeader(pipeline, metricName),
+            description,
+            align: "right",
+            valueGetter: ({ row: { [pipeline]: metrics } }) =>
+              metrics?.customMetrics[metricName] ?? NaN,
```
JosephMarinier, line 348: The reason why in https://github.com/ServiceNow/azimuth/pull/228#discussion_r989101289 I suggested
```ts
metrics ? metrics.customMetrics[metricName] ?? NaN : undefined
```
and not
```ts
metrics?.customMetrics[metricName] ?? NaN
```
is because when `metrics` is `undefined`, it means the data is not loaded yet. In that case, we want the cell to be empty, like in the other columns. The only case where we want `NaN` (resulting in `--%`) is when `metrics` is loaded but `metrics.customMetrics` is an empty object (wich makes `metrics.customMetrics[metricName]` undefined).

So, please change
```suggestion
              metrics ? metrics.customMetrics[metricName] ?? NaN : undefined,
```
You could also add a comment if you think it needs clarifying.

webapp/src/components/PageHeader.tsx
```diff
@@ -154,24 +158,19 @@ const PageHeader = () => {
...
-                  {config?.pipelines?.map((pipeline, pipelineIndex) => (
-                    <MenuItem key={pipelineIndex} value={pipelineIndex}>
-                      {pipeline.name}
-                    </MenuItem>
-                  ))}
-                </Select>
+                {config && (
+                  <PipelineSelect
+                    selectedPipeline={pipeline.pipelineIndex ?? ""}
+                    onChange={(value) => setPipeline(value)}
```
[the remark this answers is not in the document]
JosephMarinier: Don't forget to simplify
```diff
-                    onChange={(value) => setPipeline(value)}
+                    onChange={setPipeline}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,646 @@
...
+
+          "& .overall": {
+            background: (theme) => theme.palette.grey[200],
+          },
+          "& .no-border-right": {
+            borderRight: "none",
+          },
+          "& .border-left": {
+            borderLeft: (theme) => `1px solid ${theme.palette.grey[300]}`,
+          },
```
JosephMarinier, line 612: Not used anymore
```suggestion
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Right!.. I believe we should have one name for the select column.

nandhinibsn: Considering the "DATA_GRID_PROPS_DEFAULT_VALUES.headerHeight" being constant already, I removed declaring already with const again. So there is no LABEL_HEADER in the code.

JosephMarinier: The values in that column are generally called "filter values". The field is `field: "filterValue"`. For reference, the type of each row is `MetricsPerFilterValue`.

nandhinibsn: Additionally, as the label name does not deliver a good message here, I changed the "LABEL_WIDTH" as "HEADER_WIDTH" as this CSS applies to header.

webapp/src/components/Icons/OutcomeIcon.tsx
```diff
@@ -10,7 +10,10 @@ const OutcomeIcon: React.FC<{ outcome: Outcome }> = ({ outcome }) => {
     <Tooltip title={OUTCOME_PRETTY_NAMES[outcome]}>
       <Icon
         fontSize="large"
-        sx={{ color: (theme) => theme.palette[OUTCOME_COLOR[outcome]].main }}
+        sx={{
+          color: (theme) => theme.palette[OUTCOME_COLOR[outcome]].main,
+          height: 56,
+        }}
```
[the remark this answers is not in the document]
nandhinibsn: Right! This height has been added to remove the break between borders in headers.  But adding "height: DATA_GRID_PROPS_DEFAULT_VALUES.headerHeight" to MuiDataGrid-column headers resolved it. thanks for the earlier suggestion.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: okay, I added it this way assuming "metrics?" will evaluate if a metric is undefined or not and then fetch custom metrics.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Right!. I cleaned up the unused lines, but they are yet to be pushed.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,642 @@
...
+          isLoading ||
+          isFetchingBasePipelineData ||
+          isFetchingComparedPipelineData
+        }
+        disableColumnMenu={false}
+        sortingOrder={["desc", "asc"]}
+        components={{
+          ColumnMenu,
+          Row: RowLink,
+          Header: Header,
```
JosephMarinier, line 635: Can be simplified:
```suggestion
          Header,
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,642 @@
...
+        sortModel={sortModel}
+        onSortModelChange={setSortModel}
+        columnVisibilityModel={columnVisibilityModel}
+        onColumnVisibilityModelChange={setColumnVisibilityModel}
+        getRowClassName={({ id }) =>
+          `${id === OVERALL_ROW_ID ? "overall" : ""}`
+        }
+        getCellClassName={({ id }) =>
+          `${id === OVERALL_ROW_ID ? "overall" : ""}`
+        }
```
JosephMarinier, line 621: The last point from my suggestion in https://github.com/ServiceNow/azimuth/pull/228#discussion_r988421070 is incomplete. The `getRowClassName` is not useful anymore. Also, it is unnecessary to wrap the ternary operator inside a `` `${...}` ``, so you can simplify.
```suggestion
        getCellClassName={({ id }) => (id === OVERALL_ROW_ID ? "overall" : "")}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -0,0 +1,642 @@
...
+                      </Box>
+                    </>
+                  )
+                : () => config?.pipelines?.[comparedPipeline].name,
+          }),
+    });
+
+    return [
+      {
+        id: 1,
```
JosephMarinier, line 254: `id` is not defined on `Column<Row>[]`. You'Re not using it anywhere, right?
```suggestion
```
I am not sure why typing is not catching this, but it does if we make this little change, which I would recommend:
```ts
-  const columns: Column<Row>[] = React.useMemo(() => {
+  const columns = React.useMemo((): Column<Row>[] => {
```

JosephMarinier: What do you mean by this?
> This line is carried from previous changes in "performance analysis."

webapp/src/components/Icons/OutcomeIcon.tsx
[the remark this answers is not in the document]
JosephMarinier: Don't forget this.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: Don't forget to also remove the unnecessary `` `${...}` ``.

webapp/src/components/Icons/OutcomeIcon.tsx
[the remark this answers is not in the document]
JosephMarinier: It didn't work because you used the `.MuiDataGrid-columnHeaders` class (with an `s`) instead of `.MuiDataGrid-columnHeader` as I suggested. I'll commit the necessary change.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: These are the changes from the "Performance analysis" table on the dashboard.

JosephMarinier: Can you show me what changes you're referring to? I don't see this `id: 1` anywhere in our code.

nandhinibsn: I meant that this line "const columns: Column<Row>[] = React.useMemo(() => {" is copied from "PerformanceTable" on the dashboard page. But I am curious to know the difference in pushing this "Column<Row>[]" definition inside.

JosephMarinier: > "const columns: Column[] = React.useMemo(() => {" is copied from "PerformanceTable"

Ho `PerformanceAnalysys` yes, indeed (not `PerformanceTable`).

I am not sure why the type inference doesn't work if we leave the type outside. But specifying it directly works and doesn't hurt, so let's do that.

nandhinibsn: Okay, Interesting! Thanks.

PR #230, opened 2022-09-02
Reorder confusion matrix so the rejection class is last

### LINKED ISSUE #139: Reorder the confusion matrix results so that the rejection class is always last.

gabegma:
Resolve #139 

## Description:
The order in the confusion matrix was determined by the class name orders, but it would look odd when the rejection class was not last. Now it is fixed, it will always be last.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


azimuth/modules/model_performance/confusion_matrix.py
```diff
@@ -31,13 +32,21 @@ def compute_on_dataset_split(self) -> List[ConfusionMatrixResponse]:  # type: ig
...
             y_pred=predictions,
             labels=class_ids,
             normalize="true" if self.mod_options.cf_normalized else None,
         )
+
+        # Put the rejection class last for the confusion matrix
+        rejection_idx = ds_mng.rejection_class_idx
+        max_idx = num_classes - 1
+        if rejection_idx != max_idx:
+            new_order = [x for x in range(0, max_idx + 1) if x != rejection_idx] + [rejection_idx]
```
JosephMarinier, line 48: You could avoid the `for` loop by using two ranges. Also, I think doing `max_idx = num_classes - 1` to then do `max_idx + 1` is more confusing than helpful. I would do
```suggestion
        if rejection_idx != num_classes - 1:
            new_order = [*range(0, rejection_idx), *range(rejection_idx + 1, num_classes), rejection_idx]
```

JosephMarinier: Ho! I see we already have a `class_ids` list. We can reuse it to avoid doing new ranges:
```suggestion
        if rejection_idx != num_classes - 1:
            new_order = class_ids[:rejection_idx] + class_ids[rejection_idx + 1:] + [rejection_idx]
```

gabegma: Ok, cool!!

gabegma: Is it more efficient than a for loop? Or more readable?

JosephMarinier: It is more efficient to avoid the loop. It is probably irrelevant as it take virtually no time anyway, but I find that code that looks sub-optimal is confusing. Like "Why do we need a loop? Am I missing something?". So yeah, realistically, it's more for readability.

JosephMarinier: As it was interesting, I wrote this little script to prove it, with an exaggerated `num_classes = 10_000_000`:
```python
import time
num_classes = 10_000_000
class_ids = list(range(num_classes))
rejection_idx = 12345678
times = []

start_time = time.time()
new_order = [x for x in range(0, num_classes) if x != rejection_idx] + [rejection_idx]
times.append(time.time() - start_time)

start_time = time.time()
new_order = [x for x in class_ids if x != rejection_idx] + [rejection_idx]
times.append(time.time() - start_time)

start_time = time.time()
new_order = [*range(0, rejection_idx), *range(rejection_idx + 1, num_classes), rejection_idx]
times.append(time.time() - start_time)

start_time = time.time()
new_order = class_ids[:rejection_idx] + class_ids[rejection_idx + 1 :] + [rejection_idx]
times.append(time.time() - start_time)

print(f"""
{times[0]} seconds for the loop on range(0, num_classes)
{times[1]} seconds for the loop on class_ids
{times[2]} seconds for two ranges
{times[3]} seconds for two slices of class_ids""")
```
I got
```
1.4199810028076172 seconds for the loop on range(0, num_classes)
1.0802741050720215 seconds for the loop on class_ids
0.6247909069061279 seconds for two ranges
0.4216649532318115 seconds for two slices of class_ids
```

gabegma:
@JosephMarinier - I added the class names - can you take over whenever you have the time to edit the front-end part? Thank you!

azimuth/modules/model_performance/confusion_matrix.py
```diff
@@ -31,13 +32,28 @@ def compute_on_dataset_split(self) -> List[ConfusionMatrixResponse]:  # type: ig
...
             y_pred=predictions,
             labels=class_ids,
             normalize="true" if self.mod_options.cf_normalized else None,
         )
+
+        class_names = ds_mng.get_class_names()
+
+        # Put the rejection class last for the confusion matrix
+        rejection_idx = ds_mng.rejection_class_idx
+        if rejection_idx != max(class_ids):
```
JosephMarinier, line 48: I am not too keen on the `max()` thing, as it is O(n) (looping through the whole list), whereas `num_classes - 1` is O(1) (constant). Again, it is unlikely that `class_ids` will get big enough for that to make a difference, but sub-optimal code always looks wrong to me.
```suggestion
        if rejection_idx != num_classes - 1:
```

JosephMarinier:
@christyler3030 or @nandhinibsn, could you please review the frontend part I did? :)

webapp/src/components/ConfusionMatrix.tsx
```diff
@@ -106,23 +106,32 @@ const ConfusionMatrix: React.FC<Props> = ({
...
+      ...postprocessing,
+    });
+
+  if (isLoading) {
+    return <Loading />;
+  } else if (error || data === undefined) {
+    return (
+      <Box alignItems="center" display="grid" justifyItems="center">
+        <img src={noData} width="50%" alt="no dataset info" />
+        <Typography>{error?.message || "Unexpected error"}</Typography>
```
christyler3030, line 125: For the generic error message could we have something like "An unknown error occurred"?  And maybe extract it to a constant that we can use everywhere?

JosephMarinier: Sure!

JosephMarinier: Done! [Extract `UNKNOWN_ERROR` as a constant](https://github.com/ServiceNow/azimuth/pull/230/commits/1d526bd48c28d0eeda954e4e94f3ceb29e9bd0e7)

christyler3030 (approved):

PR #233, opened 2022-09-02
Add and rename some data actions

### LINKED ISSUE #232: Add a proposed action: consider_merging_classes

gabegma:
Resolve #232 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Released changes are shown in the
 ### Added
 
 ### Changed
+- Added a proposed action: `merge_classes`, and rename `consider_new_class` to `add_new_class`.
 
 ### Deprecated/Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Released changes are shown in the
 ### Added
 
 ### Changed
-- Added a proposed action: `merge_classes`, and rename `consider_new_class` to `add_new_class`.
+- Added a proposed action: `merge_classes`, and rename `consider_new_class` to `define_new_class`.
 
 ### Deprecated/Breaking Changes
 
--- a/azimuth/routers/v1/tags.py
+++ b/azimuth/routers/v1/tags.py
@@ -24,22 +24,6 @@ router = APIRouter()
 TAGS = ["Tag v1"]
 
 
-def format_data_actions(tags) -> DataActionResponse:
-    formatted_data_actions = [
-        DataActionMapping(
-            relabel=x["relabel"],
-            add_new_class=x["add_new_class"],
-            merge_classes=x["merge_classes"],
-            remove=x["remove"],
-            augment_with_similar=x["augment_with_similar"],
-            investigate=x["investigate"],
-        )
-        for x in tags
-    ]
-
-    return DataActionResponse(data_actions=formatted_data_actions)
-
-
 @router.post(
     "",
     summary="Post data_action tags",
@@ -76,5 +60,4 @@ def post_data_actions(
         raise HTTPException(status_code=HTTP_400_BAD_REQUEST, detail=e.args[0])
     task_manager.clear_worker_cache()
     updated_tags = dataset.get_tags(list(request_data.data_actions.keys()), table_key=table_key)
-    formatted_tags = format_data_actions(updated_tags)
-    return formatted_tags
+    return DataActionResponse(data_actions=[DataActionMapping(**x) for x in updated_tags])
--- a/azimuth/routers/v1/tags.py
+++ b/azimuth/routers/v1/tags.py
@@ -28,7 +28,8 @@ def format_data_actions(tags) -> DataActionResponse:
     formatted_data_actions = [
         DataActionMapping(
             relabel=x["relabel"],
-            consider_new_class=x["consider_new_class"],
+            add_new_class=x["add_new_class"],
+            merge_classes=x["merge_classes"],
             remove=x["remove"],
             augment_with_similar=x["augment_with_similar"],
             investigate=x["investigate"],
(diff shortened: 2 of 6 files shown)


azimuth/types/tag.py
```diff
@@ -147,10 +149,11 @@ class Config:
...
                     1: {
                         "relabel": True,
-                        "consider_new_class": True,
-                        "remove": True,
-                        "augment_with_similar": True,
-                        "investigate": True,
+                        "add_new_class": False,
+                        "merge_classes": False,
+                        "remove": False,
+                        "augment_with_similar": False,
```
JosephMarinier, line 155: Tiny detail, but I notice that in most placer, you reordered `augment_with_similar` before those, so might as well do it here to for consistency:
```suggestion
                        "augment_with_similar": False,
                        "add_new_class": False,
                        "merge_classes": False,
                        "remove": False,
```

azimuth/routers/v1/tags.py
```diff
@@ -28,7 +28,8 @@ def format_data_actions(tags) -> DataActionResponse:
     formatted_data_actions = [
         DataActionMapping(
             relabel=x["relabel"],
-            consider_new_class=x["consider_new_class"],
+            add_new_class=x["add_new_class"],
+            merge_classes=x["merge_classes"],
             remove=x["remove"],
             augment_with_similar=x["augment_with_similar"],
             investigate=x["investigate"],
```
JosephMarinier, line 35: To avoid having to maintain that list, I think you can simply construct `DataActionMapping(**x)`. I tried and it passes the tests, including `test_post_tags` which seems to cover that adequately.

In fact, the funciton `format_data_actions()` gets so ridiculously small, that considering it is used only once, I would remove it. Also, I'm not sure we're really doing any "formatting" here, so I'm happy to remove the word. Something like that:

```diff
@@ -24,21 +24,6 @@ router = APIRouter()
 TAGS = ["Tag v1"]


-def format_data_actions(tags) -> DataActionResponse:
-    formatted_data_actions = [
-        DataActionMapping(
-            relabel=x["relabel"],
-            consider_new_class=x["consider_new_class"],
-            remove=x["remove"],
-            augment_with_similar=x["augment_with_similar"],
-            investigate=x["investigate"],
-        )
-        for x in tags
-    ]
-
-    return DataActionResponse(data_actions=formatted_data_actions)
-
-
 @router.post(
     "",
     summary="Post data_action tags",
@@ -75,5 +60,4 @@ def post_data_actions(
         raise HTTPException(status_code=HTTP_400_BAD_REQUEST, detail=e.args[0])
     task_manager.clear_worker_cache()
     updated_tags = dataset.get_tags(list(request_data.data_actions.keys()), table_key=table_key)
-    formatted_tags = format_data_actions(updated_tags)
-    return formatted_tags
+    return DataActionResponse(data_actions=[DataActionMapping(**x) for x in updated_tags])
```

docs/docs/key-concepts/proposed-actions.md
```diff
@@ -13,12 +13,13 @@ the [:material-link: Utterances Table](../user-guide/exploration-space/utterance
 Five proposed actions are currently supported:
 
 * **Relabel**: The label for this utterance should be changed.
-* **Consider New Class**: This data point may belong to a new class that is not available in the
+* **Augment with Similar**: This utterance lacks nearby utterances in the same class, and new,
+  similarly labeled utterances should be added.
+* **Add New Class**: This data point may belong to a new class that is not available in the
```
JosephMarinier, line 18: I know it's not new with this PR, but I realize that "not available" might be unclear. Like, is it just a question of enabling something? So maybe something like that?
```suggestion
* **Add New Class**: This data point may belong to a new class that doesn't exist in the
```

docs/docs/key-concepts/proposed-actions.md
```diff
@@ -13,12 +13,13 @@ the [:material-link: Utterances Table](../user-guide/exploration-space/utterance
...
   currently defined classes. Based on the number of data points identified with this action, a user
```
JosephMarinier, line 19: That nice wording makes me wonder if the data action would be better called `define_new_class`, since `add` might sound like adding a new class to the utterance, whereas `define` conveys that it is currently not defined. While we're here... What do you think?

JosephMarinier (commented):
Cool! I have a couple of comments, but it might also be good as is.

docs/docs/key-concepts/proposed-actions.md
lindsaydbrin: FWIW, I like "define new class," both because it is clear and because it avoids all the issues we've discussed about what "add" implies and whether it's clear it's just an annotation!

gabegma: Totally, define sounds better - thanks!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Released changes are shown in the
 ### Added
 
 ### Changed
+- Added a proposed action: `merge_classes`, and rename `consider_new_class` to `add_new_class`.
```
gabegma, line 16:
```suggestion
- Added a proposed action: `merge_classes`, and rename `consider_new_class` to `define_new_class`.
```

docs/docs/key-concepts/proposed-actions.md
```diff
@@ -13,12 +13,13 @@ the [:material-link: Utterances Table](../user-guide/exploration-space/utterance
 Five proposed actions are currently supported:
 
 * **Relabel**: The label for this utterance should be changed.
-* **Consider New Class**: This data point may belong to a new class that is not available in the
+* **Augment with Similar**: This utterance lacks nearby utterances in the same class, and new,
+  similarly labeled utterances should be added.
+* **Add New Class**: This data point may belong to a new class that doesn't exist in the
```
gabegma, line 18:
```suggestion
* **Define New Class**: This data point may belong to a new class that doesn't exist in the
```

JosephMarinier (approved):

PR #239, opened 2022-09-09
Azimuth for French with language selection in config

### LINKED ISSUE #265: Run Azimuth on French data with language selection in config
* Get Azimuth running on French data/pipelines
* Select language (and associated defaults) in the config

lindsaydbrin:
Resolve #265

### New news

No longer a draft PR; should run acceptably on French. Language-switching from config, updated behavioral tests, and multilingual FAISS encoder now included. Also more consistent reference to language-specific defaults in docs.

### Old news (for reference):

Draft PR because: 
* Goal was to get it working on a branch / not meant to be merged. 
* More work required or preferred:
  * Language-switching from config (not hard-coded French, as it is now) = required
  * Changes for behavioral tests
  * Changes for FAISS encoder

### New news

This code makes Azimuth work on French data/pipelines, and the language can be selected in the config.

Main changes:
* Part of Speech tagging: French defaults for spaCy model and dependency tags (to detect subjects and objects)
* Behavioral tests: French neutral tokens; fixed removal of space before punctuation (also submitted [nlpaug issue](https://github.com/makcedward/nlpaug/issues/313))
* Config: Dynamic selection of language-specific defaults in the config

Out of scope things that could improve French Azimuth:
* Multilingual FAISS encoder (English-trained option works acceptably but not optimally)
* Multilingual tokenizer (Only used for comparing average token count per utterance between train and eval, so should have minimal impact on any conclusions)

### Old news (for reference)

Addressed one aspect of Azimuth running for French models: spaCy part-of-speech tagging. This branch loads a French model (instead of English) and the `missing_xyz` smart tag POS/dependency tag lists have been updated to include French tags.

Several issues came up:
1. Single quote issue: Dataset included text with a single quote instead of an apostrophe, which broke POS-tagging (in particular, `j'` was not tagged as a subject) and search from the control panel. I added a new function to replace single quotes with apostrophes and used it both for searching and filtering. It worked reasonably well; out of 370 instances of `j'` in the eval set, 326 were tagged `missing_verb` before the fix, and 22 afterward (most of which started with "j'aimerais").
2. So then there was a problem with "j'" in "j'aimerais" being tagged as a subject (sometimes). This was solved with the `md` spaCy model (instead of `sm`).
3. SpaCy sometimes failed to identify verbs when they were the first word (common, especially with the imperative tense, e.g. "aidez-moi...") and/or capitalized (i.e. even in the middle of the sentence, which is not typical). Changing to the `md` model was a substantial improvement and fixed the majority of the issue. The exceptions seem to all be cases when the verb is capitalized (but not all capitalized verbs have problems; most don't). The capitalization issue seems to be fixed with the `lg` model...but it's like 500 MB...so I'm not sure if that's worth it (and haven't done it yet).

## Checklist:

  our users.
* [x] **DEV CHANGES.**


azimuth/utils/filtering.py
```diff
@@ -52,8 +53,10 @@ def filter_dataset_split(
     if len(filters.label) > 0 and config.columns.label in dataset_split.column_names:
         dataset_split = dataset_split.filter(lambda x: x[config.columns.label] in filters.label)
     if filters.utterance is not None and config.columns.text_input in dataset_split.column_names:
-        by = filters.utterance.lower()
-        dataset_split = dataset_split.filter(lambda x: by in x[config.columns.text_input].lower())
+        by = clean_utterance(filters.utterance.lower())
```
gabegma, line 56: Should we lowercase for POS tagging as well and include it in the `clean_utterance` function?

poetry.lock
```diff
@@ -1,14 +1,11 @@
 [[package]]
```
gabegma, line 1: It's not expected to have that much changes in the `poetry.lock`. Did you do `poetry update` instead of `poetry lock --no-update` followed by `poetry install`?

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
```diff
@@ -57,3 +57,53 @@ def test_syntax_tagging(simple_text_config):
...
+        DatasetSplitName.eval,
+        simple_text_config,
+    )
+
+    assert mod is not None
+    batch = {
+        DatasetColumn.idx: [0, 1, 2, 3],
+        "utterance": [
+            "adore les biscuits!",
+            "c'est terrible. C'est horrible pour moi d'ecrire ce test, mais je m'amuse bien.",
```
gabegma, line 73:
```suggestion
            "c'est terrible. C'est horrible pour moi d'écrire ce test, mais je m'amuse bien.",
```

gabegma (commented):
Cool, this gives un a good idea of the changes required! Some comments.

azimuth/utils/filtering.py
lindsaydbrin: I was thinking about doing that (as an option to the function), but then someone (@christyler3030 I think?) pointed out that the parser might be using it to indicate proper nouns, so I hesitated. It would help with the capitalized verb issue, though. And so far, we're not concerned with proper nouns per se, so in theory anything in the position of a proper noun should still be considered a noun. What do you think?

poetry.lock
lindsaydbrin: I don't _think_ I did `poetry update` but I probably did `poetry lock` without `--no-update`! How do I fix that? Can I delete the `poetry.lock` and re-lock with no update? Or pull the one from where I branched from main and then re-lock?

azimuth/utils/filtering.py
gabegma: Yeah, getting verbs right seems more important to me than nouns. Could it interfere though with how Spacy understands that something is a sentence?

poetry.lock
gabegma: Yes, the second one, although pick the current one on `main`, I just updated it. We should always start with the existing `poetry.lock`; otherwise, we lose all manual changes that were done to the `.lock` to fix certain library versions.

azimuth/utils/filtering.py
christyler3030: I don't totally understand how the tagger works but my thought was that *maybe* the tagger might get some information from the presence of a proper noun that might help in the rest of the sentence. However if I had to choose between getting verbs right ad maybe getting some indirect benefit from correctly tagging PROPN I would go for verbs.

lindsaydbrin: @gabegma I wouldn't think it'd be a problem. @christyler3030 Fair point. Ok, I'll try it out.

azimuth/utils/string_processing.py
```diff
@@ -0,0 +1,18 @@
+# Copyright ServiceNow, Inc. 2021 – 2022
+# This source code is licensed under the Apache 2.0 license found in the LICENSE file
+# in the root directory of this source tree.
+
+
+def clean_utterance(utterance: str, lowercase: bool = True):
```
JosephMarinier, line 6: Those are tiny details:
- I find there is a little mismatch between the super generic file name pertaining to a "string" and the quite more specific function name about "utterance". My gut feeling is that we can make the file specific to utterances. I see other existing functions that do some "string processing", but they are in defined in some specific files where they are relevant, like `to_camel_case(string: str) -> str` or `get_utterances_diff(original_utterance: str, perturbed_utterance: str) -> List[str]`. Maybe this file could be in the `ml/` directory& Like `azimuth/utils/ml/utterance.py`? What do you think?
- If we don't have a reason to use the `lowercase` flag, I would remove it. Worse case, we add it if we need it one day. Otherwise, it's dead code that will stay there forever, and one more "branch" to unit-test.
```suggestion
def clean_utterance(utterance: str):
```

lindsaydbrin: Fair points. I'll remove the lowercase parameter and rename to `utterance.py` (and we can rename it if anything else goes into it that necessitates making it more specific, etc.). I don't think it should be in the ML directory - looking at the other files there, they seem to actually be related directly to models/ML, and this is for POS-tagging and filtering. Sound ok to you?

JosephMarinier: Sounds perfect to me!

JosephMarinier: Perfect! Thank you!

azimuth/config.py
```diff
@@ -308,15 +372,45 @@ class SyntaxConfig(CommonFieldsConfig):
...
 class AzimuthConfig(
     PerturbationTestingConfig,
     SimilarityConfig,
     DatasetWarningConfig,
     SyntaxConfig,
+    LanguageConfig,
     extra=Extra.forbid,
 ):
+    # TODO @fred But _why_ should this class remain empty? :) Should it just be empty of data
+    #  attributes, and are methods ok?
```
JosephMarinier, line 390: I think we can move the validator in `LanguageConfig`, making it closer to where `values["language"]` is defined, and avoiding this question altogether.

azimuth/config.py
```diff
@@ -308,15 +372,45 @@ class SyntaxConfig(CommonFieldsConfig):
...
+            )
+        values["syntax"].subj_tags = (
+            values["syntax"].subj_tags or subj_tags_per_language[values["language"]]
+        )
+        values["syntax"].obj_tags = (
+            values["syntax"].obj_tags or obj_tags_per_language[values["language"]]
+        )
+        values["syntax"].spacy_model = (
+            values["syntax"].spacy_model or spacy_model_per_language[values["language"]]
+        )
```
JosephMarinier, line 412: I got an idea that would define the defaulting logic closer to where the fields are defined:
- Create a method in `SyntaxOptions`:
```py
    def set_defaults(self, language: SupportedLanguage):
        # To see all dep tag options for a spacy model: spacy.load(SPACY_MODEL).get_pipe("parser").labels
        if language == SupportedLanguage.en:
            self.subj_tags = self.subj_tags or ["nsubj", "nsubjpass"]
            self.obj_tags = self.obj_tags or ["dobj", "pobj", "obj"]
            self.obj_tags = self.obj_tags or "en_core_web_sm"
        elif language == SupportedLanguage.fr:
            self.subj_tags = self.subj_tags or ["nsubj", "nsubjpass"]
            self.obj_tags = self.obj_tags or ["obj", "iobj", "obl:arg", "obl:agent", "obl:mod"]
            self.obj_tags = self.obj_tags or "fr_core_news_md"
```
- And call it here:
```suggestion
        values["syntax"].set_defaults(values["language"])
```
- With that approach, I think we can get rid of the constant dictionaries that we were not sure where and how to define.

What do you think?

lindsaydbrin: Let me check that I understand what you're suggesting? So, the first part of `SyntaxOptions` would still be the same, where we set things to empty lists or strings, and the user can provide their own value(s). And we still have the `root_validator` doing the dynamic selection, but we're just replacing code that checks an external dictionary with code that runs an internal method? 

I'm assuming that by the time the root_validator is run, the empty list/string is considered falsy because we're not doing `pre=True`, so that part of the logic still works? I.e. the logic is written in the parent class, but run in the child class, so it's ok?

I like the idea of defining a `set_defaults` method in `SyntaxOptions` so that the logic is closer to where the defaults are set, as you said. My only question is about making it easy to see all the things that are changed with a different language. Take a look at the commented code and question at the top of the file - I'll tag you - and let me know what you think in the context of that suggestion? In theory, all the language-specific values could still be set in a (MEGA) dictionary like that, but we could still have a `set_defaults` method to make it more clear what's happening when you're reading the code.

azimuth/config.py
lindsaydbrin: I don't understand how it would have access/visibility to the other configs (`SyntaxConfig`, etc.) that are also parents of `AzimuthConfig`?

JosephMarinier: `AzimuthConfig` inherits from `LanguageConfig`, so it takes all its attributes and methods, including the validator. We only instantiate an `AzimuthConfig`, and never a `LanguageConfig` directly (in which case, you are right, it would be missing the other fields).

lindsaydbrin: Hmm. Ok, so it works if we do that, but is it confusing, since it runs when you run AzimuthConfig? Like, isn't it more clear what happens if it's in AzimuthConfig? Also, what problem are we solving by moving it?

JosephMarinier: Exactly, that's the language-related validator. I think it will be consistent with the other validators. For example, `ThresholdConfig` has a `root_validator` too. Other than that, I am not sure if it is a "problem", but I'm glad to avoid touching the `# This class should remain empty!` comment. 😅

lindsaydbrin: Lol, I was enjoying the impetuousness of it. 😂

Ok, I think your suggestion makes sense! But I'm still curious. @Dref360: Why does AzimuthConfig need to remain empty? Empty of every/anything?

JosephMarinier: Or are methods OK?

azimuth/config.py
```diff
@@ -19,6 +20,44 @@
...
+
+class LanguageDefaultValues(AliasModel):
+    suffix_list: List[str]
+    prefix_list: List[str]
+    spacy_model: str
+    subj_tags: List[str]
+    obj_tags: List[str]
+
+
+# To see all dep tag options for a spacy model: spacy.load(SPACY_MODEL).get_pipe("parser").labels
```
JosephMarinier, line 42: Should we move this comment over the `class SupportedSpacyModels` definition?

azimuth/config.py
```diff
@@ -131,12 +170,15 @@ class DatasetWarningsOptions(BaseModel):
 class SyntaxOptions(BaseModel):
     short_sentence_max_token: int = 3
     long_sentence_min_token: int = 16
+    spacy_model: str = ""  # Language-based default value set by AzimuthConfig
```
JosephMarinier, line 173: So, "set by AzimuthConfig" is already outdated. 😅 To avoid that, and because I don't think it's necessary, maybe we can remove that part of the comment. If a dev wants to know where/how the defaults are set, he can search for the field name. What do you think?
```suggestion
    spacy_model: str = ""  # Language-based dynamic default value
```

lindsaydbrin: Yes, perfect!

azimuth/config.py
lindsaydbrin: Hmm. I agree that it's awkward where it is. It's more related to the tags, though - do you think it works to put it above `subj_tags` in `LanguageDefaultValues`?

lindsaydbrin:
Just a note that I realized that I will have to update docs once the code is settled!

azimuth/config.py
```diff
@@ -19,6 +20,44 @@
...
+# spaCy models must be in pyproject.toml to be loaded by Azimuth
+class SupportedSpacyModels(str, Enum):
+    en_core_web_sm = "en_core_web_sm"
+    fr_core_news_md = "fr_core_news_md"
+
+
+class LanguageDefaultValues(AliasModel):
+    suffix_list: List[str]
+    prefix_list: List[str]
+    spacy_model: str
```
JosephMarinier, line 37: So here we want a `SupportedSpacyModels`, right? Not any `str`. Does that work?
```suggestion
    spacy_model: SupportedSpacyModels
```

azimuth/config.py
JosephMarinier, line 173: Similarly, does that work?
```suggestion
    spacy_model: Union[SupportedSpacyModels, ""] = ""  # Language-based default value set by AzimuthConfig
```

azimuth/config.py
JosephMarinier: But before using a new value there, the person will have to add the value to the `SupportedSpacyModels` enum, right? In that case, my understanding is that this information is useful when editing `SupportedSpacyModels`.

lindsaydbrin: You're right that that will likely be the most common use case. I was thinking that someone might want to add just a tag, to catch some other dependency in one of the smart tags. In practice that might be rare, but it's feasible.

lindsaydbrin: But I could go either way. Either way, it's not too far from the other spot in the code.

azimuth/config.py
```diff
@@ -311,11 +353,40 @@ class SyntaxConfig(CommonFieldsConfig):
...
+        values["syntax"].spacy_model = (
+            values["syntax"].spacy_model
+            or config_defaults_per_language[values["language"]].spacy_model
+        )
+        values["syntax"].subj_tags = (
+            values["syntax"].subj_tags or config_defaults_per_language[values["language"]].subj_tags
+        )
+        values["syntax"].obj_tags = (
+            values["syntax"].obj_tags or config_defaults_per_language[values["language"]].obj_tags
+        )
```
JosephMarinier, line 380: Detail, but with two variables, we can make the code a lot cleaner, with less repetition of the error-prone strings.
```suggestion
        defaults = config_defaults_per_language[values["language"]]
        if values["behavioral_testing"]:
            neutral_token = values["behavioral_testing"].neutral_token
            neutral_token.prefix_list = neutral_token.prefix_list or defaults.prefix_list
            neutral_token.suffix_list = neutral_token.suffix_list or defaults.suffix_list
        values["syntax"].spacy_model = values["syntax"].spacy_model or defaults.spacy_model
        values["syntax"].subj_tags = values["syntax"].subj_tags or defaults.subj_tags
        values["syntax"].obj_tags = values["syntax"].obj_tags or defaults.obj_tags
```

azimuth/config.py
JosephMarinier: Oh ok, I didn't get that. As you wish, then.

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -139,6 +140,7 @@ def remove_or_add_final_punctuation(
     # Replace with specified punctuation if last character is another punctuation sign.
     elif any([p in last_char for p in [".", "!", "?", ","]]):
         perturb_utt = original[:-1] + punctuation_sign
+        perturb_utt = re.sub(r" ([.!?,])", r"\1", perturb_utt)  # Remove any penultimate space
```
JosephMarinier, line 143: I think we can simply remove the trailing space from `original[:-1]` with `rstrip()`.
```suggestion
        perturb_utt = original[:-1].rstrip() + punctuation_sign
```

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +309,19 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
...
+        perturbed_utterance = re.sub(r" (['’]) ", r"\1", perturbed_utterance)
+        perturbed_utterance = re.sub(r"([$]) ", r"\1", perturbed_utterance)
+        # nac replaces apostrophes with quotes; revert to original.
+        perturbed_utterance = (
+            re.sub(r"’", r"'", perturbed_utterance) if "'" in original else perturbed_utterance
+        )
+        # nac removes space before the final punctuation; re-add them.
+        perturbed_utterance = (
+            perturbed_utterance[:-1] + " " + perturbed_utterance[-1]
+            if re.search(r" [.!?,]", original[-2:])
```
JosephMarinier, line 322: What does the `[-2:]` change?

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +309,19 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
         with contextlib.redirect_stdout(None):
             # While nlpaug fixes their useless print, we ignore it.
             perturbed_utterance = aug.augment(original, n=1)
-        # Issue 854 nac adds spaces between the hyphen and $ sign.
-        perturbed_utterance = perturbed_utterance.replace(" ' ", "'")
-        perturbed_utterance = perturbed_utterance.replace("$ ", "$")
+        # Issue 854 nac adds spaces around the quote/apostrophe and $ sign; remove them.
+        perturbed_utterance = re.sub(r" (['’]) ", r"\1", perturbed_utterance)
+        perturbed_utterance = re.sub(r"([$]) ", r"\1", perturbed_utterance)
```
JosephMarinier, line 314: You might have noticed that `$` is a special character in regular expressions. Inside a `[]`, special characters are not considered special (except `-`), so your regexp works. Another way to tell that you want a literal `$` is to escape it with a `\`.
```suggestion
        perturbed_utterance = re.sub(r"(\$) ", r"\1", perturbed_utterance)
```
By the way, if you're interested in regular expressions, here is another way to approach the problem:
We can match only the space, on the condition that it is preceded with a `$` (but not matching it), with a lookbehind non-capturing group. That way, we don't need to use a capturing group (which takes a little bit of memory and time), we simply replace the space with an empty string.
```suggestion
        perturbed_utterance = re.sub(r"(?<=\$) ", "", perturbed_utterance)
```
In this situation, however, I don't think it makes a difference. It's just cool to know.

azimuth/modules/dataset_analysis/syntax_tagging.py
```diff
@@ -81,10 +78,10 @@ def compute(self, batch: Dataset) -> List[TaggingResponse]:  # type: ignore
...
                     tag[SmartTag.short] = True
 
-                tokens = self.spacy_pos(utterance)
-                sub_toks = [tok for tok in tokens if (tok.dep_ in self.subj_tags)]
-                obj_toks = [tok for tok in tokens if (tok.dep_ in self.obj_tags)]
-                vrb_toks = [tok for tok in tokens if (tok.pos_ in self.verb_tags)]
+                tokens_doc = spacy_pos(clean_utterance(utterance))
+                sub_toks = [tok for tok in tokens_doc if (tok.dep_ in self.config.syntax.subj_tags)]
+                obj_toks = [tok for tok in tokens_doc if (tok.dep_ in self.config.syntax.obj_tags)]
+                vrb_toks = [tok for tok in tokens_doc if (tok.pos_ in self.verb_tags)]
```
JosephMarinier, line 84: So as of now, there is no difference in the verb tags between English and French? While we are at it, we might as well add them to the config too, no? WHat do you think?

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
```diff
@@ -57,3 +56,53 @@ def test_syntax_tagging(simple_text_config):
...
+    assert json_output[0].tags[SmartTag.no_subj]
+    assert not any([json_output[0].tags[SmartTag.no_obj], json_output[0].tags[SmartTag.no_verb]])
+
+    assert not any(
+        [
+            json_output[1].tags[SmartTag.no_subj],
+            json_output[1].tags[SmartTag.no_verb],
+            json_output[1].tags[SmartTag.no_verb],
+        ]
+    )
```
JosephMarinier, line 91: I don't see a huge advantage using `any()` instead of the following, which might even be easier to read/parse. That's a detail, though, so as you wish.
```suggestion
    assert not json_output[1].tags[SmartTag.no_subj]
    assert not json_output[1].tags[SmartTag.no_verb]
    assert not json_output[1].tags[SmartTag.no_verb]
```

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +309,19 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
...
+        # nac replaces apostrophes with quotes; revert to original.
```
JosephMarinier, line 315: So that's only in French, right? Worth mentioning in the comment? In case one day someone tries it with English and says "Ho this bit of code is not necessary anymore".

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +309,19 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
...
+        perturbed_utterance = (
+            re.sub(r"’", r"'", perturbed_utterance) if "'" in original else perturbed_utterance
+        )
+        # nac removes space before the final punctuation; re-add them.
+        perturbed_utterance = (
+            perturbed_utterance[:-1] + " " + perturbed_utterance[-1]
+            if re.search(r" [.!?,]", original[-2:])
+            else perturbed_utterance
+        )
```
JosephMarinier, line 324: Here is a try at reverting all changes other than the desired perturbation. I think it works. By the way, a nice optimisation is to "compile" the regular expressions only once, in a constant like `re_punctuation = re.compile(r"( *[.!?,'’] *)")`, and then use `re_punctuation` in the code instead of re-compiling the same regexp on every use.
```suggestion
        # nac removes space before the final punctuation; re-add them.
        punctuations = re_punctuation.split(original)
        rest = re_punctuation.split(perturbed_utterance)
        perturbed_utterance = "".join([p if i % 2 else rest[i] for i, p in enumerate(punctuations)])
```
Here is a little snippet to try it out:
```python
import re
re_punctuation = re.compile(r"( *[.!?,'’] *)")

original = "Allô ! Est-ce que c'est correct ça?"
perturbed_utterance = "Allô! Est-ce que c’est coreect ça?"

punctuations = re_punctuation.split(original)
rest = re_punctuation.split(perturbed_utterance)
"".join([p if i % 2 else rest[i] for i, p in enumerate(punctuations)])
# "Allô ! Est-ce que c'est coreect ça?"
```
Only "correct" has been perturbed into "coreect", and the other punctuation changes have been reverted.

JosephMarinier, line 324: Here is a try at reverting all punctuation/spacing changes with a single regular expression. I think it works. By the way, a nice optimization is to "compile" the regular expression only once and save it to a constant, instead of re-compiling the same regexp on every use. For example, at the top of the file, I would write:
```python
re_punctuation = re.compile(r"( *[.!?,'’$] *)")
```
And here is my suggestion that uses it:
```suggestion
        # nac removes space before the final punctuation; re-add them.
        # nac removes space before the final punctuation; re-add them.
        punctuations = re_punctuation.split(original)
        rest = re_punctuation.split(perturbed_utterance)
        perturbed_utterance = "".join([p if i % 2 else rest[i] for i, p in enumerate(punctuations)])
```
Here is a little snippet to try it out:
```python
import re
re_punctuation = re.compile(r"( *[.!?,'’$] *)")

original = "Allô ! Est-ce que c'est $42?"
perturbed_utterance = "Allô! Est-ce qui c ’ est $ 42?"

punctuations = re_punctuation.split(original)
rest = re_punctuation.split(perturbed_utterance)
assert "".join([p if i % 2 else rest[i] for i, p in enumerate(punctuations)]) == "Allô ! Est-ce qui c'est $42?"
```
Only "que" has been perturbed into "qui", and the other punctuation changes have been reverted.

azimuth/config.py
lindsaydbrin: Oh, yes, thanks for catching that!

azimuth/config.py
lindsaydbrin: That doesn't work...it doesn't like the empty string in the `Union[]`. This works and passes my test. Am I missing any reason why it's not a good idea?

```suggestion
    spacy_model: Union[SupportedSpacyModels, None] = None  # Language-based dynamic default value
```

azimuth/config.py
lindsaydbrin: Yes, great! Thanks. Did the same with `values["syntax"]` for kicks and also consistency. Plus it gets rid of a few lines of code because it makes things fit on one line.

azimuth/utils/ml/perturbation_functions.py
lindsaydbrin: (Sorry, unedited comment had a bunch of ?<->$ confusion because I was thinking about the other regex!)

azimuth/utils/ml/perturbation_functions.py
lindsaydbrin: It takes just the last two characters in the string (i.e., starting from -2, where -1 is the last one), to check if it's a space and then punctuation. I.e., if there was a space there, put it back in, and if not, keep `perturbed_utterance` as is.

But you're right that this only looks at any space before the final punctuation, and not if that pattern happens in the middle, so I need to figure out a better solution. I'm going to resolve this conversation and continue that part of it on the relevant thread below.

azimuth/utils/ml/perturbation_functions.py
lindsaydbrin: Oh, just moved on to your next comment to check for spaces around punctuation throughout. So don't respond to this yet. :)

azimuth/modules/dataset_analysis/syntax_tagging.py
lindsaydbrin: Yeah, no difference between English and French. I wondered about adding them, but hesitated before we need it. There are two different types of spaCy tags we're working with: part of speech tags (POS; noun, verb, proper noun, etc.) and dependency tags (relationships between words, hence subject- and object-related). For verb we're just using the first one - super simple - but for nouns types we need the second. (If we added a `missing_noun` smart tag, we could just use the POS tags.) So as it is, the dependency tags are in the config, but the POS ones aren't. I could add it - I'd probably add some comment about how to see the POS tags. On one hand, it's a little different from the others and may not be necessary for new languages, but on the other hand, maybe it just adds flexibility and maybe someone would want it. And on the first hand...there isn't really another POS tag that makes sense for verbs.

With that context - what do you think?

azimuth/config.py
Dref360: Empty of fields.

Modules are based on Scope, if you add fields to AzimuthConfig, it will depends on all scope and will be recomputed when any field is changed.

lindsaydbrin: Ok great - I'll edit the comment to be more clear! To be sure that I get it right, does it make sense to say 

```
# Do not add fields, as modules (scope-dependent) would be recomputed when a field is changed.
```
? (I was struggling to word it better than that without adding a second line to the comment.)

lindsaydbrin: I.e., instead of as it is (`# This class should remain empty!`)

azimuth/config.py
JosephMarinier: So that is equivalent to `Optional[SupportedSpacyModels]` and makes `None` a valid option. What we disliked about it is that later in the code, the type of `spacy_model` will be `Optional[SupportedSpacyModels]`, even though we know it can only be `SupportedSpacyModels`. For example, the type checking might complain about it being potentially `None` where it can't be, which would be annoying.

If it's not the case in the current situation, and `Optional` works well, awesome, you could indeed make it `Optional[SupportedSpacyModels]`.

Another option is to add something like `default_value = "default_value"` to `SupportedSpacyModels`, have
```suggestion
    spacy_model: SupportedSpacyModels = SupportedSpacyModels.default_value  # Language-based dynamic default value
```
And check for it with an `if else` in the validator.

azimuth/utils/ml/perturbation_functions.py
JosephMarinier: Correct. As soon as there is a `?` after the `(`, it's not a capturing group. I emphasized on the fact it was non-capturing to explain why it is more efficient to use it.
Your suggestion wouldn't work as it would remove the `$`. The `(?:)` is non-capturing, but it is still matching, whereas a lookbehind or lookahead doesn't include the `$` in the match. You can think of capturing as saving some "sub-matches" in memory to reuse it later, like with your `\1` replacement. That's what is a tiny bit more expensive ($, expensive, lol). Makes sense?
The basic non-capturing group is useful for making a string (more than one character) optional, without wasting memory to capture the group. For example,
```
(?:ex)?press
```
would match `express` or `press`, but not `xpress` nor `epress` (whereas the regexp `e?x?press` would)

azimuth/modules/dataset_analysis/syntax_tagging.py
JosephMarinier: Ah I see, thank you. We can cross the bear's skin when we get to the bear.

azimuth/utils/ml/perturbation_functions.py
JosephMarinier: Ho I made a wonderful discovery. In standard regular expressions, `\w` is equivalent to `[a-zA-Z0-9_]`. It matches the characters that can be part of a "word", including numbers and the underscore (I think it's a "word" in the sense of a variable name in code). So, `\w+` would match a word. Unfortunately, accents are a problem. But not in python! The `\w`
> includes most characters that can be part of a word in any language, as well as numbers and the underscore.

So! We could generalize our approach by splitting words and reverting all changes to the characters between words. No need to list special characters, that would support any punctuation automatically. That would also support multiple spaces in a row, even without punctuation, and even at the beginning or end of the utterance. Here is a new example with the perturbation `ô` -> `x`, proving that the `ô` was treated as a word character, and not a special character.
```python
import re
re_punctuation = re.compile(r"(\w+)")

original = " Allô ! Est- ce   que c'est $42? "
perturbed_utterance = "Allx! Est-ce que c ’ est $ 42?"
expected = " Allx ! Est- ce   que c'est $42? "

punctuations = re_punctuation.split(original)
words = re_punctuation.split(perturbed_utterance)
actual = "".join([words[i] if i % 2 else p for i, p in enumerate(punctuations)])

assert actual == expected, f'\n      expected: "{expected}"\n       but got: "{actual}"'
```

lindsaydbrin: Re `\w`, cool, I checked some characters that I've seen problems with in python in other contexts (struggling to remember what, though...), and it works as expected:
```
re.compile(r"(\w+)").split(" _ů_ : _œ_  _ø_")
```
gives
```
[' ', '_ů_', ' : ', '_œ_', '  ', '_ø_', '']
```

JosephMarinier: Awesome! Thanks for checking. Indeed, I have had problems with `œ` and `ø` in other contexts in JavaScript.

lindsaydbrin: Hmm. So there are two questions:

1. When the keyboard or random character augmenters swap characters, can they replace a letter with punctuation or a space? You wouldn't get `do not` -> `don't`, but could you get `do not` -> `do n't` (or, like, `do n$t` or anything else). 
2. Does it matter? By which I mean, actually getting the prediction means running the utterance through the usual pipeline preprocessing, which involves a tokenizer, so does anything change that will actually affect a prediction, or just the UI? 

for answers:
1. Yes. :( See below. However...the defaults for the augmenter include `include_special_char=True, include_numeric=True`, and I think we could consider setting both to `False`.
2. With the example above, the original becomes `["do", " ", "not"]` and the perturbed utterance becomes `["do", " ", "n", "$", "t"`] and then the line that puts the pieces together loses the end of the utterance. So yeah, it would be a problem. (It wouldn't matter if a space were added and then lost around punctuation or something like that.)

So this happens:
```
nac.KeyboardAug(min_char=4, aug_word_max=1, aug_char_p=0.1).augment("$pppp")
Out[120]: '$ ppp('
```

But if you do `include_special_char=False`, then that doesn't seem to happen anymore (n=a bunch of times in the Python console).

lindsaydbrin: So the question is - do we want to exclude special characters from the tests to make this particular formatting issue easier to solve? I think it's probably ok - I can think of $ and % that might be useful, but it might be more appropriate to test them specifically anyhow (like we're doing with punctuation). 

@gabegma do you have thoughts on this?

azimuth/config.py
lindsaydbrin: I did the following (took your suggestion but modified it somewhat), and left the validator as is, and it works for my test. Does it seem ok to you?

```
class SupportedSpacyModels(str, Enum):
    null_value = ""
    en_core_web_sm = "en_core_web_sm"
    fr_core_news_md = "fr_core_news_md"

...

class SyntaxOptions(BaseModel):
    short_sentence_max_token: int = 3
    long_sentence_min_token: int = 16
    spacy_model: SupportedSpacyModels = SupportedSpacyModels.null_value  # Language-based default
    subj_tags: List[str] = []  # Language-based dynamic default value
    obj_tags: List[str] = []  # Language-based dynamic default value
```

azimuth/utils/ml/perturbation_functions.py
JosephMarinier: > Does it matter? [...] just the UI?

Correct, I don't think it matters for the model, it's only for displaying in the UI, including the modifications (git-like +/- diff thingy) which we can only accurately compute if there is only one perturbation.

Another thing we could do is use this output to show in the UI, but use the raw perturbed utterance for the perturbation testing. That way, if our regexp messes up the utterance, at least the test has the correct result. What do you think? If we go that way, I would encourage doing this heavier refactor in a separate PR, to unblock this one.

azimuth/config.py
JosephMarinier: Makes sense to me!

JosephMarinier: I liked the idea of a name like `default_value` as it was telling the user what to expect. I don't think `null_value` conveys that as well.

JosephMarinier: Good idea for the empty string though. I just checked and it works.

If you are curious, that's because `SyntaxOptions` inherits from `str` before `Enum`, so it is `str.__bool__` that is used to determine the truthiness of `values["spacy_model"]` and `values["spacy_model"] = values["spacy_model"] or ...` works. If we were to remove the `str` inheritance, it would stop working.

lindsaydbrin: Hmm. I was specifically thinking something other than `default_value` because the default is actually the thing we replace it with later. Is there another name that sounds good to you? Or you still think `default...` is fine because it is the thing that leads to the default being implemented?

lindsaydbrin: `use_default` ftw

lindsaydbrin: Wait, bad idea, that implies a boolean. Going to do `default_indicator`. Let me know if you have a better idea!

lindsaydbrin: Back to `use_default` after discussion.

azimuth/utils/ml/perturbation_functions.py
lindsaydbrin: (For posterity, we're making the changes discussed and adding the augmenter arguments to avoid using special characters.)

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +323,17 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
         with contextlib.redirect_stdout(None):
             # While nlpaug fixes their useless print, we ignore it.
             perturbed_utterance = aug.augment(original, n=1)
-        # Issue 854 nac adds spaces between the hyphen and $ sign.
-        perturbed_utterance = perturbed_utterance.replace(" ' ", "'")
-        perturbed_utterance = perturbed_utterance.replace("$ ", "$")
+        # nac alters spacing around punctuation (e.g., around '; after $; before ?) and replaces
+        # apostrophes with quotes; revert to original.
+        re_punctuation = re.compile(r"(\w+)")
```
JosephMarinier, line 328: Can you move that to the top of the file, after the imports? It's like a constant. We never have to redefine it.

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +323,17 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
...
+        punctuations = re_punctuation.split(original)[::2]
+        words = re_punctuation.split(perturbed_utterance)[1::2]
+        perturbed_utterance = "".join(
+            [
+                punctuation + word
+                for punctuation, word in zip_longest(punctuations, words, fillvalue="")
+            ]
+        )
```
JosephMarinier, line 336: We don't need a list comprehension here, `str.join()` supports the generator directly.
```suggestion
        perturbed_utterance = "".join(
            punctuation + word
            for punctuation, word in zip_longest(punctuations, words, fillvalue="")
        )
```

docs/docs/reference/configuration/analyses/syntax.md
```diff
@@ -13,10 +19,18 @@ Some thresholds can be modified to determine what is considered a short or a lon
     class SyntaxOptions(BaseModel):
         short_sentence_max_token: int = 3 # (1)
         long_sentence_min_token: int = 16 # (2)
+        subj_tags: List[str] = []  # Language-based default value set by AzimuthConfig (3)
+        obj_tags: List[str] = []  # Language-based default value set by AzimuthConfig (4)
+        spacy_model: str = ""  # Language-based default value set by AzimuthConfig (5)
```
JosephMarinier, line 24: We need to update that

pyproject.toml
```diff
@@ -32,6 +32,8 @@ transformers = "4.16.2"
 sentence-transformers = "2.1.0"
 spacy = "3.1.5"
 en_core_web_sm = {url = "https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.1.0/en_core_web_sm-3.1.0.tar.gz#egg=en_core_web_sm"}
+fr_core_news_sm = {url = "https://github.com/explosion/spacy-models/releases/download/fr_core_news_sm-3.1.0/fr_core_news_sm-3.1.0.tar.gz#egg=fr_core_news_sm"}
```
JosephMarinier, line 35: Is that necessary? Or do we have to add `fr_core_news_sm` to `SupportedSpacyModels`?

azimuth/modules/dataset_analysis/syntax_tagging.py
```diff
@@ -81,10 +78,10 @@ def compute(self, batch: Dataset) -> List[TaggingResponse]:  # type: ignore
                 if len(tokens) <= syntax_options.short_sentence_max_token:
                     tag[SmartTag.short] = True
 
-                tokens = self.spacy_pos(utterance)
-                sub_toks = [tok for tok in tokens if (tok.dep_ in self.subj_tags)]
-                obj_toks = [tok for tok in tokens if (tok.dep_ in self.obj_tags)]
-                vrb_toks = [tok for tok in tokens if (tok.pos_ in self.verb_tags)]
+                tokens_doc = spacy_pos(clean_utterance(utterance))
```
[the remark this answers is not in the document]
lindsaydbrin: "document" - it's consistent with its usage in the spaCy docs (and also nlp generally in terms of the notion of document). I thought that would make it easier for anyone who was trying to work through the code. E.g., see the code example [here](https://spacy.io/usage/spacy-101#annotations).

azimuth/utils/ml/perturbation_functions.py
lindsaydbrin: On it.

docs/docs/reference/configuration/analyses/syntax.md
lindsaydbrin: Yup, docs commit just pushed!

pyproject.toml
lindsaydbrin: I don't understand the question. We have to download a model (to the poetry environment) to be able to load it...that needs to be happen before anything happening in the config or app itself, no?

lindsaydbrin: Oh oh oh...we're using the medium model now. Right. Yes, unnecessary! Will remove it.

lindsaydbrin: Good catch! Thanks.

JosephMarinier (approved):
Hooo yeah! 👍

lindsaydbrin:
@gabegma I'll wait for you to read docs (at your leisure - hope you enjoy nonfiction 😄 📖 ) before merging!

docs/docs/reference/configuration/language.md
```diff
@@ -0,0 +1,29 @@
...
+    ```python
+    from enum import Enum
+
+    class SupportedLanguage(str, Enum):
+        en = "en"
+        fr = "fr"
+
+    class LanguageConfig:
+        language: SupportedLanguage = SupportedLanguage.en
+    ```
```
gabegma, line 27: Could we add an example here of what this looks like in the config? e.g. "language":  "fr".

docs/docs/reference/configuration/project.md
```diff
@@ -136,7 +137,16 @@ follows:
...
 
+## Language
+
+:blue_circle: **Default value**: `"en"`
+
+The field `language` determines how Azimuth will process the dataset for language-related
+analyses. Azimuth currently supports English (`"en"`) and French (`"fr"`). More details are
+available in [Syntax Analysis](../../key-concepts/syntax-analysis.md) and
+[Behavioral Testing](../../key-concepts/behavioral-testing.md).
+
```
gabegma, line 151:
```suggestion
```

tests/conftest.py
```diff
@@ -137,6 +137,22 @@ def tiny_text_task_manager(tiny_text_config, dask_client):
...
+    return AzimuthConfig(
+        name="sentiment-analysis-french",
+        dataset=DATASET_CFG,
+        pipelines=[PIPELINE_CFG],
+        artifact_path=str(tmp_path),
+        batch_size=10,
+        model_contract="hf_text_classification",
+        saliency_layer="distilbert.embeddings.word_embeddings",
+        rejection_class=None,
+        behavioral_testing=SIMPLE_PERTURBATION_TESTING_CONFIG,
```
gabegma, line 151: This config will have the English prefixes/suffixes. I think we should test that the perturbation tests are using the french tokens when setting the language to `fr`.

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
```diff
@@ -57,3 +56,45 @@ def test_syntax_tagging(simple_text_config):
...
+    }
+    json_output = mod.compute(batch)
+    assert len(json_output) == 5
+
+    assert json_output[0].tags[SmartTag.no_subj]
+    assert not any([json_output[0].tags[SmartTag.no_obj], json_output[0].tags[SmartTag.no_verb]])
+
+    assert not json_output[1].tags[SmartTag.no_subj]
+    assert not json_output[1].tags[SmartTag.no_verb]
+    assert not json_output[1].tags[SmartTag.no_verb]
```
gabegma, line 87:
```suggestion
    assert not json_output[1].tags[SmartTag.no_obj]
```

gabegma (commented):
Awesome PR! And great work with the documentation. I have a few comments.

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -138,7 +142,7 @@ def remove_or_add_final_punctuation(
         perturbation_type = PerturbationType.Deletion
     # Replace with specified punctuation if last character is another punctuation sign.
     elif any([p in last_char for p in [".", "!", "?", ","]]):
```
JosephMarinier, line 144: I know this is not new, but... Am I crazy or is that equivalent to this?
```suggestion
    elif last_char in [".", "!", "?", ","]:
```

azimuth/utils/ml/perturbation_functions.py
JosephMarinier: @gabegma
> I've left the regex review to @JosephMarinier ;)

Haha, fair enough. Lindsay and I came up with it together, in this thread.

azimuth/utils/ml/perturbation_functions.py
```diff
@@ -307,9 +325,14 @@ def typo(original: str, config: PerturbationTestingConfig) -> List[PerturbedUtte
         with contextlib.redirect_stdout(None):
             # While nlpaug fixes their useless print, we ignore it.
             perturbed_utterance = aug.augment(original, n=1)
-        # Issue 854 nac adds spaces between the hyphen and $ sign.
-        perturbed_utterance = perturbed_utterance.replace(" ' ", "'")
-        perturbed_utterance = perturbed_utterance.replace("$ ", "$")
+        # nac alters spacing around punctuation (e.g., around '; after $; before ?) and replaces
+        # apostrophes with quotes; revert to original.
```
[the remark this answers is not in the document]
lindsaydbrin: There's no issue 854 in nlpaug, so I assumed this was an old issue number from some old repo. I can't find it anywhere, though. In any case, I'll add the correct nlpaug issue (313, which I submitted).

azimuth/utils/ml/perturbation_functions.py
lindsaydbrin: Yes...I think you're right. I think that if `last_char` was defined as more than one character, the logic would technically be different, but since `last_char` is just the last, single character, your suggestion is much much cleaner (and easier to read!). Not my code :) but happy to make the change!

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
lindsaydbrin: Oooh, good catch, thanks!

tests/conftest.py
lindsaydbrin: Oh, good call. I'll add that to the test, check that it fails, then remove this line and check that it passes.

tests/test_load_config.py
```diff
@@ -118,3 +120,26 @@ def test_pipeline_names():
...
         )
+
+
+def test_french_defaults_and_override():
+    subj_tags_potatoes = ["russet", "yukon_gold"]
+    cfg = AzimuthConfig(
+        **{
+            **DEFAULT_CONFIG,
+            **{"language": "fr", "syntax": {"subj_tags": subj_tags_potatoes}},
+        }
```
JosephMarinier, line 131: Let's clean up a couple fo `*`s.
```suggestion
        **DEFAULT_CONFIG,
        language="fr",
        syntax={"subj_tags": subj_tags_potatoes}},
```

azimuth/config.py
JosephMarinier: @Dref360, I would like to understand this better. Can you help us?
How would a field defined directly in `AzimuthConfig` affect the scopes? I thought `AzimuthConfig` inherited from all the scopes, not the other way around.

JosephMarinier: Summarizing the discussion I had with Dref offline:
If a module needs a field `potato` and you put `potato` directly in `AzimuthConfig`, then the module will have to "scope" the whole `AzimuthConfig` and recompute anytime any field changes (not just `potato`). That would not affect any of the other existing modules, nor would it particularly affect `potato`. Alternatively, if you define `potato` in a new class, say `PotatoConfig`, and have `AzimuthConfig` inherit from it, then your module can "scope" `PotatoConfig`.

azimuth/utils/ml/perturbation_functions.py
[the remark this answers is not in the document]
gabegma: Perfect!

tests/conftest.py
lindsaydbrin: For posterity: Discussed and decided to remove this line from this fixture and, separately, add code to the test in `test_load_config.py` to test that the french tokens are taken appropriately.

docs/docs/reference/configuration/language.md
lindsaydbrin: Oh, yeah, good call!

docs/docs/reference/configuration/project.md
lindsaydbrin: Ack...thank you!

tests/test_load_config.py
lindsaydbrin: `*` suggestion!

😆

lindsaydbrin: @JosephMarinier Does the same logic apply to lines 66 and 81?

lindsaydbrin: It doesn't work...it says, "a variable annotation cannot be combined with tuple unpacking. I am doing what you suggested minus the second `}` on line 130 (since it doesn't match anything before it anymore).

JosephMarinier: > Does the same logic apply to lines 66 and 81?

Yes. You have probably noticed that this was addressed on `main`, and just got merged into your branch.

You're correct, I included an extra `}` by mistake.

I am not able to reproduce the error you mention. I'll react out so we take a look together.

lindsaydbrin: All sorted, my syntax mistake, thanks!

poetry.lock
```diff
@@ -158,7 +158,7 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 dev = ["cloudpickle", "coverage[toml] (>=5.0.2)", "furo", "hypothesis", "mypy", "pre-commit", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six", "sphinx", "sphinx-notfound-page", "zope.interface"]
 docs = ["furo", "sphinx", "sphinx-notfound-page", "zope.interface"]
 tests = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six", "zope.interface"]
-tests-no-zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six"]
+tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "six"]
```
gabegma, line 161: Not a big problem for this PR, but are you still using poetry 1.2.0? I think you are reverting some changes here because you are using an older poetry version. Worth updating soon :)

gabegma (approved):
Awesome work!! Love the documentation :)

poetry.lock
lindsaydbrin: Yes! I can fix it. Update to newest with `poetry self update`? We don't have the version we're using in our docs.

azimuth/config.py
```diff
@@ -316,10 +368,27 @@ class AzimuthConfig(
     SimilarityConfig,
     DatasetWarningConfig,
     SyntaxConfig,
+    LanguageConfig,
     extra=Extra.forbid,
 ):
-    # This class should remain empty!
-    pass
+    # Before adding attributes: Remember that dependence on an attribute in AzimuthConfig will
+    # cause a module to include all other configs in its scope.
```
JosephMarinier, line 375: I'd change only one word to be clear that it won't automatically "cause" that, but it will "force" you to set the whole `AzimuthConfig` as the module's scope.
```suggestion
    # force a module to include all other configs in its scope.
```

azimuth/config.py
```diff
@@ -289,6 +335,12 @@ def check_pipeline_names(cls, pipeline_definitions):
         return pipeline_definitions
 
 
+class LanguageConfig(CommonFieldsConfig):
+    # Language configuration sets multiple config values; see reference dictionary for details
```
JosephMarinier, line 339: Now that we moved the `root_validator` away, maybe it's not so clear what the "reference dictionary" is?
```suggestion
    # Language configuration sets multiple config values; see `config_defaults_per_language` for details
```

JosephMarinier (approved):
Beautiful! Thank you!

poetry.lock
gabegma: I just checked and I don't think we are specifying a version in the doc. `poetry self update` might not work out of the box, but you can try it and follow the guidelines if it tells you that it can't update.

gabegma: Nevermind - I think you have the right version of poetry! Or anyhow I just did a `poetry lock` and it did the same changes for me. Sorry for the confusion!

azimuth/config.py
lindsaydbrin:
```suggestion
    # force a module to include all other configs in its scope.
```
Sounds good! Just going to use "a" instead of "the."

poetry.lock
lindsaydbrin: Ok, no problem! Also, I mis-read earlier - I _am_ actually using version `1.2.0`, but somehow I interpreted it like, "are you _still_ using..." So yes, we have the same version, and sorry, I could have clarified that to sort things out earlier!

azimuth/config.py
```diff
@@ -289,6 +335,12 @@ def check_pipeline_names(cls, pipeline_definitions):
         return pipeline_definitions
 
 
+class LanguageConfig(CommonFieldsConfig):
+    # Language configuration sets multiple config values; see `config_defaults_per_language` for details
```
lindsaydbrin, line 339:
```suggestion
    # Language config sets multiple config values; see `config_defaults_per_language` for details
```

poetry.lock
gabegma: You didn't misread, I edited my comment and removed `still` because it was wrong 😅

lindsaydbrin: Oh ha. Ok, we're all good now. :)

azimuth/config.py
JosephMarinier: The reason why I used "the" was to indicate that I am referring to **_the_** same module as the one with a "dependence on an attribute in AzimuthConfig". I didn't want to sentence to be understood as "dependence on an attribute [...] will force [another random] module to include all other configs in its scope".

Maybe I should have modified the first line
> Remember that if a module depends on an attribute in AzimuthConfig, it will need to include all other configs in its scope.

lindsaydbrin: Yes, I understand the intention exactly! When I read what I committed, I read it as exactly that, whereas with "the," the pronoun feels surprising. I'm having trouble figuring out how to explain the grammatical reasoning... 

Currently:
> Before adding attributes: Remember that dependence on an attribute in AzimuthConfig will cause a module to include all other configs in its scope.
It is clear to me that the first phrase ("dependence...AzimuthConfig") belongs to the "a module" mentioned later - that it's a specific module that is the topic (subject?) of the sentence. 

With _the_: 
> Before adding attributes: Remember that dependence on an attribute in AzimuthConfig will cause the module to include all other configs in its scope.
...I'm like, wait, what? Which module were we talking about? _Were_ we talking about a module already?

But, all that said, even if I had a sound grammatical argument, it doesn't matter if it is confusing to you (or any other user). So to be super clear, we could change it to what you suggested, with a slight rephrasing in line with your earlier suggestion, and one of mine to be concise:
> Reminder: If a module depends on an attribute in AzimuthConfig, the module will be forced to include all other configs in its scope.

What do you think? I'm not opposed to doing a new PR for one comment, all in the name of clear communication. :)

JosephMarinier: That's perfect! PR: #310

PR #242, opened 2022-09-09
Reorder confusion matrix based on reverse cuthill mckee

### LINKED ISSUE #140: Can we order rows/columns in the confusion matrix in terms of meaningful confusion clusters.

gabegma:
Resolve #140 

## Description:
Reorder confusion matrix by default with reverse cuthill mckee, so that classes with confusion are closer to one another in the matrix. 

Still **TODO**:

- [x] A toggle needs to be added in the UI to preserve the original class order if that is preferred.
- [x] The documentation also needs to be updated, but I'll wait for the UI to take the screenshot.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


CHANGELOG.md
```diff
@@ -13,8 +13,8 @@ Released changes are shown in the
 ### Added
 
 ### Changed
-- The rejection class is always the last row/column in the confusion matrix.
 - Added a proposed action: `merge_classes`, and rename `consider_new_class` to `define_new_class`.
+- The order of the classes in the confusion matrix is now smarter: classes where the model gets confused will be closer to one another. The rejection class is always the last row/column in the confusion matrix. A toggle allows the user to keep the original order from the dataset if preferred.
```
lindsaydbrin, line 17: This might help the wording (not important for changelog, but to make things easier when you get to the docs) - what do you think?
```suggestion
- The order of the classes in the confusion matrix is now smarter: classes where the model gets similarly confused will be closer to one another. The rejection class is always the last row/column in the confusion matrix. A toggle allows the user to keep the original order from the dataset if preferred.
```

azimuth/modules/model_performance/confusion_matrix.py
```diff
@@ -34,26 +41,58 @@ def compute_on_dataset_split(self) -> List[ConfusionMatrixResponse]:  # type: ig
...
+            cf_no_rejection = np.delete(
+                np.delete(cf_normalized, rejection_idx, 0), rejection_idx, 1
+            )
+
+            # Get order based on reverse_cuthill_mckee algorithm
+            order_no_rejection = reverse_cuthill_mckee(
+                csr_matrix(cf_no_rejection >= MIN_CONFUSION_CUTHILL_MCKEE)
+            )
+
+            # Get right class indices and add rejection_idx at the end.
```
lindsaydbrin, line 74: I wasn't sure what "right" meant at first, and thought it would be helpful to indicate that it was the indices for the reduced-bandwidth matrix - but there may be a better way to word it than how I suggested!
```suggestion
            # Get class indices by reduced-bandwidth order and add rejection_idx at the end.
```

lindsaydbrin: Could take this approach, too:
```suggestion
            # Get class indices (ordered for plotting) and add rejection_idx at the end.
```

azimuth/modules/model_performance/confusion_matrix.py
```diff
@@ -34,26 +41,58 @@ def compute_on_dataset_split(self) -> List[ConfusionMatrixResponse]:  # type: ig
...
+            classes_no_rejection = np.delete(class_ids, rejection_idx)
+            order = np.append(classes_no_rejection[order_no_rejection], rejection_idx)
+
+            # Put the rejection class last for the confusion matrix
+            if rejection_idx != order[-1]:
+                rejection_position = np.where(order == rejection_idx)[0][0]
+                # From the rejection position, roll backwards all values
+                order[rejection_position:] = np.roll(order[rejection_position:], -1)
```
lindsaydbrin, line 82: This confuses me (is it outdated code?):
1. Wouldn't `rejection_idx` always `== order[-1]` because of the code above, where you just set it to be there?
2. Didn't we decide not to roll values, but just do what you've done above, which is apply the algorithm to the matrix without the rejection class, and then append it?

azimuth/types/model_performance.py
```diff
@@ -118,6 +118,8 @@ class ConfusionMatrixResponse(ModuleResponse):
     confusion_matrix: Array[float] = Field(..., title="Confusion Matrix normalized")
     class_names: List[str] = Field(..., title="Ordered class names for the confusion matrix")
     normalized: bool = Field(..., title="Normalized state of Confusion Matrix")
+    preserved_class_order: bool = Field(..., title="State of the order of the Confusion Matrix")
+    rejection_class_position: int = Field(..., title="Position of the rejection class")
```
JosephMarinier, line 122: Did you see a problem with using "index" instead of "position"? It seems to be the vocabulary we are currently using in the API, with `pipeline_index`. To me, "position" could be confused with a one-based position, with the 1st position being the index 0.
```suggestion
    rejection_class_index: int = Field(..., title="Index of the rejection class")
```

azimuth/routers/v1/model_performance/confusion_matrix.py
```diff
@@ -40,12 +40,14 @@ def get_confusion_matrix(
     pipeline_index: int = Depends(require_pipeline_index),
     without_postprocessing: bool = Query(False, title="Without Postprocessing"),
     normalized: bool = Query(True, title="Normalized"),
+    preserved_class_order: bool = Query(False, title="Preserved Class Order"),
```
lindsaydbrin, line 43: You might consider renaming `preserved_class_order` to `class_order_preserved` to make it more clear that it's a boolean, and also to be consistent with the boolean `cf_normalized`. As is, to me at least, it implies that it's the original class order itself (i.e., a "class order").

azimuth/types/model_performance.py
gabegma: So the reason is that it's no longer the index of the rejection class, but its position in the matrix -> the indices are no longer ordered.

lindsaydbrin: I don't have an opinion about "position," but I think "index" overlaps with the actual index of the class (like, as specified by the user), which is different from its position in the confusion matrix.

lindsaydbrin: Oh, sorry, didn't see Gab's comment before I wrote mine! Exactly.

tests/test_modules/test_model_performance/test_confusion_matrix.py
```diff
@@ -76,9 +87,22 @@ def test_confusion_matrix_reorder_rejection_class(guse_text_config):
     dm = mod.get_dataset_split_manager()
 
     # Assert that the rejection class in not the last one already
```
lindsaydbrin, line 89:
```suggestion
    # Assert that the rejection class is not the last one already
```
Sorry, out of scope, but really small typo since you're working in the file...

tests/test_modules/test_model_performance/test_confusion_matrix.py
```diff
@@ -76,9 +87,22 @@ def test_confusion_matrix_reorder_rejection_class(guse_text_config):
...
+    # Preserving the class order from the dataset
+    mod_preserved_order = ConfusionMatrixModule(
+        DatasetSplitName.eval,
+        guse_text_config,
+        mod_options=ModuleOptions(pipeline_index=0, cf_preserved_class_order=True),
+    )
+    [json_output_preserved_order] = mod_preserved_order.compute_on_dataset_split()
+    assert json_output_preserved_order.class_names == dm.get_class_names()
+    # The rejection class should not be last
+    assert json_output_preserved_order.rejection_class_position == dm.rejection_class_idx
```
lindsaydbrin, line 108: Oh, interesting. Am I understanding this correctly? 

Originally we did:
1. Preserve class order (but move rejection class last)

And now our options are:
1. Preserve class order (do not put rejection class last)
2. Re-order with Cuthill-McKee (rejection class last)
?

I am thinking that it makes sense for 1. to stay as it was before even though I know it changes the logic somewhat (so it's, like, "preserve non-rejection class order"). What do you think? (And/or am I understanding correctly?)

azimuth/routers/v1/model_performance/confusion_matrix.py
JosephMarinier, line 43: Est-ce que ça te gosse les query parameters in `ed`? We could also go without. I don't have a strong preference, as long as it is uniform. Your call. Je voulais juste te faire penser.
```suggestion
    normalize: bool = Query(True, title="Normalize"),
    preserve_class_order: bool = Query(False, title="Preserve Class Order"),
```

azimuth/modules/model_performance/confusion_matrix.py
```diff
@@ -34,26 +41,58 @@ def compute_on_dataset_split(self) -> List[ConfusionMatrixResponse]:  # type: ig
...
+
+            # Get class indices by reduced-bandwidth order and add rejection_idx at the end.
+            classes_no_rejection = np.delete(class_ids, rejection_idx)
+            order = np.append(classes_no_rejection[order_no_rejection], rejection_idx)
+
+            # Put the rejection class last for the confusion matrix
+            if rejection_idx != order[-1]:
+                rejection_position = np.where(order == rejection_idx)[0][0]
+                # From the rejection position, roll backwards all values
+                order[rejection_position:] = np.roll(order[rejection_position:], -1)
```
gabegma, line 82:
```suggestion
```

tests/test_modules/test_model_performance/test_confusion_matrix.py
gabegma: > Originally we did:
>  1. Preserve class order (but move rejection class last)

That has only been true for [3 days](https://github.com/ServiceNow/azimuth/pull/230). Before, we didn't move the rejection class last.  I think it makes sense to keep the user's logic for all classes; if the order is meaningful, my assumption is the rejection class will but put in a meaningful place as well. It could be first for example. Overwriting doesn't seem right to me.

azimuth/routers/v1/model_performance/confusion_matrix.py
lindsaydbrin: Oh 100% your suggestion is even better, parfait!

gabegma: Cool! I think I have a slight preference for without `-ed`.

tests/test_modules/test_model_performance/test_confusion_matrix.py
lindsaydbrin: Ok, fair point! Lol, how quickly I forget.

lindsaydbrin:
This looks great! A+ stamp of back-end approval :)

azimuth/types/model_performance.py
JosephMarinier: Correct, but in the context of this response, the only "order" provided is the confusion matrix and the list of `class_names`. This number is the index in this `class_names` list. I think it would be correct.
That being said, if we worry about this, we can instead return the `rejection_class` as a string. I can do with that in the front end, it's fine.

azimuth/routers/v1/model_performance/confusion_matrix.py
JosephMarinier: Also/or, if you prefer, you can flip the boolean, like `reverse_cuthill_mckee`, with `True` by default (like `normalized`).
```suggestion
    normalized: bool = Query(True, title="Normalized"),
    reverse_cuthill_mckee: bool = Query(True, title="Reorder classes to reduce the bandwidth of the matrix using reverse Cuthill–McKee algorithm"),
```

gabegma: I would keep the same logic as the switch. I thought it would be best for the switch to display 'Keep user-provided class order' or something like that, and to be `False` by default. But the reverse could also work, to display on the switch 'Reorder classes based on reverse Cuthill–McKee algorithm'.

JosephMarinier: Right, as you wish.
By the way, the other switch says `Normalize`, and not `Normalized`, so removing the `d` seems perfect anyway.

JosephMarinier:
image

azimuth/types/model_performance.py
gabegma: @JosephMarinier - wdyt? Do you want me to make the change?

JosephMarinier: Ho sorry!
Yes, that's what I mean. I wouldn't really have to "search" for its position, as I am already looping over the matrix cells to render them, so it would simply mean changing the check to this:
```diff
-                columnIndex === data.rejectionClassPosition
+                data.classNames[columnIndex] === data.rejectionClass
```

gabegma: I can add the name of the rejection class to dataset info, it might be useful for other modules

JosephMarinier: In addition, sure, if you wish, but I would like to have it here as I don't need `/dataset_info` in this page. It simplifies the front-end code if I don't need to handle loading and failure modes of `/dataset_info`. I think it goes very well with the `class_names` that you return.

gabegma: Ok, done!

gabegma: I replacing the line above in `webapp` but it's not working. Can you take a look?

azimuth/routers/v1/model_performance/confusion_matrix.py
JosephMarinier: If we reverse the switch and query parameter, we could also label the switch something short and sweet like "Reorder classes" and add a Tooltip to provide info about how it is reordered. I think that's a (small) plus. And I like that the two switches are on the same side by default. It feels consistent.
image

azimuth/types/model_performance.py
JosephMarinier: What was not working? It seems fine to me. I committed it.

gabegma: Yeah it was my mistake, I need to kill and restart the back end.

azimuth/routers/v1/model_performance/confusion_matrix.py
gabegma: Ok, I can make the change

gabegma: Done - I will let you push your changes as well.

JosephMarinier (approved):
It's so cool! Good job! 🚢 🇮🇹

PR #250, opened 2022-09-21
Pr/ggm/improve metrics per filter

### LINKED ISSUE #243: Refactor `MetricsPerFilter`
It takes too long and can cause issue with memory.

gabegma:
Resolve #243 

## Description:
On clinc, it takes from 75 sec to 14 sec with the new method which reduces the amount of filtering.  The main changes are modifying metric and bin modules so their methods can directly receive the filtered dataset, instead of going through filters in module options.
[log_less_filtering.txt](https://github.com/ServiceNow/azimuth/files/9612833/log_less_filtering.txt)
[log.txt](https://github.com/ServiceNow/azimuth/files/9612834/log.txt)

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/model_performance/confidence_binning.py
+++ b/azimuth/modules/model_performance/confidence_binning.py
@@ -41,11 +41,15 @@ class ConfidenceHistogramModule(FilterableModule[ModelContractConfig]):
     def get_bins(
         cls, ds: Dataset, without_postprocessing: bool = False
     ) -> List[ConfidenceBinDetails]:
-        """Compute the bins on the dataset split.
+        """Compute the bins on the specified dataset split.
+
+        Note: This lives outside of `compute_on_dataset_split()` so that it can be called without
+        going through calling the module and filtering the dataset.
 
         Args:
             ds: Dataset Split on which to compute bins
-            without_postprocessing: Whether to use outcomes and confidences without pipeline postprocessing
+            without_postprocessing: Whether to use outcomes and confidences without pipeline
+                postprocessing
 
         Returns:
             List of the confidence bins with their confidence and the outcome count.
--- a/azimuth/modules/model_performance/confidence_binning.py
+++ b/azimuth/modules/model_performance/confidence_binning.py
@@ -16,6 +16,10 @@ from azimuth.types.model_performance import (
     ConfidenceHistogramResponse,
 )
 from azimuth.types.outcomes import ALL_OUTCOMES, OutcomeName
+from azimuth.utils.dataset_operations import (
+    get_confidences_from_ds,
+    get_outcomes_from_ds,
+)
 from azimuth.utils.validation import assert_not_none
 
 CONFIDENCE_BINS_COUNT = 20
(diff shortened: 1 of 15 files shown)


tests/profiling_debugging.py
```diff
@@ -43,9 +43,13 @@ def main():
...
         mod = MODULE(dataset_split_name=DATASET_SPLIT, config=config, mod_options=MOD_OPTIONS)
-        mod.compute_on_dataset_split()
+        res = mod.compute_on_dataset_split()
 
-    p.print()
+    file_name = "logs.txt"
+    with open(file_name, "w") as f:
+        p.print(file=f)
+
+    print(res)
```
lindsaydbrin, line 52: This prints to the file and also the screen, right? (Checking my understanding.)

gabegma: This file is only used to profile a module, this is what outputted the logs about the number of seconds that is spent in each function.  Like 49/50 is to print the results of the profiling in a log file, instead of just in the command line.

azimuth/modules/model_performance/metrics.py
```diff
@@ -58,65 +64,75 @@ def first_value(di: Optional[Dict]) -> Optional[float]:
 class MetricsModule(FilterableModule[ModelContractConfig]):
     """Computes different metrics on each dataset split."""
 
-    def compute_on_dataset_split(self) -> List[MetricsModuleResponse]:  # type: ignore
-        ds: Dataset = assert_not_none(self.get_dataset_split())
-        indices = self.get_indices()
-        if len(indices) == 0:
+    def compute_metrics(self, ds: Dataset) -> List[MetricsModuleResponse]:
+        """Compute all metrics on a given dataset split.
```
lindsaydbrin, line 68: It took me some time (and separate conversation!) to understand why we needed to move this code into `compute_metrics` when it's still (aaaaaalmost) all that's being done in `compute_on_dataset_split()`. I think it would be useful to have a comment somewhere in the code documenting this rationale. Maybe the best way to do that would be to revise the docstring more or less like this? (But there could be a better way that I'm not thinking of.)
```suggestion
        """Compute all metrics on a given dataset subset (partial or full dataset split).
             Lives outside of `compute_on_dataset_split()` so that it can be called without module computation.
```
"subset" isn't necessarily right, but I wanted to indicate that it doesn't have to be a "dataset split," i.e. train or eval.

azimuth/modules/model_performance/confidence_binning.py
```diff
@@ -24,23 +28,34 @@
...
+        return [
+            utterance_outcome == outcome
+            for utterance_outcome in get_outcomes_from_ds(ds, without_postprocessing)
+        ]
+
+    @classmethod
+    def get_bins(
+        cls, ds: Dataset, without_postprocessing: bool = False
+    ) -> List[ConfidenceBinDetails]:
+        """Compute the bins on the dataset split.
```
lindsaydbrin, line 44:
```suggestion
        """Compute the bins on the (partial or full) dataset split.
```
Or something else indicating that it's not necessarily the full train or eval splits? (See similar (but broader) comment on metrics.)

azimuth/modules/model_performance/metrics.py
lindsaydbrin: If you think this is helpful, we could consider it for the bin one as well, but for some reason that didn't stand out to me in the same way - maybe for the simple reason that its `compute_on_dataset_split()` has a little more code? I'm not sure.

lindsaydbrin (approved):
Looks great! Thanks for doing this. My comments are mostly communication-related.

azimuth/modules/model_performance/confidence_binning.py
gabegma: I'm not certain about this distinction, because all modules can run on partial or full dataset split, the only difference here is that the dataset is sent as an arg, instead of being loaded with `ds = get.dataset_split()`.  Would `Compute the bins on the specified dataset split.` be better? Or something like that?

azimuth/modules/model_performance/metrics.py
```diff
@@ -58,65 +64,75 @@ def first_value(di: Optional[Dict]) -> Optional[float]:
 class MetricsModule(FilterableModule[ModelContractConfig]):
     """Computes different metrics on each dataset split."""
 
-    def compute_on_dataset_split(self) -> List[MetricsModuleResponse]:  # type: ignore
-        ds: Dataset = assert_not_none(self.get_dataset_split())
-        indices = self.get_indices()
```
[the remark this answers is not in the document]
gabegma: Yeah I think so, but maybe `len(indices)` is faster to compute given it's a list? @Dref360

azimuth/modules/model_performance/confidence_binning.py
lindsaydbrin: Oh, good point. Sure, that's a little more specific!

azimuth/modules/model_performance/metrics.py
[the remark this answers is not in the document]
Dref360: I think this is leftovers of when `get_dataset_split` wouldn't select the indices? They have the same cost

tests/profiling_debugging.py
```diff
@@ -27,7 +29,13 @@
 MOD_OPTIONS = ModuleOptions(pipeline_index=0)
 
 
+def print_memory():
+    threading.Timer(2.0, print_memory).start()
```
Dref360, line 33: [Misc]

Usually we try to not start threads from threads as it gets hard to understand quickly.

You could have:

```python
def print_memory():
	def fn():
		while True:
			print(f"Memory used: {psutil.virtual_memory().used / 1024 / 1024}")
			time.sleep(2)
	threading.Thread(target=print_memory).start()
```

PR #256, opened 2022-09-24
BUG: 404 route not working on root route '/'

### LINKED ISSUE #257: Redirect to page not found not working on root route
**Describe the bug**
When we enter http://localhost:3005/asdfasdfasdfadf . This route is redirecting to /:JobID route rather than page not found.

**To Reproduce**
open localhost and enter random route like  http://localhost:3005/asdfasdfasdfa

Some advices:
* Please don't use customer data or custom paths.
* Link to a gist would be appreciated.

**Expected behavior**
The route should be redirected to page not found

**Vesion (please complete the following information):**
 - OS: Windows (WSL2)
 - Azimuth version: 2.3.0

**Additional context**
Adding a screenshot of the issue

saivishwak:
Resolve #257

## Description:
Redirect to page not found not working on root route

Usually everyone tends to type something on the root route ("/") to navigate to the next page. With the existing approach this feels a like a bad ux experience. And also later when the app needs to have different routes which are not related to data, they can go to their route like /settings/my_profile , /users/:userID. This will be useful when the app grows.

## Checklist:

  our users.
* [ ] **DEV CHANGES.**

### DIFF
--- a/webapp/src/App.tsx
+++ b/webapp/src/App.tsx
@@ -70,49 +70,49 @@ export default class App extends React.Component<Props> {
                 <ErrorBoundary>
                   <Switch>
                     <Route exact path="/">
-                      <Redirect to="/local?pipeline_index=0" />
+                      <Redirect to="/app/local?pipeline_index=0" />
                     </Route>
-                    <Route path="/:jobId">
+                    <Route path="/app/:jobId">
                       <StatusCheck>
                         <PipelineCheck>
                           <Switch>
-                            <Route path="/:jobId" exact>
+                            <Route path="/app/:jobId" exact>
                               <BasicLayout maxWidth="md">
                                 <Dashboard />
                               </BasicLayout>
                             </Route>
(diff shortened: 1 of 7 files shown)


gabegma:
Thank you @saivishwak for adding issues to our repo and creating your first PR! This is really appreciated. @christyler3030, could you take a look?

christyler3030:
Hi @saivishwak thanks for raising this issue. We definitely need to handle these urls better but I think the fix will need to be  deeper than this fix. For example, with this fix we still have the same undesired behaviour at the url: http://localhost:3005/app/asdfasdfasdfadf instead of http://localhost:3005/asdfasdfasdfadf

saivishwak:
> 

Hi @christyler3030 ,

Sure thanks a lot for your reply. Yes it does have same behavior at app/asdasfdasdf. My though process was its easier to get messy from home root ("/") so added another additional route.

PR #260, opened 2022-09-28
Add large dask cluster to config

gabegma:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -222,6 +222,9 @@ class CommonFieldsConfig(ProjectConfig, extra=Extra.ignore):
     # Will use CUDA and will need GPUs if set to True.
     # If "auto" we check if CUDA is available.
     use_cuda: Union[Literal["auto"], bool] = "auto"
+    # Memory of the dask cluster. Regular is 6GB, Large is 12GB.
+    # For bigger models, large might be needed.
+    large_dask_cluster: bool = False
 
     def get_artifact_path(self) -> str:
         """Generate a path for caching.
--- a/docs/docs/reference/configuration/common.md
+++ b/docs/docs/reference/configuration/common.md
@@ -10,6 +10,7 @@ These fields are generic and can be adapted based on the user's machine.
         artifact_path: str = "/cache"
         batch_size: int = 32
         use_cuda: Union[Literal["auto"], bool] = "auto"
+        large_dask_cluster: bool = False
     ```
 
 === "Config Example"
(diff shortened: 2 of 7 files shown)


azimuth/config.py
```diff
@@ -222,6 +222,9 @@ class CommonFieldsConfig(ProjectConfig, extra=Extra.ignore):
     # Will use CUDA and will need GPUs if set to True.
     # If "auto" we check if CUDA is available.
     use_cuda: Union[Literal["auto"], bool] = "auto"
+    # Memory of the dask cluster. Regular is 6GB, Large is 12GB.
+    # For bigger models, large might be needed.
+    large_dask_cluster: bool = False
```
JosephMarinier, line 227: My only interrogation is if we should generalize the config to a string and let the user choose whatever size.
```suggestion
    large_dask_cluster: string = "12GB"
```

JosephMarinier (approved):
That looks good to me! I have a small comment, but it might only be something to keep in mind if we need more flexibility later.

docs/docs/reference/configuration/common.md
```diff
@@ -48,4 +49,10 @@ the memory available on your machine.
 If cuda is available on your machine, set to `true`, otherwise `false`. Can also be set to "auto"
 and let the user-code take care of it.
 
+## Large Dask Cluster
+
+:blue_circle: **Default value**: False
+
+The memory of the dask cluster is usually 6GB. For bigger models, that might not be enough. It can be set to 12GB, by setting `large_dask_cluster` to `True`.
```
lindsaydbrin, line 56:
```suggestion
The memory of the dask cluster is usually 6GB. For bigger models (including encoders), that might not be enough. It can be set to 12GB, by setting `large_dask_cluster` to `True`.
```
Is it helpful to add this to help users think about their needs? Or unnecessary specification?

Also not sure if this is helpful or unnecessary:
```suggestion
The memory of the dask cluster is usually 6GB. For bigger models, especially on a CPU, that might not be enough. It can be set to 12GB, by setting `large_dask_cluster` to `True`.
```

lindsaydbrin: I think we should add something indicating the problem that arises that indicates the need for a large dask cluster. I'm not sure how to word it - maybe something like this, but maybe you or @JosephMarinier can improve it?
```suggestion
The memory of the dask cluster is usually 6GB. Bigger models might need more memory to avoid garbage collection errors. It can be set to 12GB, by setting `large_dask_cluster` to `True`.
```

lindsaydbrin (approved):
LGTM, thanks for doing this!

azimuth/config.py
gabegma: I'm not sure - @Dref360 set that up initally.

Dref360: 6GB fit well for a Mac and 12GB is enough to run most models

gabegma: I would keep it as if for now, until this no longer works.

docs/docs/reference/configuration/common.md
gabegma, line 56:
```suggestion
The memory of the dask cluster is usually 6GB. If your models are big or if you encounter garbage collection errors,  you can set the memory to 12GB by setting `large_dask_cluster` to `True`.
```

PR #262, opened 2022-10-03
Fix bug with empty HF filtering

### LINKED ISSUE #264: Using a combination of 2 filters that lead to an empty table fails.
Issue from HF. We will file a bug. 
We can do a fix locally until it is fixed.

gabegma:
Resolve #264 

## Description:
- Added 2 new tests which used to fail
- Added check that the dataset is not empty at each filtering `if`. This fixes the issue and also makes the function more efficient. 
- I submitted an issue on HF so they can fix the issue: https://github.com/huggingface/datasets/issues/5085

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/utils/dataset_operations.py
+++ b/azimuth/utils/dataset_operations.py
@@ -50,12 +50,20 @@ def filter_dataset_split(
                 <= x[confidence_column][0]
                 <= filters.confidence_max
             )
-    if len(filters.label) > 0 and config.columns.label in dataset_split.column_names:
+    if (
+        len(filters.label) > 0
+        and config.columns.label in dataset_split.column_names
+        and dataset_split.num_rows != 0
+    ):
         dataset_split = dataset_split.filter(lambda x: x[config.columns.label] in filters.label)
-    if filters.utterance is not None and config.columns.text_input in dataset_split.column_names:
+    if (
+        filters.utterance is not None
+        and config.columns.text_input in dataset_split.column_names
+        and dataset_split.num_rows != 0
+    ):
         by = filters.utterance.lower()
         dataset_split = dataset_split.filter(lambda x: by in x[config.columns.text_input].lower())
-    if len(filters.prediction) > 0:
+    if len(filters.prediction) > 0 and dataset_split.num_rows != 0:
         prediction_column = (
             DatasetColumn.model_predictions
             if without_postprocessing
@@ -70,7 +78,7 @@ def filter_dataset_split(
                 dataset_split = dataset_split.filter(
                     lambda x: x[DatasetColumn.postprocessed_prediction] in filters.prediction
                 )
-    if len(filters.data_action) > 0:
+    if len(filters.data_action) > 0 and dataset_split.num_rows != 0:
         # We do OR for data_action tags.
         dataset_split = dataset_split.filter(
             lambda x: any(
@@ -78,7 +86,7 @@ def filter_dataset_split(
                 for v in filters.data_action
             )
         )
-    if len(filters.outcome) > 0:
+    if len(filters.outcome) > 0 and dataset_split.num_rows != 0:
         outcome_column = (
             DatasetColumn.model_outcome
             if without_postprocessing
@@ -91,8 +99,10 @@ def filter_dataset_split(
         # For each smart tag family, we do OR, but AND between families
         # If NO_SMART_TAGS, it is none of them.
         family = cast(SmartTagFamily, key)
-        if len(tags_in_family) > 0 and all(
-            tag in dataset_split.column_names for tag in SMART_TAGS_FAMILY_MAPPING[family]
+        if (
+            len(tags_in_family) > 0
+            and all(tag in dataset_split.column_names for tag in SMART_TAGS_FAMILY_MAPPING[family])
+            and dataset_split.num_rows != 0
         ):
             dataset_split = dataset_split.filter(
                 lambda x: any(
--- a/tests/test_routers/test_model_performance/test_outcome_count.py
+++ b/tests/test_routers/test_model_performance/test_outcome_count.py
@@ -24,3 +24,14 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
     data = resp.json()
     metrics = data.pop("countPerFilter")
     assert "label" in metrics and len(metrics["label"]) == 3
+
+
+def test_empty_search_with_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get("/dataset_splits/eval/outcome_count?utterance=rrgeth").json()
+    assert len(resp["utterances"]) == 0
+
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count?utterance=rrgeth&data_action=relabel"
+    ).json()
+    assert len(resp["utterances"]) == 0
--- a/tests/test_routers/test_model_performance/test_outcome_count.py
+++ b/tests/test_routers/test_model_performance/test_outcome_count.py
@@ -35,7 +35,7 @@ def test_outcome_count_empty_filters(app: FastAPI):
     data = resp.json()
     assert data["utteranceCount"] == 0
 
-# This used to fail when we were filtering on an empty dataset
+    # This used to fail when we were filtering on an empty dataset
     resp = client.get(
         "/dataset_splits/eval/outcome_count/per_filter"
         "?pipeline_index=0&utterance=yukongold&data_action=relabel"
--- a/tests/test_routers/test_model_performance/test_outcome_count.py
+++ b/tests/test_routers/test_model_performance/test_outcome_count.py
@@ -29,15 +29,16 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
 def test_outcome_count_empty_filters(app: FastAPI):
     client = TestClient(app)
     resp = client.get(
-        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=rrgeth"
+        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=yukongold"
     )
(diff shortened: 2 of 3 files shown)


azimuth/utils/dataset_operations.py
```diff
@@ -50,12 +50,20 @@ def filter_dataset_split(
                 <= x[confidence_column][0]
                 <= filters.confidence_max
             )
-    if len(filters.label) > 0 and config.columns.label in dataset_split.column_names:
+    if (
+        len(filters.label) > 0
+        and config.columns.label in dataset_split.column_names
+        and dataset_split.num_rows != 0
```
lindsaydbrin, line 56: I had a thought...does it take much time to do the check on `dataset_split.num_rows` for every if statement? If I'm thinking about this correctly, fewer than or equal if statements would run if we moved the check to the end of each if block and then did a `return` if it were true (so then none of the following if statements would even be checked). I wonder if that would save any time?

tests/test_routers/test_model_performance/test_outcome_count.py
```diff
@@ -24,3 +24,21 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
     data = resp.json()
     metrics = data.pop("countPerFilter")
     assert "label" in metrics and len(metrics["label"]) == 3
+
+
+def test_outcome_count_empty_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=rrgeth"
```
lindsaydbrin, line 32: Small suggestion for entertainment:
```suggestion
        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=yukongold"
```
I really wanted a potato variety that was a QC reference in some way, but the best I could find was [Asterix](https://inspection.canada.ca/plant-health/potatoes/potato-varieties/asterix/eng/1312587385641/1312587385642).

Feel free to ignore this suggestion. 😆

tests/test_routers/test_model_performance/test_outcome_count.py
```diff
@@ -24,3 +24,21 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
...
+    )
+    assert resp.status_code == HTTP_200_OK, resp.text
+    data = resp.json()
+    assert data["utteranceCount"] == 0
+
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count/per_filter"
+        "?pipeline_index=0&utterance=rrgeth&data_action=relabel"
+    )
```
lindsaydbrin, line 41: I'm confused - why would you expect different behavior here when the dataset is already empty without the `data_action` filter? (I.e., why do you need to test this response as well?)

azimuth/utils/dataset_operations.py
lindsaydbrin: (I.e., `if dataset_split.num_rows == 0: return(dataset_split)` or `dataset_split.num_rows == 0 and return(dataset_split)`.)

JosephMarinier: Evaluating `dataset_split.num_rows != 0` should be insignificantly quick. Considering it is done at most 6 times during a request, it is probably not worth optimizing.
That being said, if it was expensive, the question would still have an interesting aspect (surprising for some), so let's suppose that... In normal operation, it wouldn't make that much of a difference. Since Gab placed the `and dataset_split.num_rows != 0` at the end of each `if` statement, it won't be evaluated if the earlier conditions are not met. For example, here, if `len(filters.label)` is not `> 0`, it stops evaluating the conditions and skips to the next filter. The only time your suggestion would make a difference is when we get an empty dataset, and then another filter is applied. That is arguably not part of "normal operation" since we are normally not interested in an empty dataset. We can then optimize for non-empty datasets.
There is even one situation when your suggestion makes it worse: if we filter by confidence (the first filter) and nothing else. In the current implementation, `dataset_split.num_rows != 0` is never checked, but it would be with your suggestion.

lindsaydbrin: Ok, thanks for this! 

Good point with the example of where it's worse...I hadn't considered that of course the length won't be checked if an early conditional is not met. I guess the question would be whether that situation is more or less frequent than getting an empty dataset early on and having to check for other filters anyhow. But that's clearly not necessary analysis if the operation is quick!

tests/test_routers/test_model_performance/test_outcome_count.py
JosephMarinier: That is the situation where the API was crashing before, right? With an already-empty dataset that we try to filter a second time. This test ensures that we fixed the bug and that we get the expected empty list, with no error.

JosephMarinier: Good job testing that, Gab. It might be worth a small comment so that the next reader understands why the test is relevant. It is not obvious without historical knowledge.

JosephMarinier (approved):

tests/test_routers/test_model_performance/test_outcome_count.py
lindsaydbrin: Right, ok - so then why test the first response? To make it really clear where the failure is happening (i.e., not just returning an empty dataset, but filtering further on an empty dataset)?

JosephMarinier: I think it's good. We are doing some unit tests, so one thing at a time; we first assert what the normal behavior is, then we assert that this more difficult case still displays the normal behavior.

tests/test_routers/test_model_performance/test_outcome_count.py
```diff
@@ -24,3 +24,21 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
...
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=rrgeth"
+    )
+    assert resp.status_code == HTTP_200_OK, resp.text
+    data = resp.json()
+    assert data["utteranceCount"] == 0
+
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count/per_filter"
+        "?pipeline_index=0&utterance=rrgeth&data_action=relabel"
```
gabegma, line 40:
```suggestion
        "?pipeline_index=0&utterance=yukongold&data_action=relabel"
```

tests/test_routers/test_model_performance/test_outcome_count.py
```diff
@@ -24,3 +24,21 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
...
+
+def test_outcome_count_empty_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=rrgeth"
+    )
+    assert resp.status_code == HTTP_200_OK, resp.text
+    data = resp.json()
+    assert data["utteranceCount"] == 0
+
```
gabegma, line 37:
```suggestion

# This used to fail when we were filtering on an empty dataset
```

tests/test_routers/test_utterances.py
```diff
@@ -74,6 +74,15 @@ def test_get_utterances(app: FastAPI, monkeypatch):
     assert len(first_utterance["modelSaliency"]) == 2
 
 
+def test_get_utterances_empty_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get("/dataset_splits/eval/utterances?utterance=rrgeth").json()
+    assert len(resp["utterances"]) == 0
+
```
gabegma, line 81:
```suggestion

# This used to fail when we were filtering on an empty dataset
```

tests/test_routers/test_utterances.py
```diff
@@ -74,6 +74,15 @@ def test_get_utterances(app: FastAPI, monkeypatch):
     assert len(first_utterance["modelSaliency"]) == 2
 
 
+def test_get_utterances_empty_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get("/dataset_splits/eval/utterances?utterance=rrgeth").json()
```
gabegma, line 79:
```suggestion
    resp = client.get("/dataset_splits/eval/utterances?utterance=yukongold").json()
```

gabegma (commented):
Changes according to review

tests/test_routers/test_utterances.py
```diff
@@ -74,6 +74,16 @@ def test_get_utterances(app: FastAPI, monkeypatch):
     assert len(first_utterance["modelSaliency"]) == 2
 
 
+def test_get_utterances_empty_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get("/dataset_splits/eval/utterances?utterance=yukongold").json()
+    assert len(resp["utterances"]) == 0
+
+# This used to fail when we were filtering on an empty dataset
```
JosephMarinier, line 82: I think black will tell you to
```suggestion
    # This used to fail when we were filtering on an empty dataset
```

tests/test_routers/test_model_performance/test_outcome_count.py
```diff
@@ -24,3 +24,22 @@ def test_get_outcome_count_per_filter(app: FastAPI) -> None:
...
+def test_outcome_count_empty_filters(app: FastAPI):
+    client = TestClient(app)
+    resp = client.get(
+        "/dataset_splits/eval/outcome_count/per_filter?pipeline_index=0&utterance=yukongold"
+    )
+    assert resp.status_code == HTTP_200_OK, resp.text
+    data = resp.json()
+    assert data["utteranceCount"] == 0
+
+# This used to fail when we were filtering on an empty dataset
```
JosephMarinier, line 38:
```suggestion
    # This used to fail when we were filtering on an empty dataset
```

PR #263, opened 2022-10-05
Add pipeline pre/post-processing steps breakdown in the utterance detail page

### LINKED ISSUE #283: Show the details of the preprocessing and postprocessing steps in the utterance details.

gabegma:
Resolve #283 

## Description:
- For the documentation changes, I recommend looking at [cdd8190](https://github.com/ServiceNow/azimuth/pull/263/commits/cdd8190df53a813b0d37b497bd31f828b6393073) alone so you can see the diff.
- I added a performance test so that it doesn't take too long to load the HF tables as we add columns to them, but I'm afraid that the tests will be a bit flaky. Not sure what is the best way to proceed and how much buffer to leave.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


azimuth/utils/ml/postprocessing.py
```diff
@@ -26,6 +26,28 @@ def __getitem__(self, item: int) -> "PostProcessingIO":
         )
 
 
+class PreprocessingStep(AliasModel):
+    order: int
+    text: List[str]
```
JosephMarinier, line 31: Why a list? Related to that, I am confused by the `__getitem__()` returning another `PreprocessingStep`. Can you tell me more?

gabegma: Yes, it's because pipelines usually return the results for a batch. So this class could contain the preprocessing step from all utterances. The `__getitem__()` is implemented so that we can easily retrieve the result for one utterance, by doing `PreprocessingStep()[idx]`.

docs/docs/reference/custom-objects/processors.md
```diff
@@ -0,0 +1,265 @@
...
+        postprocessed_predictions: MyPipelineOutput
+    ```
+
+### In the Config
+
+`{"postprocessors": null}` should then be added to the config, to avoid re-postprocessing in
+Azimuth.
+
+### `PipelineOutputProtocolV2`
+If using V2, to new fields need to be provided: preprocessing steps and postprocessing steps.
```
JosephMarinier, line 120:
```suggestion
If using V2, two new fields need to be provided: preprocessing steps and postprocessing steps.
```

azimuth/modules/model_contracts/text_classification.py
```diff
@@ -35,6 +36,18 @@ class PipelineOutputProtocol(Protocol):
...
+@runtime_checkable
+class PipelineOutputProtocolV2(Protocol):
+    """Class containing result of a batch with pre and postprocessing steps"""
+
+    # model output without passing through post-processing stage: texts, logits, probs, preds
+    model_output: PostProcessingIO
+    # output after passing through post-processing: pre-processed texts, logits, probs, preds
+    postprocessor_output: PostProcessingIO
+    preprocessing_steps: List[Dict[str, Union[str, List[str], int]]]
+    postprocessing_steps: List[Dict[str, Union[str, PostProcessingIO, int]]]
```
JosephMarinier, line 48: Why not just add the fields to the existing `PipelineOutputProtocol` as `Optional`?
```suggestion
    preprocessing_steps: Optional[List[Dict[str, Union[str, List[str], int]]]]
    postprocessing_steps: Optional[List[Dict[str, Union[str, PostProcessingIO, int]]]]
```

azimuth/modules/model_contracts/text_classification.py
```diff
@@ -226,6 +296,8 @@ def get_postprocessed_output(
                     preds=pipeline_output.postprocessor_output.preds,
                     logits=pipeline_output.postprocessor_output.logits,
                 ),
+                [],
+                [],
```
JosephMarinier, line 300: We could do something like this here, and remove the `if isinstance(pipeline_output, PipelineOutputProtocolV2):` with the duplicated code in it.
```suggestion
                [] if pipeline_output.preprocessing_steps is None else [
                    PreprocessingStep(
                        order=step["order"], class_name=step["class_name"], text=step["text"]
                    )
                    for step in pipeline_output.preprocessing_steps
                ],
                [] if pipeline_output.postprocessing_steps is None else [
                    PostprocessingStep(
                        order=step["order"],
                        class_name=step["class_name"],
                        output=PostProcessingIO(
                            texts=input_batch[self.config.columns.text_input],
                            probs=cast(PostProcessingIO, step["output"]).probs,
                            preds=cast(PostProcessingIO, step["output"]).preds,
                            logits=cast(PostProcessingIO, step["output"]).logits,
                        ),
                    )
                    for step in pipeline_output.postprocessing_steps
                ],
```

JosephMarinier (commented):
Small architecture question...

tests/test_modules/test_model_contracts/test_text_classification.py
```diff
@@ -90,6 +90,13 @@ def test_post_process(simple_text_config):
     # With a low threshold, the post-processing is not affecting the predicted class with a random
     # model.
     assert np.all([pred.model_output.preds == pred.postprocessed_output.preds for pred in res_low])
+    # The prediction of the last pipeline step should be the same as the postprocessed prediction.
+    assert np.all(
+        [
+            pred.postprocessed_output.preds == pred.postprocessing_steps[-1].output.preds
+            for pred in res_low
+        ]
+    )
```
JosephMarinier, line 99: Small trick just to make your life easier in the future: `all()` (not `np.all()`) accepts a generator directly, so no need to create a list comprehension.
```suggestion
    assert all(
        pred.postprocessed_output.preds == pred.postprocessing_steps[-1].output.preds
        for pred in res_low
    )
```

docs/docs/reference/custom-objects/processors.md
```diff
@@ -0,0 +1,265 @@
...
+The preprocessing steps need to be returned as a a `List` of `Dict` with the following fields and types.
+
+=== "Dict Fields and Types"
+
+    ```python
+    from typing import List
+
+    class_name: str  # (1)
+    text: List[str]  # (2)
+    order: int # (3)
```
JosephMarinier, line 131: Aren't we missing the class definition?
```suggestion

    class PreprocessingStep(AliasModel):
        """Class received from the pipeline, and used in the Prediction Module."""

        class_name: str  # (1)
        text: List[str]  # (2)
        order: int # (3)
```

azimuth/utils/ml/postprocessing.py
```diff
@@ -26,6 +28,40 @@ def __getitem__(self, item: int) -> "PostProcessingIO":
...
+class PredictionDetails(AliasModel):
+    # Prediction class names, sorted by confidences
+    predictions: List[str]
+    # Predicted class, which can be different than the first element of predictions, when
+    # thresholding for instance.
+    prediction: str
+    # Sorted confidences
+    confidences: List[float]
+    # Outcome based on label and prediction
+    outcome: OutcomeName
```
JosephMarinier, line 40: Would you like to move these comments to a real docstring? Everything is there!

azimuth/utils/ml/model_performance.py
```diff
@@ -24,3 +25,27 @@ def sorted_by_utterance_count_with_last(metrics: List[T], index_to_put_first: in
...
+    """
+    if prediction == label:
+        if label == rejection_class_idx:
+            return OutcomeName.CorrectAndRejected
+        else:
+            return OutcomeName.CorrectAndPredicted
+    elif prediction == rejection_class_idx:
+        return OutcomeName.IncorrectAndRejected
+    else:
+        return OutcomeName.IncorrectAndPredicted
```
JosephMarinier, line 51: I know this is not new code, but I find it hard to read, considering how simple it is conceptually. I thought I'd give a small suggestion to bring a little bit more structure.
```suggestion
    rejected = prediction == rejection_class_idx
    if prediction == label:
        return OutcomeName.CorrectAndRejected if rejected else OutcomeName.CorrectAndPredicted
    else:
        return OutcomeName.IncorrectAndRejected if rejected else OutcomeName.IncorrectAndPredicted
```

JosephMarinier (commented):
As discussed offline, let's remove the temperature scaler of 1 that is there by default.

azimuth/modules/base_classes/indexable_module.py
```diff
@@ -158,20 +165,34 @@ def run_postprocessing(self, output: PostProcessingIO, **kwargs) -> PostProcessi
...
                 #    It is less burdensome to just reload it.
                 fn = load_custom_object(post, **kwargs)
                 output = fn(output)
-        return output
+                # When updating to python 3.9, use .removeprefix()
+                class_name = (
+                    post.class_name[32:]
+                    if post.class_name.startswith("azimuth.utils.ml.postprocessing.")
+                    else post.class_name
+                )
```
JosephMarinier, line 181: Two small suggestion to make this bit more robust:
1. We can use `removeprefix()`. That avoids hardcoding the length `32`.
2. Instead of hardcoding the string `"azimuth.utils.ml.postprocessing"`, which might change, we can use `PostprocessingStep.__module__`, or add `from azimuth.utils.ml import postprocessing` and use `postprocessing.__name__`.
```suggestion
                class_name = post.class_name.removeprefix(f"{PostprocessingStep.__module__}.")
```

azimuth/modules/base_classes/indexable_module.py
```diff
@@ -202,6 +223,40 @@ def _save_result(self, res, dm):
...
+                                class_name=step.class_name,
+                                output=PredictionDetails(
+                                    predictions=[
+                                        class_names[cl_idx]
+                                        for cl_idx in reversed(np.argsort(step.output.probs[0]))
+                                    ],
+                                    prediction=class_names[step.output.preds[0]],
+                                    confidences=[
+                                        prob for prob in reversed(np.sort(step.output.probs[0]))
+                                    ],
```
JosephMarinier, line 247: I think it's more efficient to reverse numpy arrays using `[::-1]` than using `reverse()`. That would apply for the other `reverse()` above too. However, the gain would probably be insignificant, so feel free to keep `reverse()` if you prefer it for readability.
Also, instead of `[prob for prob in ...]` we can use `list(...)`.
```suggestion
                                    confidences=list(np.sort(step.output.probs[0])[::-1]),
```

azimuth/utils/ml/postprocessing.py
```diff
@@ -18,6 +19,7 @@ def is_multiclass(self):
         return np.allclose(self.probs.sum(-1), 1.0)
 
     def __getitem__(self, item: int) -> "PostProcessingIO":
+        """Useful to get from a batch result to a single utterance result."""
```
JosephMarinier, line 22: I think we can skip the "Useful to", and maybe invert the sentence? That's candy, though.
```suggestion
        """Get a single utterance result from a batch result."""
```

docs/docs/reference/custom-objects/processors.md
```diff
@@ -0,0 +1,251 @@
...
+            "postprocessors": [
+                  {
+                    "class_name": "loading_resources.TemperatureScaling",
+                    "remote": "/azimuth_shr",
+                    "kwargs": {"temperature": 3}
+                  }
+                 ]
+         }
+      ]
+    ```
```
JosephMarinier, line 249: Houf, you have been creative with indentation! 😝 
```suggestion
    ```json

    "pipelines": [
        {
            "model": ...,
            "postprocessors": [
                {
                    "class_name": "loading_resources.TemperatureScaling",
                    "remote": "/azimuth_shr",
                    "kwargs": {"temperature": 3}
                }
            ]
        }
    ]
    ```
```

azimuth/modules/base_classes/indexable_module.py
JosephMarinier: By the way, does that mean I should expect the `class_name` of custom steps to contain some module names (`something.something.something.`) or not? I'm asking with in mind how I treat the names in the front end.

gabegma: It all depends on what people put as a value there in their pipeline, but I don't think that should be the case by default.

azimuth/modules/base_classes/indexable_module.py
gabegma: > `[prob for prob in ...]` we can use `list(...)`

Oh, awesome, I never realized I could simplify this. Thanks!!

docs/docs/reference/custom-objects/processors.md
gabegma: Thanks for fixing it! I probably just copy paste and forgot.

azimuth/config.py
```diff
@@ -87,15 +87,9 @@ def check_threshold(cls, values):
...
-    # Postprocessors
     postprocessors: Optional[
         List[Union[TemperatureScaling, ThresholdConfig, CustomObject]]
-    ] = Field(
-        [
-            TemperatureScaling(temperature=1.0),
-            ThresholdConfig(threshold=0.5),
-        ]
-    )
+    ] = Field([ThresholdConfig(threshold=0.5)], nullable=True)
```
JosephMarinier, line 92: This is the problem that @nandhinibsn has been experiencing; it was missing `nullable=True`. It's not related to this PR, but I committed it here to avoid conflicts: 7a053daa36f66f9cca442b04a1816a7821903153

azimuth/modules/base_classes/indexable_module.py
```diff
@@ -158,34 +165,46 @@ def run_postprocessing(self, output: PostProcessingIO, **kwargs) -> PostProcessi
...
+                        output=output,
+                    )
+                )
+        return postprocessing_steps
 
     def _save_result(self, res, dm):
-        table_key = self._get_table_key()
-
-        # Save result in a DatasetSplitManager
-        if len(res) > 0 and isinstance(res[0], PredictionResponse):
```
JosephMarinier, line 174: Just to make sure this is intentional; why is the `if len(res) > 0` not necessary anymore?

gabegma: Because since then we refactored `save_result()` and it already checks the `len` of `res` in https://github.com/ServiceNow/azimuth/blob/d690cc4d37a6c8e4d5e2ff8658c1ae3fefbb7a8d/azimuth/modules/base_classes/indexable_module.py#L72

JosephMarinier: Perfect! Thanks!

JosephMarinier:
Hello @christyler3030, @nandhinibsn, could you please review my work on the front end? I recommend reviewing those 3 commits separately:
* Preliminary work:
  * [Replace `makeStyles`](https://github.com/ServiceNow/azimuth/pull/263/commits/437eec70fa90bb665d3df43aa5691dd7faa4bbfb)
  * [Add and test `camelToTitleCase`](https://github.com/ServiceNow/azimuth/pull/263/commits/cd7311f933d46266feae349fd49729406de4dc23)
* Preliminary work:
  * [Add processing step selector](https://github.com/ServiceNow/azimuth/pull/263/commits/8cbd68b04984acf526eb874e6c85f52a98be930b)
 
image

webapp/src/pages/UtteranceDetail.tsx
```diff
@@ -109,6 +126,36 @@ export const UtteranceDetail = () => {
 
   const { modelPrediction } = utterance;
 
+  const preprocessingSteps = [
+    { className: "Initial", text: utterance.utterance },
```
christyler3030, line 130: Nitpick: this will work fine but 'className' already means something in jsx

JosephMarinier: Good point. I'll make sure it's not used as a JSX attribute to avoid confusion.

It is difficult to find alternatives to the word "class", which in our code might refer to
- a Python class;
- the HTML class attribute; or
- the classification labels/predictions (like in `datasetInfo.classNames`).

Something to keep in mind.

christyler3030 (approved):

PR #275, opened 2022-10-19
Add smart tag analysis plot

### LINKED ISSUE #167: Plot to highlight patterns between smart tags and labels (or else?)
Might be a 2D matrix (heatmap) with smart tags and labels.

JosephMarinier:
Resolve #167

## Description:

Add a smart tag analysys page, as well as a card on the dashboard.

A screenshot is worth a thousand words:
image
image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/SmartTagsTable.tsx
+++ b/webapp/src/components/SmartTagsTable.tsx
@@ -264,7 +264,7 @@ const SmartTagsTable: React.FC<{
           }}
         >
           <TableSortLabel {...getSortLabelProps("accuracy")}>
-            Class Accuracy
+            Accuracy
           </TableSortLabel>
         </Typography>
         <Typography
@@ -416,7 +416,9 @@ const SmartTagsTable: React.FC<{
                             width="100%"
                             to={`/${jobId}/dataset_splits/${datasetSplitName}/performance_overview${constructSearchString(
                               {
-                                label: [row.filterValue],
+                                [selectedMetricPerFilterOption]: [
+                                  row.filterValue,
+                                ],
                                 outcome: [outcome],
                                 [family]: smartTags(family),
                                 ...pipeline,
--- a/webapp/src/components/SmartTagsTable.tsx
+++ b/webapp/src/components/SmartTagsTable.tsx
@@ -189,7 +189,7 @@ const SmartTagsTable: React.FC<{
         columnGap={4}
         sx={{ overscrollBehaviorX: "contain" }} // Stops accidental navigation on horizontal scroll with touch pad
       >
-        {sortedRows?.flatMap((classCount, classIndex) => (
+        {sortedRows?.map((classCount, classIndex) => (
           <React.Fragment key={classCount.filterValue}>
             <Typography
               variant="subtitle2"
--- a/webapp/src/components/SmartTagsTable.tsx
+++ b/webapp/src/components/SmartTagsTable.tsx
@@ -289,7 +289,6 @@ const SmartTagsTable: React.FC<{
               display="flex"
               flexDirection="column"
               gap={1}
-              width={160}
               position="sticky"
               bgcolor={(theme) => theme.palette.background.paper}
               {...{
@@ -344,6 +343,7 @@ const SmartTagsTable: React.FC<{
                   borderLeft="solid 1px"
                   borderRight="solid 1px"
                   paddingY={0.5}
+                  width={160}
                   {...{
                     [`grid${transpose ? "Column" : "Row"}`]: classIndex + 2,
                     [`grid${transpose ? "Row" : "Column"}`]: familyIndex + 4,
--- a/webapp/src/components/SmartTagsTable.tsx
+++ b/webapp/src/components/SmartTagsTable.tsx
@@ -1,0 +1,476 @@
+import { ArrowForward } from "@mui/icons-material";
+import {
+  Box,
+  FormControlLabel,
+  Link,
+  MenuItem,
+  Select,
+  Switch,
+  TableSortLabel,
+  TableSortLabelProps,
+  Tooltip,
+  Typography,
+} from "@mui/material";
+import noData from "assets/void.svg";
+import Loading from "components/Loading";
+import { motion } from "framer-motion";
+import React from "react";
+import { Link as RouterLink } from "react-router-dom";
+import { getOutcomeCountPerFilterEndpoint } from "services/api";
+import {
+  AvailableDatasetSplits,
+  DatasetSplitName,
+  OutcomeCountPerFilterValue,
+} from "types/api";
+import { QueryPipelineState } from "types/models";
+import {
+  ALL_OUTCOMES,
+  OUTCOME_COLOR,
+  OUTCOME_PRETTY_NAMES,
+  SmartTagFamily,
+  SMART_TAG_FAMILIES,
+  SMART_TAG_FAMILY_ICONS,
+  SMART_TAG_FAMILY_PRETTY_NAMES,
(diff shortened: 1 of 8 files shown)


christyler3030 (approved):

lindsaydbrin:
This looks awesome! 

One bug I found (I'm going to try to review in the code, but I'm not sure where this one is): When you click on a bar in the plot, if you are looking at Predictions instead of Labels, the link still goes to the exploration space with the Label filtered by the class in question.

webapp/src/pages/Dashboard.tsx
```diff
@@ -97,6 +99,25 @@ const Dashboard = () => {
           />
         </PreviewCard>
       )}
+      {isPipelineSelected(pipeline) && (
+        <PreviewCard
+          title="Smart Tags Aggregation"
```
lindsaydbrin, line 104: Minimally:
```suggestion
          title="Smart Tag Aggregation"
```

But I wonder if there's a better word than "Aggregation" (although I know that's what I've been calling it myself 🙃), both to be a little broader than the technical action leading to the plot, and maybe also to be consistent with the rest of the Dashboard. So far, we use "Testing" and "Analysis" on the Dashboard...I was going to suggest "Summary," but "Analysis" would also work.
```suggestion
          title="Smart Tag Analysis"
```

```suggestion
          title="Smart Tag Summary"
```

What do you think?

lindsaydbrin:
I noticed that when you switch between the training and eval sets, the order of the classes changes. I think this is because it's going to a default that has to do with the sorting logic? I think it would be more expected behaviour to maintain the order of classes when you change dataset splits, e.g. to be able to compare the same class in both splits, then to maintain the order of the last column you sorted on. What do you think?

webapp/src/components/SmartTagsTable.tsx
```diff
@@ -0,0 +1,476 @@
...
+}> = ({ jobId, pipeline, availableDatasetSplits }) => {
+  const [datasetSplitName, setDatasetSplitName] =
+    React.useState<DatasetSplitName>("eval");
+  const [transpose, setTranspose] = React.useState(false);
+
+  const [selectedMetricPerFilterOption, setSelectedMetricPerFilterOption] =
+    React.useState<"label" | "prediction">("label");
+
+  const [ascending, setAscending] = React.useState(false);
+  const [sortBy, setSortBy] = React.useState<SortBy>("utteranceCount");
```
JosephMarinier, line 63: @lindsaydbrin

> I noticed that when you switch between the training and eval sets, the order of the classes changes.

The table is initially sorted by the total utterance count column. I chose that to be consistent with the performance analysis table, which is just above this one in the dashboard, but we can change that here. Of course, the utterance counts change between the dataset splits.

If you sort by label/prediction classes, then the order persists between the dataset splits. Maybe that should be the default order when you load the page? So that if your first interaction is changing the dataset split, the order persists. What do you think?
```suggestion
  const [sortBy, setSortBy] = React.useState<SortBy>("filterValue");
```

lindsaydbrin: Oh, interesting. I wasn't sure about the default because I'd already been sorting things, and then I was startled to see the order change. Like, if you sort by accuracy and see a class you're interested in, you might want to have it stay put when you switch to the other dataset split.

But that said, I think it's reasonable to have the same behavior as the performance analysis table (default/initial view as well as switching sets). So maybe my suggestion/use case is a design question that we can think about later.

webapp/src/components/SmartTagsTable.tsx
```diff
@@ -0,0 +1,476 @@
...
+                          initial={false}
+                          transition={{ type: "tween" }}
+                          display="flex"
+                        >
+                          <Link
+                            component={RouterLink}
+                            width="100%"
+                            to={`/${jobId}/dataset_splits/${datasetSplitName}/performance_overview${constructSearchString(
+                              {
+                                label: [row.filterValue],
```
JosephMarinier, line 419: @lindsaydbrin
> if you are looking at Predictions instead of Labels, the link still goes to the exploration space with the Label filtered by the class in question
Good catch! Thank you very much! That will be the fix:
```suggestion
                                [selectedMetricPerFilterOption]: [
                                  row.filterValue,
                                ],
```

webapp/src/components/SmartTagsTable.tsx
```diff
@@ -0,0 +1,476 @@
...
+          />
+        </Box>
+        <FormControlLabel
+          control={
+            <Switch
+              value={transpose}
+              onChange={(_, checked) => setTranspose(checked)}
+            />
+          }
+          label="Transpose"
```
lindsaydbrin, line 181: I think the word "Transpose" is clear for technical users. If there's a better, more user-friendly option, we can get feedback/discuss with design later! My only question is whether something like "Transpose axes" is more clear, but I'm not sure it's necessary.

webapp/src/components/SmartTagsTable.tsx
```diff
@@ -0,0 +1,476 @@
...
+          position="sticky"
+          bgcolor={(theme) => theme.palette.background.paper}
+          {...{
+            [`grid${transpose ? "Column" : "Row"}`]: 1,
+            [`grid${transpose ? "Row" : "Column"}`]: 2,
+            [transpose ? "left" : "top"]: 0,
+          }}
+        >
+          <TableSortLabel {...getSortLabelProps("accuracy")}>
+            Class Accuracy
```
lindsaydbrin, line 267: I think we can and should drop the "Class" part, for a couple reasons: (1) it's redundant with the row labels, (2) it could become incorrect if we expand the row labels (i.e., individual smart tags, for smart tag inception), (3) it's actually even a little misleading when you have "prediction" as a row label, because I might think it still meant labelled class accuracy.

webapp/src/components/SmartTagsTable.tsx
lindsaydbrin: Awesome, so then in theory it's robust if we come up with more options for row labels! :)

webapp/src/components/SmartTagsTable.tsx
JosephMarinier: Agree.
I'm not sure about "Transpose axes". I prefer "Transpose" as we are doing precisely that. We are not doing a special thing with only the axes. We are transposing the whole matrix.
If "Transpose" is too technical, we can go with something like `Switch axes`. Or, going away from maths even further, Google Sheets has a similar option as a checkbox called "Switch rows / columns". Not bad.

webapp/src/components/SmartTagsTable.tsx
JosephMarinier: Cool! Thank you!

webapp/src/pages/Dashboard.tsx
JosephMarinier: I have no attachment for "aggregation". I think this plot is too verbose to be a "summary". I'll go with "analysis". Thank you! :)

webapp/src/components/SmartTagsTable.tsx
JosephMarinier: Ho yeas! 🤟

webapp/src/components/SmartTagsTable.tsx
lindsaydbrin: My mathy brain prefers "transpose." I was going to say, if Google thinks that is a user friendly approach, who am I to argue...but then I realized that they also have a broader audience, and I prefer transpose for our more technical one. :) But I don't feel strongly, so, comme tu veux!

PR #282, opened 2022-10-25
Added new fields in config page

nandhinibsn:
Resolve #

## Description: 
This is the feature to enable next important fields in the configuration settings page for the user to modify and see the changes made in the UI.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/AccordianLayout.tsx
+++ b/webapp/src/components/AccordianLayout.tsx
@@ -1,0 +1,51 @@
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import {
+  Accordion,
+  AccordionDetails,
+  AccordionSummary,
+  Box,
+  Divider,
+  Typography,
+} from "@mui/material";
+import React from "react";
+import Description from "components/Description";
+
+type Props = {
+  name: string;
+  description: string;
+  docURL: string;
+  children: React.ReactNode;
+};
+
+const AccordianLayout: React.FC<Props> = ({
+  name,
+  description,
+  docURL,
+  children,
+}) => {
+  return (
+    <Accordion>
+      <AccordionSummary expandIcon={<ExpandMoreIcon />} id={name}>
+        <Typography
+          variant="body1"
+          sx={{
+            width: "15%",
+            flexShrink: 0,
+          }}
+        >
+          {name}
+        </Typography>
+        <Box display="flex" gap={2}>
+          <Typography variant="body1">{description}</Typography>
+          <Description link={docURL} />
+        </Box>
+      </AccordionSummary>
+      <AccordionDetails>
+        <Divider sx={{ marginBottom: 1 }} />
+        {children}
+      </AccordionDetails>
+    </Accordion>
+  );
+};
+
+export default React.memo(AccordianLayout);
--- a/webapp/src/components/AccordionLayout.test.tsx
+++ b/webapp/src/components/AccordionLayout.test.tsx
@@ -3,25 +3,16 @@ import { fireEvent, screen } from "@testing-library/react";
 import AccordionLayout from "components/AccordionLayout";
 import { renderWithTheme } from "mocks/utils";
 
-type Props = {
-  name: string;
-  description: string;
-  link: string;
-  children: React.ReactNode;
-};
-
-const renderAccordionLayout = (accordionLayout: Props) =>
-  renderWithTheme(<AccordionLayout {...accordionLayout} />);
-
 describe("AccordionLayout", () => {
-  it("should display name, description and description component on the accordian layout", () => {
-    renderAccordionLayout({
+  it("should display name, description and description component on the accordion layout", () => {
+    const props = {
       name: "Project Config",
       description:
         "Contains mandatory fields that specify the dataset to load in Azimuth",
       link: "reference/configuration/project/",
       children: <Typography>Project Config Section</Typography>,
-    });
+    };
+    renderWithTheme(<AccordionLayout {...props} />);
     expect(screen.getByText("Project Config")).toBeVisible();
     expect(
       screen.getByText(
--- a/webapp/src/components/AccordionLayout.test.tsx
+++ b/webapp/src/components/AccordionLayout.test.tsx
@@ -1,0 +1,40 @@
+import { Typography } from "@mui/material";
+import { fireEvent, screen } from "@testing-library/react";
+import AccordionLayout from "components/AccordionLayout";
+import { renderWithTheme } from "mocks/utils";
+
+type Props = {
+  name: string;
+  description: string;
+  link: string;
+  children: React.ReactNode;
+};
+
+const renderAccordionLayout = (accordionLayout: Props) =>
+  renderWithTheme(<AccordionLayout {...accordionLayout} />);
+
+describe("AccordionLayout", () => {
+  it("should display name, description and description component on the accordian layout", () => {
+    renderAccordionLayout({
+      name: "Project Config",
+      description:
+        "Contains mandatory fields that specify the dataset to load in Azimuth",
+      link: "reference/configuration/project/",
+      children: <Typography>Project Config Section</Typography>,
+    });
+    expect(screen.getByText("Project Config")).toBeVisible();
+    expect(
+      screen.getByText(
+        "Contains mandatory fields that specify the dataset to load in Azimuth"
+      )
+    ).toBeVisible();
+    const link: HTMLAnchorElement = screen.getByRole("link");
+    expect(link.href).toContain(
+      "https://servicenow.github.io/azimuth/main/reference/configuration/project/"
+    );
+    screen.getByText("Learn more");
+    screen.getByTestId("OpenInNewIcon");
+    fireEvent.click(screen.getAllByRole("button")[0]);
+    expect(screen.getByText("Project Config Section")).toBeVisible();
+  });
+});
--- a/webapp/src/components/AccordionLayout.test.tsx
+++ b/webapp/src/components/AccordionLayout.test.tsx
@@ -8,7 +8,7 @@ describe("AccordionLayout", () => {
     const props = {
       name: "Project Config",
       description:
-        "Contains mandatory fields that specify the dataset to load in Azimuth",
+        "contains mandatory fields that specify the dataset to load in Azimuth",
       link: "reference/configuration/project/",
       children: <Typography>Project Config Section</Typography>,
     };
@@ -16,7 +16,7 @@ describe("AccordionLayout", () => {
     expect(screen.getByText("Project Config")).toBeVisible();
     expect(
       screen.getByText(
-        "Contains mandatory fields that specify the dataset to load in Azimuth"
+        "contains mandatory fields that specify the dataset to load in Azimuth"
       )
     ).toBeVisible();
     const link: HTMLAnchorElement = screen.getByRole("link");
--- a/webapp/src/components/AccordionLayout.tsx
+++ b/webapp/src/components/AccordionLayout.tsx
@@ -28,6 +28,7 @@ const AccordionLayout: React.FC<Props> = ({
       <AccordionSummary expandIcon={<ExpandMoreIcon />} id={name}>
         <Typography
           variant="body1"
+          fontWeight="bold"
           sx={{
             width: "15%",
             flexShrink: 0,
--- a/webapp/src/components/AccordionLayout.tsx
+++ b/webapp/src/components/AccordionLayout.tsx
@@ -25,24 +25,19 @@ const AccordionLayout: React.FC<Props> = ({
 }) => (
   <Accordion>
     <AccordionSummary expandIcon={<ExpandMoreIcon />} id={name}>
-      <Typography
-        variant="body1"
-        fontWeight="bold"
-        sx={{
-          width: "20%",
-          flexShrink: 0,
-        }}
-      >
+      <Typography fontWeight="bold" width="20%" flexShrink={0}>
         {name}
       </Typography>
-      <Divider orientation="vertical" flexItem sx={{ marginX: 2 }} />
       <Box display="flex" gap={2}>
-        <Typography variant="body1">{description}</Typography>
+        <Divider orientation="vertical" />
+        <Typography>{description}</Typography>
         <Description link={link} />
       </Box>
     </AccordionSummary>
-    <AccordionDetails>
-      <Divider sx={{ marginBottom: 1 }} />
+    <AccordionDetails
+      sx={{ display: "flex", flexDirection: "column", paddingY: 0 }}
+    >
+      <Divider />
       {children}
     </AccordionDetails>
   </Accordion>
--- a/webapp/src/components/AccordionLayout.tsx
+++ b/webapp/src/components/AccordionLayout.tsx
@@ -30,7 +30,7 @@ const AccordionLayout: React.FC<Props> = ({
           variant="body1"
           fontWeight="bold"
           sx={{
-            width: "15%",
+            width: "20%",
             flexShrink: 0,
           }}
         >
--- a/webapp/src/components/AccordionLayout.tsx
+++ b/webapp/src/components/AccordionLayout.tsx
@@ -13,14 +13,14 @@ import Description from "components/Description";
 type Props = {
   name: string;
   description: string;
-  docURL: string;
+  link: string;
   children: React.ReactNode;
 };
 
-const AccordianLayout: React.FC<Props> = ({
+const AccordionLayout: React.FC<Props> = ({
   name,
   description,
-  docURL,
+  link,
   children,
 }) => {
   return (
@@ -37,7 +37,7 @@ const AccordianLayout: React.FC<Props> = ({
         </Typography>
         <Box display="flex" gap={2}>
           <Typography variant="body1">{description}</Typography>
-          <Description link={docURL} />
+          <Description link={link} />
         </Box>
       </AccordionSummary>
       <AccordionDetails>
@@ -48,4 +48,4 @@ const AccordianLayout: React.FC<Props> = ({
   );
 };
 
-export default React.memo(AccordianLayout);
+export default React.memo(AccordionLayout);
--- a/webapp/src/components/AccordionLayout.tsx
+++ b/webapp/src/components/AccordionLayout.tsx
@@ -36,6 +36,7 @@ const AccordionLayout: React.FC<Props> = ({
         >
           {name}
         </Typography>
+        <Divider orientation="vertical" flexItem sx={{ marginX: 2 }} />
         <Box display="flex" gap={2}>
           <Typography variant="body1">{description}</Typography>
           <Description link={link} />
--- a/webapp/src/components/AccordionLayout.tsx
+++ b/webapp/src/components/AccordionLayout.tsx
@@ -7,8 +7,8 @@ import {
   Divider,
   Typography,
 } from "@mui/material";
-import React from "react";
 import Description from "components/Description";
+import React from "react";
 
 type Props = {
   name: string;
@@ -22,32 +22,30 @@ const AccordionLayout: React.FC<Props> = ({
   description,
   link,
   children,
-}) => {
-  return (
-    <Accordion>
-      <AccordionSummary expandIcon={<ExpandMoreIcon />} id={name}>
-        <Typography
-          variant="body1"
-          fontWeight="bold"
-          sx={{
-            width: "20%",
-            flexShrink: 0,
-          }}
-        >
-          {name}
-        </Typography>
-        <Divider orientation="vertical" flexItem sx={{ marginX: 2 }} />
-        <Box display="flex" gap={2}>
-          <Typography variant="body1">{description}</Typography>
-          <Description link={link} />
-        </Box>
-      </AccordionSummary>
-      <AccordionDetails>
-        <Divider sx={{ marginBottom: 1 }} />
-        {children}
-      </AccordionDetails>
-    </Accordion>
-  );
-};
+}) => (
+  <Accordion>
+    <AccordionSummary expandIcon={<ExpandMoreIcon />} id={name}>
+      <Typography
+        variant="body1"
+        fontWeight="bold"
+        sx={{
+          width: "20%",
+          flexShrink: 0,
+        }}
+      >
+        {name}
+      </Typography>
+      <Divider orientation="vertical" flexItem sx={{ marginX: 2 }} />
+      <Box display="flex" gap={2}>
+        <Typography variant="body1">{description}</Typography>
+        <Description link={link} />
+      </Box>
+    </AccordionSummary>
+    <AccordionDetails>
+      <Divider sx={{ marginBottom: 1 }} />
+      {children}
+    </AccordionDetails>
+  </Accordion>
+);
 
 export default React.memo(AccordionLayout);
--- a/webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
+++ b/webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
@@ -10,13 +10,13 @@ const DESCRIPTION = {
   "General Warnings": (
     <Description
       text="Investigate issues with class size, class imbalance, and dataset shift."
-      link="/dataset-warnings/#general-warnings"
+      link="user-guide/dataset-warnings/"
     />
   ),
   "Syntactic Warnings": (
     <Description
       text="Compare utterance lengths for your training and evaluation data."
-      link="/dataset-warnings/#syntactic-warnings"
+      link="user-guide/dataset-warnings/"
     />
   ),
 };
--- a/webapp/src/components/Description.test.tsx
+++ b/webapp/src/components/Description.test.tsx
@@ -2,24 +2,26 @@ import { screen } from "@testing-library/react";
 import Description from "components/Description";
 import { renderWithTheme } from "mocks/utils";
 
-test("display description with just text", async () => {
-  renderWithTheme(<Description text="test description" />);
+describe("Description", () => {
+  test("display description with just text", async () => {
+    renderWithTheme(<Description text="test description" />);
 
-  screen.getByText("test description");
-});
+    screen.getByText("test description");
+  });
 
-test("display description with both link and text", async () => {
-  renderWithTheme(
-    <Description
-      text="test description"
-      link="https://servicenow.github.io/azimuth/"
-    />
-  );
+  test("display description with both link and text", async () => {
+    renderWithTheme(
+      <Description
+        text="test description"
+        link="https://servicenow.github.io/azimuth/"
+      />
+    );
 
-  screen.getByText("test description");
+    screen.getByText("test description");
 
-  const link: HTMLAnchorElement = screen.getByRole("link");
-  expect(link.href).toContain("https://servicenow.github.io/azimuth/");
-  screen.getByText("Learn more");
-  screen.getByTestId("LinkIcon");
+    const link: HTMLAnchorElement = screen.getByRole("link");
+    expect(link.href).toContain("https://servicenow.github.io/azimuth/");
+    screen.getByText("Learn more");
+    screen.getByTestId("OpenInNewIcon");
+  });
 });
--- a/webapp/src/components/Description.test.tsx
+++ b/webapp/src/components/Description.test.tsx
@@ -3,13 +3,13 @@ import Description from "components/Description";
 import { renderWithTheme } from "mocks/utils";
 
 describe("Description", () => {
-  test("display description with just text", async () => {
+  it("should display description with text only", async () => {
     renderWithTheme(<Description text="test description" />);
 
     screen.getByText("test description");
   });
 
-  test("display description with both link and text", async () => {
+  it("should display description with both link and text", async () => {
     renderWithTheme(
       <Description
         text="test description"
--- a/webapp/src/pages/ClassOverlap.tsx
+++ b/webapp/src/pages/ClassOverlap.tsx
@@ -23,13 +23,6 @@ import { getClassOverlapPlotEndpoint } from "services/api";
 import { QueryClassOverlapState } from "types/models";
 import { constructSearchString } from "utils/helpers";
 
-export const classAnalysisDescription = (
-  <Description
-    text="Assess overlap between class pairs."
-    link="user-guide/class-analysis/"
-  />
-);
-
 const OVERLAP_THRESHOLD_INPUT_PROPS = { step: 0.01, min: 0, max: 1 };
 
 const ClassOverlap = () => {
@@ -78,7 +71,7 @@ const ClassOverlap = () => {
         {
           <Description
             text="Assess semantic overlap between class pairs."
-            link="/class-overlap/"
+            link="user-guide/class-overlap/"
           />
         }
       </Box>
@@ -114,7 +107,7 @@ const ClassOverlap = () => {
               text={
                 "Assess magnitude of overlap and select class pairs to explore further. For suggested workflow:"
               }
-              link="/class-overlap/#suggested_workflow/"
+              link="user-guide/class-overlap/#suggested_workflow/"
             />
             <Typography variant="body2" marginTop={0.25}>
               Flows between class nodes indicate whether a source class's
--- a/webapp/src/pages/ClassOverlap.tsx
+++ b/webapp/src/pages/ClassOverlap.tsx
@@ -23,6 +23,13 @@ import { getClassOverlapPlotEndpoint } from "services/api";
 import { QueryClassOverlapState } from "types/models";
 import { constructSearchString } from "utils/helpers";
 
+export const classAnalysisDescription = (
+  <Description
+    text="Assess overlap between class pairs."
+    link="user-guide/class-analysis/"
+  />
+);
+
 const OVERLAP_THRESHOLD_INPUT_PROPS = { step: 0.01, min: 0, max: 1 };
 
 const ClassOverlap = () => {
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -12,9 +12,9 @@ import {
   Tooltip,
   Typography,
 } from "@mui/material";
+import AccordionLayout from "components/AccordionLayout";
 import _ from "lodash";
 import React from "react";
-import AccordionLayout from "components/AccordionLayout";
 import { useParams } from "react-router-dom";
 import { getConfigEndpoint, updateConfigEndpoint } from "services/api";
 import { AzimuthConfig, PipelineDefinition } from "types/api";
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -154,6 +154,47 @@ const Settings: React.FC = () => {
     />
   );
 
+  const displayArgumentsList = (
+    name: string,
+    argEntries: Record<string, string> | Record<string, string>[]
+  ) => (
+    <Box display="flex" flexDirection="column" paddingTop={1}>
+      <Typography variant="caption">{name}</Typography>
+      {Object.entries(argEntries).map(([field, value], index) => (
+        <Box
+          key={index}
+          display="flex"
+          flexDirection="row"
+          justifyContent="flex-start"
+        >
+          <Typography
+            variant="body2"
+            sx={{
+              s: 1,
+              width: "100%",
+            }}
+          >
+            {field}:
+          </Typography>
+          <Tooltip title={value} placement="bottom">
+            <Typography
+              variant="body2"
+              sx={{
+                s: 1,
+                width: "20ch",
+                whiteSpace: "nowrap",
+                overflow: "hidden",
+                textOverflow: "ellipsis",
+              }}
+            >
+              {value}
+            </Typography>
+          </Tooltip>
+        </Box>
+      ))}
+    </Box>
+  );
+
   const displayReadonlyFields = (label: string, value: string | undefined) => (
     <TextField
       sx={{ m: 1, width: "40ch" }}
@@ -336,85 +377,11 @@ const Settings: React.FC = () => {
           resultingConfig.dataset?.class_name
         )}
         {displayReadonlyFields("remote", resultingConfig.dataset?.remote)}
-        {resultingConfig.dataset?.kwargs && (
-          <Box display="flex" flexDirection="column" paddingTop={1}>
-            <Typography variant="caption">kwargs</Typography>
-            {Object.entries(resultingConfig.dataset.kwargs).map(
-              ([field, value], index) => (
-                <Box
-                  key={index}
-                  display="flex"
-                  flexDirection="row"
-                  justifyContent="flex-start"
-                >
-                  <Typography
-                    variant="body2"
-                    sx={{
-                      s: 1,
-                      width: "100px",
-                    }}
-                  >
-                    {field}:
-                  </Typography>
-                  <Tooltip title={value} placement="bottom">
-                    <Typography
-                      variant="body2"
-                      sx={{
-                        s: 1,
-                        width: "100px",
-                        whiteSpace: "nowrap",
-                        overflow: "hidden",
-                        textOverflow: "ellipsis",
-                      }}
-                    >
-                      {value}
-                    </Typography>
-                  </Tooltip>
-                </Box>
-              )
-            )}
-          </Box>
-        )}
+        {resultingConfig.dataset?.kwargs &&
+          displayArgumentsList("kwargs", resultingConfig.dataset.kwargs)}
         {resultingConfig.dataset?.args &&
-          resultingConfig.dataset.args.length > 0 && (
-            <Box display="flex" flexDirection="column" paddingTop={1}>
-              <Typography variant="caption">args</Typography>
-              {Object.entries(resultingConfig.dataset.args).map(
-                ([field, value], index) => (
-                  <Box
-                    key={index}
-                    display="flex"
-                    flexDirection="row"
-                    justifyContent="flex-start"
-                  >
-                    <Typography
-                      variant="body2"
-                      sx={{
-                        s: 1,
-                        width: "100px",
-                      }}
-                    >
-                      {field}:
-                    </Typography>
-                    <Tooltip title={value} placement="bottom">
-                      <Typography
-                        variant="body2"
-                        sx={{
-                          s: 1,
-                          width: "100px",
-                          whiteSpace: "nowrap",
-                          overflow: "hidden",
-                          textOverflow: "ellipsis",
-                        }}
-                      >
-                        {value}
-                      </Typography>
-                    </Tooltip>
-                  </Box>
-                )
-              )}
-            </Box>
-          )}
+          resultingConfig.dataset.args.length > 0 &&
+          displayArgumentsList("args", resultingConfig.dataset.args)}
       </Box>
     </Box>
   );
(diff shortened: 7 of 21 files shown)


nandhinibsn:
Recent screenshot of the config page:
image
image

webapp/src/pages/Settings.tsx
```diff
@@ -41,18 +46,191 @@ const Settings: React.FC = () => {
...
+                "Long sentence min token",
+                "long_sentence_min_token",
+                { min: 0, max: 100, step: 1 },
+                resultingConfig.syntax?.long_sentence_min_token
+              )}
+            </Box>
+          )}
+          <FormControlLabel
+            control={switchNullOrDefault("uncertainty")}
+            label="Uncertainity"
```
JosephMarinier, line 202:
```suggestion
            label="Uncertainty"
```

webapp/src/pages/Settings.tsx
```diff
@@ -8,12 +9,18 @@ import {
...
+  Typography,
 } from "@mui/material";
 import React from "react";
 import { useParams } from "react-router-dom";
 import { getConfigEndpoint, updateConfigEndpoint } from "services/api";
 import { AzimuthConfig } from "types/api";
+import { Info } from "@mui/icons-material";
+
+const DOCS_URL =
+  "https://servicenow.github.io/azimuth/main/reference/configuration/";
```
JosephMarinier, line 23: I think the part `https://servicenow.github.io/azimuth/main` part should be saved in a constant and reused for `DOCS_URL` in `Description.tsx`.

nandhinibsn:
Hey, @JosephMarinier and @gabegma, To disable the postprocessors, the type should allow null to go through the patch req. Right now, the postprocessors type allows only these (Partial<components["schemas"]["TemperatureScaling"]> &
        Partial<components["schemas"]["ThresholdConfig"]> &
        Partial<components["schemas"]["CustomObject"]>)[]

webapp/src/pages/Settings.tsx
```diff
@@ -41,38 +69,434 @@ const Settings: React.FC = () => {
...
+  ) => (
+    <TextField
+      sx={{ width: FIELDS_WIDTH, marginLeft: 5, marginBottom: 2 }}
+      id={field}
+      label={label}
+      type="number"
+      InputLabelProps={{
+        shrink: true,
+      }}
+      defaultValue={value}
```
JosephMarinier, line 121: We should always "control" our inputs, and never use `defaultValue`. Otherwise, if a field is affected by any other input, like a reset button, a relead button, or even the URL, then it would not work.
```suggestion
      value={value}
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,6 +45,20 @@ const Settings: React.FC = () => {
 
   const resultingConfig = { ...data, ...partialConfig };
```
JosephMarinier, line 46: I suspect this would also need `_.merge()`.
```suggestion
  const resultingConfig = _.merge(data, partialConfig);
```

webapp/src/pages/Settings.tsx
JosephMarinier: Don't forget this. I would suggest `webapp/src/utils/const.ts`.

webapp/src/components/AccordianLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Description from "components/Description";
+
+type Props = {
+  name: string;
+  description: string;
+  docURL: string;
+  children: React.ReactNode;
+};
+
+const AccordianLayout: React.FC<Props> = ({
```
JosephMarinier, line 20: Oups! `Accordion` with an `o`.
```suggestion
const AccordionLayout: React.FC<Props> = ({
```

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,618 @@ import {
...
+  );
+
+  const handleCustomMetricUpdate = (checked: boolean, metricName: string) => {
+    checked
+      ? setPartialConfig({
+          ...partialConfig,
+          ...{
+            metrics: _.merge({}, resultingConfig?.metrics, {
+              [metricName]: {
+                class_name: "datasets.load_metric",
```
JosephMarinier, line 287: For precision, recall and f1, you will also need to provide `additional_kwargs: {average: "weighted"}`, but not for accuracy.

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,618 @@ import {
...
+              ),
+            })
+          : setPartialConfig({
+              ...partialConfig,
+              pipelines: _.unionBy(
+                updatedPipeline,
+                resultingConfig?.pipelines,
+                "name"
+              ),
+            });
```
JosephMarinier, line 275: I think this is equivalent to this simpler form:
```suggestion
        setPartialConfig({
          ...partialConfig,
          pipelines: _.unionBy(
            updatedPipeline,
            existedPipeline ? partialConfig.pipelines : resultingConfig.pipelines,
            "name"
          ),
        });
```

JosephMarinier: By the way, sorry for repeating, but you can remove the 19 unnecessary `?`s after `resultingConfig`, since it is always defined.

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,618 @@ import {
...
+      defaultValue={value}
+      inputProps={{ min: 0, max: 1, step: 0.1 }}
+      variant="standard"
+      onChange={(event) => {
+        const existedPipeline = partialConfig.pipelines
+          ? partialConfig.pipelines.find(({ name }) => name === pipeline.name)
+          : undefined;
+        const current_postprocessor = pipeline?.postprocessors?.find(
+          (_, index) => index === postprocessorIdx
+        );
```
JosephMarinier, line 245: No need to `find`, just use the index.
```suggestion
        const current_postprocessor = pipeline.postprocessors?.[postprocessorIdx];
```

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,618 @@ import {
...
+        const updatedPostprocessor = _.assign({}, current_postprocessor, {
+          [field]: Number(event.target.value),
+        });
```
JosephMarinier, line 248: This use of `_.assign` only results in a shallow copy (not deep), which might be fine, so it is equivalent to
```suggestion
        const updatedPostprocessor = {
          ...currentPostprocessor,
          [field]: Number(event.target.value),
        };
```

webapp/src/pages/Settings.tsx
JosephMarinier: Second reading... It seems overkill to search for the pipeline with `_.unionBy(..., "name")`. Why not pass the pipeline index to `displayPostprocessorSubField`?

webapp/src/pages/Settings.tsx
JosephMarinier: Don't forget this two-week-old comment.

webapp/src/pages/Settings.tsx
JosephMarinier: Your solution to that is excellent. 👍

webapp/src/components/AccordianLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+  Divider,
+  Typography,
+} from "@mui/material";
+import React from "react";
+import Description from "components/Description";
+
+type Props = {
+  name: string;
+  description: string;
+  docURL: string;
```
JosephMarinier, line 16: Just to be rigorous, this is not a URL. It's just part of the path, which is part of the complete URL. I would suggest calling it `link` to stay consistent with the `Description`'s prop.
```suggestion
  link: string;
```

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,618 @@ import {
...
+                })}
+                <Box
+                  key={index}
+                  display="flex"
+                  flexDirection="column"
+                  gap={1}
+                  marginLeft={2}
+                >
+                  {postprocessors &&
+                    postprocessors.map((postprocessor, index) => (
```
JosephMarinier, line 483: Here the `?` would be best.
```suggestion
                  {postprocessors?.map((postprocessor, index) => (
```

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,618 @@ import {
...
+          View and edit certain fields from your config file. Once your changes
+          are saved, expect some delays for recomputing the affected tasks.
+        </Typography>
+        {configSection.map((section, index) => (
+          <AccordianLayout key={index} {...section}>
+            {index === 0 && getProjectConfigSection()}
+            {index === 1 && getModalContractConfigSection()}
+            {index === 2 && getAnalysesCustomization()}
+          </AccordianLayout>
+        ))}
```
JosephMarinier, line 606: I would recommend simply defining the 3 sections in-line here. This will save
- the `section` type (which by the way should have been capitalized);
- the `configSection` dictionary;
- this loop;
- those 3 conditional renders with `&&`; and
- in total 12 lines of code.

```diff
-type section = {
-  name: string;
-  description: string;
-  docURL: string;
-};
-
-const configSection: section[] = [
-  {
-    name: "Project Config",
-    description:
-      "Contains mandatory fields that specify the dataset to load in Azimuth",
-    docURL: "reference/configuration/project/",
-  },
-  {
-    name: "Model Contract Config",
-    description:
-      "Defines how Azimuth interacts with the ML pipelines and the metrics",
-    docURL: "reference/configuration/model_contract/",
-  },
-  {
-    name: "Analyses Customization",
-    description: "Four analyses configured in Azimuth",
-    docURL: "reference/configuration/analyses/",
-  },
-];
-
```
```suggestion
        <AccordianLayout
          name="Project Config"
          description="Contains mandatory fields that specify the dataset to load in Azimuth"
          docURL="reference/configuration/project/"
        >
          {getProjectConfigSection()}
        </AccordianLayout>
        <AccordianLayout
          name="Model Contract Config"
          description="Defines how Azimuth interacts with the ML pipelines and the metrics"
          docURL="reference/configuration/model_contract/"
        >
          {getModalContractConfigSection()}
        </AccordianLayout>
        <AccordianLayout
          name="Analyses Customization"
          description="Four analyses configured in Azimuth"
          docURL="reference/configuration/analyses/"
        >
          {getAnalysesCustomization()}
        </AccordianLayout>
```

webapp/src/pages/Settings.tsx
```diff
@@ -2,58 +2,623 @@ import {
...
+        sx={{
+          display: "flex",
+          justifyContent: "space-between",
+          alignItems: "center",
+          paddingY: 2,
+        }}
+      >
+        <Button
+          variant="contained"
+          onClick={() => setPartialConfig({ ...data })}
```
JosephMarinier, line 618: The `partialConfig` should always only contain what's new. Here, `data` is the full config. Let's never assign it to the partial config. Instead, discarding the changes is as simple as resetting `partialConfig` to its initial value: an empty object.
```suggestion
          onClick={() => setPartialConfig({})}
```

webapp/src/pages/Settings.tsx
nandhinibsn: I have already taken care of this commit, But unfortunately, the field behaves like read only if I change to value attribute. I am not able to find the reason why. I am figuring out as well why it is behaving this way. I could also see that you used text field in class overlap component. have you faced this issue earlier?

webapp/src/pages/Settings.tsx
nandhinibsn: Okay, I added the dict to write it short. But,  true is the suggestion to avoid declaration, conditional rendering and loops.

webapp/src/pages/Settings.tsx
nandhinibsn: I understood that the reset won't work if we use the "defaultValue" attribute for the text field. But, I mentioned in the previous comment's reply that the value attribute is making the text field link "read-only". So, I am trying to analyze and fix it.

webapp/src/pages/Settings.tsx
JosephMarinier: In the future, if you try to apply a review suggestion and you get an issue, please reply (in less than 2 weeks) so we can look at the issue together.
At a glance, I have no idea what issue you faced. Let's Zoom tomorrow to try it together?

webapp/src/components/AccordionLayout.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+describe("AccordionLayout", () => {
+  it("should display name, description and description component on the accordian layout", () => {
+    renderAccordionLayout({
+      name: "Project Config",
+      description:
+        "Contains mandatory fields that specify the dataset to load in Azimuth",
+      link: "reference/configuration/project/",
+      children: <Typography>Project Config Section</Typography>,
+    });
```
JosephMarinier, line 24: I don't see value in `renderAccordionLayout()`, as well as duplicating the definition of `Props` here. React components are already meant to be rendered, no need to wrap them in functions. Here is what I'd do. Note the "accordian" misspelled with an "a".
```suggestion
describe("AccordionLayout", () => {
  it("should display name, description and description component on the accordion layout", () => {
    renderWithTheme(
      <AccordionLayout
        name="Project Config"
        description="Contains mandatory fields that specify the dataset to load in Azimuth"
        link="reference/configuration/project/"
      >
        <Typography>Project Config Section</Typography>
      </AccordionLayout>
    );
```

webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
```diff
@@ -10,13 +10,13 @@ const DESCRIPTION = {
   "General Warnings": (
     <Description
       text="Investigate issues with class size, class imbalance, and dataset shift."
-      link="/dataset-warnings/#general-warnings"
```
JosephMarinier, line 13: Good catch! 👍

webapp/src/components/Description.test.tsx
```diff
@@ -2,24 +2,26 @@ import { screen } from "@testing-library/react";
 import Description from "components/Description";
 import { renderWithTheme } from "mocks/utils";
 
-test("display description with just text", async () => {
-  renderWithTheme(<Description text="test description" />);
+describe("Description", () => {
+  test("display description with just text", async () => {
```
JosephMarinier, line 6: That's just a detail, but if you switch to the `describe()` pattern, it usually goes well with the `it("should ...` pattern.
```suggestion
  it("should display description with text only", async () => {
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,29 +62,540 @@ const Settings: React.FC = () => {
...
+                <Box key={index} gap={(theme) => theme.spacing(1)}>
+                  {displayReadonlyFields(field, String(value))}
+                </Box>
+              )
+            )}
+          </Box>
+        )}
+    </Box>
+  );
+  const getModalContractConfigSection = () => (
```
JosephMarinier, line 332: I think you mean "model", and not "modal".
```suggestion
  const getModelContractConfigSection = () => (
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,29 +62,540 @@ const Settings: React.FC = () => {
...
+            ([field, value], index) => (
+              <Box key={index} gap={(theme) => theme.spacing(1)}>
+                {displayReadonlyFields(field, String(value))}
+              </Box>
+            )
+          )}
+        </Box>
+      )}
+      {resultingConfig.dataset?.args &&
+        resultingConfig.dataset?.args.length > 0 && (
```
JosephMarinier, line 318: Detail, but we could shorten that up by relying on the fact that `0` is falsy, just like `undefined`.
```suggestion
      {resultingConfig.dataset?.args?.length && (
```

webapp/src/pages/Settings.tsx
```diff
@@ -2,27 +2,58 @@ import {
...
+import _ from "lodash";
 import React from "react";
+import AccordionLayout from "components/AccordionLayout";
 import { useParams } from "react-router-dom";
 import { getConfigEndpoint, updateConfigEndpoint } from "services/api";
-import { AzimuthConfig } from "types/api";
+import { AzimuthConfig, PipelineDefinition } from "types/api";
+
+type stepper = { [key: string]: InputBaseComponentProps };
+const STEPPER: stepper = {
```
JosephMarinier, line 21: Types should be capitalized. Anyway, I think we don't need to save the `type`, we can use it directly:
```suggestion
const STEPPER: Record<string, InputBaseComponentProps> = {
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,29 +62,540 @@ const Settings: React.FC = () => {
...
+        checked={Boolean(pipeline.postprocessors)}
+        disabled={isError || isFetching}
+        onChange={(event) => {
+          setPartialConfig({
+            ...partialConfig,
+            pipelines: _.unionBy(
+              [
+                {
+                  ...pipeline,
+                  postprocessors: event.target.checked ? [] : null,
```
JosephMarinier, line 112: The `Checkbox`'s `onChange` handler provides a second argument which is directly the `checked` boolean, so no need to mess with `event.target`:
```suggestion
        onChange={(_, checked) => {
          setPartialConfig({
            ...partialConfig,
            pipelines: _.unionBy(
              [
                {
                  ...pipeline,
                  postprocessors: checked ? [] : null,
```

nandhinibsn: Okay, thank you for the suggestion here. I had to write it that way because the "_" confuse between onchange attribute and lodash attributes.

webapp/src/pages/Settings.tsx
nandhinibsn: Oh, Yes. thanks for this suggestion :)

webapp/src/components/DatasetWarnings/DatasetDistribution.tsx
nandhinibsn: thank you!

webapp/src/components/AccordionLayout.test.tsx
nandhinibsn: Earlier I planned to write different test cases for title and description, so I declared this common function. But it makes sense to me now to declare it inside, as there is one test case inside the describe.

webapp/src/pages/Settings.tsx
nandhinibsn: I could find the reason behind the issue with the value attribute and pushed the fix.

webapp/src/pages/Settings.tsx
nandhinibsn: So, I tried simplifying the line to check the empty array, But 0 is rendered for the following reason in JS [For reference](https://javascript.plainenglish.io/javascript-tip-why-0-is-rendered-when-using-the-logical-and-operator-with-array-length-aae24d05c9d7). So, I reverted the change here. Please comment if you find any other simplified way to write this.

nandhinibsn:
Screenshot of the latest work after adding design review comments: Please let me know your feedback @gabegma @JosephMarinier 
image
image
image
image

webapp/src/pages/Settings.tsx
JosephMarinier: Ha, that's true! Because we are inside a JSX content. Good catch! My solution would be to cast it to a `Boolean`. This has the same result as this article's suggestion to add a double `!!` in front, but I find it cleaner. What do you think?
```suggestion
      {Boolean(resultingConfig.dataset?.args?.length) && (
```

webapp/src/pages/Settings.tsx
JosephMarinier: Of course! I think that's a cool trick with the `...[]`. This is my most up-voted [answer in stack overflow](https://stackoverflow.com/questions/32197927/standard-conventions-for-indicating-a-function-argument-is-unused-in-javascript/58738236#58738236).

If you're curious, another neat solution is to destructure the unused argument into nothing with `{}`. However, this is less general as it doesn't work if the argument can be `undefined` or `null`. So, I prefer `...[]`.
```diff
-        onChange={(...[, checked]) => {
+        onChange={({}, checked) => {
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+                  : {},
+              },
+            }),
+          },
+        })
+      : setPartialConfig({
+          ...partialConfig,
+          ...{
+            metrics: _.omit(resultingConfig.metrics, metricName),
+          },
```
JosephMarinier, line 313:
```suggestion
          metrics: _.omit(resultingConfig.metrics, metricName),
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+    field: string,
+    value: number | undefined
+  ) => (
+    <TextField
+      id={field}
+      size="small"
+      label={<Typography fontWeight="bold">{field}</Typography>}
+      type="number"
+      value={value}
+      disabled={!Boolean(resultingConfig[config])}
```
JosephMarinier, line 225: No need for `Boolean` if you are inverting with `!`.
```suggestion
      disabled={!resultingConfig[config]}
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+                  >
+                    {field}:
+                  </Typography>
+                  <TextField
+                    id={field}
+                    sx={{ s: 1, width: "5ch" }}
+                    size="small"
+                    type="number"
+                    value={value}
+                    disabled={!Boolean(resultingConfig.uncertainty)}
```
JosephMarinier, line 428: Same
```suggestion
                    disabled={!resultingConfig.uncertainty}
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+                    style={{ fontSize: 14 }}
+                    inputProps={
+                      (STEPPER[field],
+                      { "data-testid": `${field}` },
+                      { style: { fontSize: 14 } })
+                    }
```
JosephMarinier, line 434: The comma operator, inside those parentheses, means that `STEPPER[field]` and `{ "data-testid": `${field}` }` are completely ignored, returning only `{ style: { fontSize: 14 } }`. This should be:
```suggestion
                    inputProps={{
                      ...STEPPER[field],
                      "data-testid": `${field}`,
                      style: { fontSize: 14 },
                    }}
```

webapp/src/pages/Settings.tsx
```diff
@@ -1,28 +1,79 @@
...
+  high_epistemic_threshold: { min: 0, max: 1, step: 0.1 },
+  conflicting_neighbors_threshold: { min: 0, max: 1, step: 0.1 },
+  no_close_threshold: { min: 0, max: 1, step: 0.1 },
+  min_num_per_class: { min: 0, max: 100, step: 10 },
+  max_delta_class_imbalance: { min: 0, max: 1, step: 0.1 },
+  max_delta_representation: { min: 0, max: 0.1, step: 0.01 },
+  max_delta_mean_tokens: { min: 0.0, max: 10.0, step: 1.0 },
+  max_delta_std_tokens: { min: 0.0, max: 10.0, step: 1.0 },
+  short_sentence_max_token: { min: 0, max: 100, step: 1 },
+  long_sentence_min_token: { min: 0, max: 100, step: 1 },
```
JosephMarinier, line 33: All those stepper properties have no effect now. The first two don't because of [that bug](https://github.com/ServiceNow/azimuth/pull/282/files#r1042671678), and all the others don't seem to be used at all. For example, if you click once on `conflicting_neighbors_threshold ` to go up, it goes from `0.9` to `1.9`, which is not a valid value.

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+      variant="standard"
+      onChange={(event) =>
+        setPartialConfig(
+          _.merge({}, partialConfig, {
+            [config]: _.merge(
+              { [field]: Number(event.target.value) },
+              config in partialConfig
+                ? _.get(partialConfig, config)
+                : _.get(resultingConfig, config)
+            ),
```
JosephMarinier, line 235: This doesn't work at the moment. The number pickers in the Analyses Customization are not editable. Multiple things:
1. Is there a reason for using `_.get(resultingConfig, config)` instead of native `resultingConfig[config]`?
2. About the ternary operator: if something is `in partialConfig`, it will automatically be in `resultingConfig`. That's the purpose of `resultingConfig`: as a merge of `partialConfig` over `data`. So, this is equivalent to
```diff
-              config in partialConfig
-                ? _.get(partialConfig, config)
-                : _.get(resultingConfig, config)
+              resultingConfig.config
```
3. The order in your `_.merge` is important. You are currently overwriting the new value with `resultingConfig.config`, so it never takes effect. They should be merged in inverse order. In that case, as you have correctly done elsewhere, you need to pass `{}` first to avoid modifying the `resultingConfig` in place.
4. A `_.assign` would be sufficient here, as we don't need a recursive `_.merge`.
```suggestion
            [config]: _.assign({}, resultingConfig[config], {
              [field]: Number(event.target.value),
            }),
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+                  ...pipeline,
+                  postprocessors: checked
+                    ? _.find(data?.pipelines, ["name", pipeline.name])
+                        ?.postprocessors ?? []
+                    : null,
+                },
+              ],
+              _.has(partialConfig, "pipelines")
+                ? partialConfig.pipelines
+                : resultingConfig.pipelines,
```
JosephMarinier, line 130: For the same reason:
```suggestion
              resultingConfig.pipelines,
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+                          postprocessor.class_name
+                        )}
+                        {postprocessor.temperature &&
+                          displayPostprocessorNumberField(
+                            { name, model, postprocessors },
+                            "temperature",
+                            index,
+                            postprocessor.temperature
+                          )}
+                        {postprocessor.threshold &&
```
JosephMarinier, line 533: If you type 0 in the field, it disappears. You must check for `undefined`, not just any falsy value.
```suggestion
                        {postprocessor.temperature !== undefined &&
                          displayPostprocessorNumberField(
                            { name, model, postprocessors },
                            "temperature",
                            index,
                            postprocessor.temperature
                          )}
                        {postprocessor.threshold !== undefined &&
```

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+        >
+          {FIELDS_TRIGGERING_STARTUP_TASKS.some((f) => partialConfig[f]) && (
+            <FormHelperText>
+              <Warning
+                color="primary"
+                sx={{
+                  position: "relative",
+                  top: 3,
+                  right: 0.5,
+                  color: (theme) => theme.palette.warning.contrastText,
```
JosephMarinier, line 687: I would suggest setting `color="warning"` and not overwriting the `color` in `sx`.
```suggestion
                color="warning"
                sx={{
                  position: "relative",
                  top: 3,
                  right: 0.5,
```

webapp/src/pages/Settings.tsx
nandhinibsn: Ohh, right!.. checking only undefined here is appropriate

webapp/src/pages/Settings.tsx
nandhinibsn: As we discussed, I have added the suggestion.

webapp/src/pages/Settings.tsx
```diff
@@ -31,48 +82,625 @@ const Settings: React.FC = () => {
...
+                    }
+                    variant="standard"
+                    onChange={(event) =>
+                      setPartialConfig(
+                        _.merge({}, partialConfig, {
+                          uncertainty: _.merge(
+                            { [field]: Number(event.target.value) },
+                            _.has(partialConfig, "uncertainty")
+                              ? {}
+                              : _.get(resultingConfig, "uncertainty")
```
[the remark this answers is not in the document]
nandhinibsn: As we discussed, I have added spread JS syntax all over and got rid of lodash attributes but, just used _.omit in the custom metrics delete scenario.

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,699 @@
...
+                      ...STEPPER[field],
+                      style: { fontSize: 14 },
+                    }}
+                    variant="standard"
+                    onChange={(event) =>
+                      setPartialConfig({
+                        ...partialConfig,
+                        uncertainty: _.assign({}, resultingConfig.uncertainty, {
+                          [field]: Number(event.target.value),
+                        }),
```
JosephMarinier, line 437: As seen in my suggestion https://github.com/ServiceNow/azimuth/pull/282#discussion_r1042739485, you can get rid of lodash completely, here:
```suggestion
                        uncertainty: {
                          ...resultingConfig.uncertainty,
                          [field]: Number(event.target.value),
                        },
```

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,699 @@
...
+          })
+        }
+      />
+    </Box>
+  );
+
+  const switchNullOrDefault = (
+    field: keyof AzimuthConfig,
+    selected: boolean
+  ) => (
```
JosephMarinier, line 146: 3 small things here:
1. You added the `selected` argument, but I think it's redundant with passing `field`. The `checked={Boolean(resultingConfig[field])}` would still work.
2. Since you changed the behavior from setting `{}` to setting `config![field] ?? CONFIG_SUB_FIELDS[field]`, it doesn't only `switchNullOrDefault`, so the name is outdated.
3. Considering all that, as well as the fact that this function is now used in only one place, I would simply get rid of this function and integrate the `<Checkbox />` directly in `displayToggleSectionTitle`.

Make sense?

nandhinibsn: Okay right!. my opinion is to have separate function for code readability. so, If we keep this function, would the name "enableResultingConfig" be appropriate?

nandhinibsn: I moved the checkbox logic as suggested to the "displayToggleSection"

JosephMarinier: > separate function for code readability.

That's a very good mindset, in general. 👍 But with a bad name, it doesn't help readability. I don't see how `enableResultingConfig()` is better.
Also, I notice `displayPostprocessorToggleSection()` defines its `Checkbox` inline, while `displayToggleSectionTitle()` uses this `switchNullOrDefault()` which returns a very similar checkbox. There is no reason for those two to be different. I'm still leaning towards integrating `switchNullOrDefault()` inside `displayToggleSectionTitle()` to make it uniform with `displayPostprocessorToggleSection()`.

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,691 @@
...
   FormHelperText,
+  InputBaseComponentProps,
+  Paper,
+  TextField,
+  Tooltip,
+  Typography,
 } from "@mui/material";
+import _ from "lodash";
 import React from "react";
+import AccordionLayout from "components/AccordionLayout";
```
JosephMarinier, line 17: Detail
```suggestion
import AccordionLayout from "components/AccordionLayout";
import _ from "lodash";
import React from "react";
```

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,691 @@
...
+        ...STEPPER[field],
+        style: { fontSize: 14 },
+      }}
+      disabled={!resultingConfig[config]}
+      variant="standard"
+      onChange={(event) =>
+        setPartialConfig({
+          ...partialConfig,
+          [config]: {
+            ...(resultingConfig[config] as Partial<AzimuthConfig>),
```
JosephMarinier, line 234: Interesting... I get why we needed that `as` here, but I'm not sure why `Partial<AzimuthConfig>` works... It's wrong! That's one of the objects nested inside the config, not the whole config. One more reason to hate `as`! 😝 
Here is a clan solution to get rid of it: #336

webapp/src/pages/Settings.tsx
nandhinibsn: Okay, I have no proper clarity on the order of arrangement. is there any specific rules to order the imports? It would be helpful for me to follow.

nandhinibsn: Okay, I have no proper clarity on the order of arrangement. is there any specific rules to order the imports? It would be helpful for me to follow.

JosephMarinier: Oh! It's in alphabetical order of the "module name" (the bit after `from`). I'm just following the order VSCode does automatically when you ask it to add an import.

webapp/src/pages/Settings.tsx
nandhinibsn: why not? is not one of the partial object in Azimuthconfig? is my understanding about partial right?

JosephMarinier: Check out [`Partial`'s documentation](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype).
Basically, `Partial<AzimuthConfig>` is the same as `AzimuthConfig`, but with all fields being optional.
For example, if we have
```ts
type AzimuthConfig = {
  name?: string;
  dataset: components["schemas"]["CustomObject"];
  model_contract?: components["schemas"]["SupportedModelContract"];
};
```
then `Partial<AzimuthConfig>` looks like this with an extra `?`.
```ts
{
  name?: string;
  dataset?: components["schemas"]["CustomObject"];
  model_contract?: components["schemas"]["SupportedModelContract"];
}
```
However, what `resultingConfig[config]` is rather is something like
```ts
components["schemas"]["CustomObject"] | components["schemas"]["SupportedModelContract"]
```
Quite different.

nandhinibsn: Okay, Understood. Thanks!

webapp/src/components/AccordionLayout.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {
+  Accordion,
+  AccordionDetails,
+  AccordionSummary,
+  Box,
+  Divider,
+  Typography,
+} from "@mui/material";
+import React from "react";
+import Description from "components/Description";
```
JosephMarinier, line 11:
```suggestion
import Description from "components/Description";
import React from "react";
```

webapp/src/components/AccordionLayout.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <Description link={link} />
+        </Box>
+      </AccordionSummary>
+      <AccordionDetails>
+        <Divider sx={{ marginBottom: 1 }} />
+        {children}
+      </AccordionDetails>
+    </Accordion>
+  );
+};
```
JosephMarinier, line 51: You can save one indent here:
```suggestion
}) => (
  <Accordion>
    <AccordionSummary expandIcon={<ExpandMoreIcon />} id={name}>
      <Typography
        variant="body1"
        fontWeight="bold"
        sx={{
          width: "20%",
          flexShrink: 0,
        }}
      >
        {name}
      </Typography>
      <Divider orientation="vertical" flexItem sx={{ marginX: 2 }} />
      <Box display="flex" gap={2}>
        <Typography variant="body1">{description}</Typography>
        <Description link={link} />
      </Box>
    </AccordionSummary>
    <AccordionDetails>
      <Divider sx={{ marginBottom: 1 }} />
      {children}
    </AccordionDetails>
  </Accordion>
);
```

nandhinibsn:
@gabegma ,Can you please confirm the description text for the [tabs here](https://github.com/ServiceNow/azimuth/blob/nandhini/expandConfigPage/webapp/src/pages/Settings.tsx#L627-L645)

webapp/src/components/AccordionLayout.tsx
nandhinibsn: Interesting! does specify the "return" keyword explicitly not required?

JosephMarinier: Not if you don't user brackets for the body of the arrow function. In `Settings.tsx`, You used 21 such arrow functions with an implicit return.

JosephMarinier:
Can you log the change in `CHANGELOG.md` and check the checklist in the PR description?

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,694 @@
...
+        <Box
+          key={index}
+          display="flex"
+          flexDirection="row"
+          justifyContent="flex-start"
+        >
+          <Typography
+            variant="body2"
+            sx={{
+              s: 1,
```
JosephMarinier, line 175: I see this `s` style being applied all over the place, but it seems invalid. What is it supposed to do?
image

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,694 @@
...
+                    {field}:
+                  </Typography>
+                  <TextField
+                    id={field}
+                    sx={{ s: 1, width: "5ch" }}
+                    size="small"
+                    type="number"
+                    value={value}
+                    disabled={!resultingConfig.uncertainty}
+                    style={{ fontSize: 14 }}
```
JosephMarinier, line 425: I think this has no effect. The one in `inputProps` does.
```suggestion
```

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,694 @@
...
+      key={postprocessorIdx}
+      sx={{
+        m: 1,
+        width: "15ch",
+      }}
+      size="small"
+      label={<Typography fontWeight="bold">{field}</Typography>}
+      type="number"
+      value={value}
+      inputProps={{ min: 0, max: 1, step: 0.1 }}
```
JosephMarinier, line 262: That's the only input that has the default font size of 16. To make it like the others:
```suggestion
      inputProps={{ min: 0, max: 1, step: 0.1, style: { fontSize: 14 } }}
```

webapp/src/pages/ClassOverlap.tsx
```diff
@@ -23,6 +23,13 @@ import { getClassOverlapPlotEndpoint } from "services/api";
 import { QueryClassOverlapState } from "types/models";
 import { constructSearchString } from "utils/helpers";
 
+export const classAnalysisDescription = (
+  <Description
+    text="Assess overlap between class pairs."
+    link="user-guide/class-analysis/"
+  />
+);
+
```
JosephMarinier, line 32: I think this is a merge conflict. It is not used anywhere. If you search for `link="/class-overlap` in the repo, you should find 3 links to the documentation that need to be updated with `user-guide`.
```suggestion
```

webapp/src/types/api.ts
```diff
@@ -1,6 +1,7 @@
 import { components } from "types/generated/generatedTypes";
 
 export type AzimuthConfig = components["schemas"]["AzimuthConfig"];
+export type PipelineDefinition = components["schemas"]["PipelineDefinition"];
```
JosephMarinier, line 4: All the types in this file are ordered alphabetically. Can you move this one to conform to that?

webapp/src/pages/ClassOverlap.tsx
nandhinibsn: Yeah right! I noticed now it'd not been called anywhere. The URL needs to be updated to "link="user-guide/class-overlap/."

webapp/src/types/api.ts
nandhinibsn: Okay, I noted and updated the order.

webapp/src/pages/Settings.tsx
nandhinibsn: Oh yes!. I missed one here. Added the fontsize

webapp/src/pages/Settings.tsx
nandhinibsn: Right!. It had to be removed but seems like I missed one of these on code clean up.

webapp/src/pages/Settings.tsx
nandhinibsn: Okay!. I gave this to remove extra padding around the typography.

JosephMarinier: Can you explain to me how this is supposed to be related to padding? I don't see it in the [list](https://mui.com/system/spacing/#api) of valid properties.

nandhinibsn: Sure, I can explain why I have added s in sx; when I added m:1 in sx, a margin/padding was added around the content. So, I changed to "s:1" to remove those margins around the content.

JosephMarinier: The `m` stands for `margin`. What does `s` stand for? What does it do?
I think it doesn't do anything, and it should be removed.

nandhinibsn: Okay!. I assumed it set the spacing as 1, But I am wrong in understanding it. I removed the changes in the latest commit.

nandhinibsn:
> Can you log the change in `CHANGELOG.md` and check the checklist in the PR description?

Can you please help me with the issue/resolve id for this feature?

JosephMarinier:
> > Can you log the change in `CHANGELOG.md` and check the checklist in the PR description?
> 
> Can you please help me with adding the issue/resolve id for this feature?

There was no issue created for this feature.

> If a box does not apply, check it to acknowledge it.

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,694 @@
...
+        >
+          {getModelContractConfigSection()}
+        </AccordionLayout>
+        <AccordionLayout
+          name="Analyses Customization"
+          description="four analyses configured in Azimuth"
+          link="reference/configuration/analyses/"
+        >
+          {getAnalysesCustomization()}
+        </AccordionLayout>
```
JosephMarinier, line 647: @gabegma just a reminder about [this](https://github.com/ServiceNow/azimuth/pull/282#issuecomment-1344690308). Would you like those section descriptions?

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,694 @@
...
+        <AccordionLayout
+          name="Model Contract Configurations"
+          description="defines how Azimuth interacts with the ML pipelines and the metrics"
+          link="reference/configuration/model_contract/"
+        >
+          {getModelContractConfigSection()}
+        </AccordionLayout>
+        <AccordionLayout
+          name="Analyses Customization"
+          description="four analyses configured in Azimuth"
```
gabegma, line 643:
```suggestion
          name="Analyses Customization"
          description="Enable or disable some analyses and modify corresponding thresholds."
```

webapp/src/components/AccordionLayout.tsx
JosephMarinier: I resolved this in 8c9d99b1480c423a72af22923af90bbbb4683918

webapp/src/components/AccordionLayout.tsx
JosephMarinier: I resolved this in 8c9d99b1480c423a72af22923af90bbbb4683918

webapp/src/pages/Settings.tsx
JosephMarinier: I resolved this in 8c9d99b1480c423a72af22923af90bbbb4683918

webapp/src/pages/ClassOverlap.tsx
JosephMarinier: I resolved this in 8c9d99b1480c423a72af22923af90bbbb4683918

webapp/src/pages/Settings.tsx
JosephMarinier: I resolved this in 8c9d99b1480c423a72af22923af90bbbb4683918

webapp/src/types/api.ts
JosephMarinier: I resolved this in 8c9d99b1480c423a72af22923af90bbbb4683918

webapp/src/pages/Settings.tsx
JosephMarinier: Perfect! I committed your suggestions in c00e45ee791bcab2dc0fc0d33275f4bcea86c864. Thank you for your help!

webapp/src/pages/Settings.tsx
```diff
@@ -1,78 +1,692 @@
...
+  high_epistemic_threshold: { min: 0, max: 1, step: 0.1 },
+  conflicting_neighbors_threshold: { min: 0, max: 1, step: 0.1 },
+  no_close_threshold: { min: 0, max: 1, step: 0.1 },
+  min_num_per_class: { min: 0, max: 100, step: 10 },
+  max_delta_class_imbalance: { min: 0, max: 1, step: 0.1 },
+  max_delta_representation: { min: 0, max: 0.1, step: 0.01 },
+  max_delta_mean_tokens: { min: 0.0, max: 10.0, step: 1.0 },
+  max_delta_std_tokens: { min: 0.0, max: 10.0, step: 1.0 },
+  short_sentence_max_token: { min: 0, max: 100, step: 1 },
+  long_sentence_min_token: { min: 0, max: 100, step: 1 },
```
JosephMarinier, line 34: I would
- set `min_num_per_class`'s `step` to `1` instead of `10`;
- set `max_delta_representation`'s `max` to `1` as it's a percentage and `max: 0.1` seems arbitrary;
- remove the `max`es that are not `1`, as they seem arbitrary;
- set a decimal `step` to all `float` fields, including `max_delta_mean_tokens` and `max_delta_std_tokens` to `0.1`
- remove `step: 1` as it is the default; and
- remove the unnecessary `.0`s;
```suggestion
  iterations: { min: 0 },
  high_epistemic_threshold: { min: 0, max: 1, step: 0.1 },
  conflicting_neighbors_threshold: { min: 0, max: 1, step: 0.1 },
  no_close_threshold: { min: 0, max: 1, step: 0.1 },
  min_num_per_class: { min: 0 },
  max_delta_class_imbalance: { min: 0, max: 1, step: 0.1 },
  max_delta_representation: { min: 0, max: 1, step: 0.01 },
  max_delta_mean_tokens: { min: 0, step: 0.1 },
  max_delta_std_tokens: { min: 0, step: 0.1 },
  short_sentence_max_token: { min: 0 },
  long_sentence_min_token: { min: 0 },
```
@gabegma, any comments on the values? Like steps of `0.01` vs `0.1` for example.

gabegma: All your suggestions LGTM!
The steps are for the arrows beside the box where the user can type, right? So the user could still type `0.15` and have a value in between the steps increment?

webapp/src/pages/Settings.tsx
JosephMarinier: I removed all the `s: 1`s in 662f5121bf3ff22981351caacadc919fd52aa6bc.

webapp/src/pages/Settings.tsx
JosephMarinier: Thanks! That is correct.

JosephMarinier (approved):

PR #284, opened 2022-11-01
Add class analysis

JosephMarinier:
## Description:

New class analysis section, with class overlap detection. More details are available in the User Guide and in the Key Concepts.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


webapp/src/components/ClassAnalysisTable.tsx
```diff
@@ -0,0 +1,140 @@
...
+        {
+          flex: 1,
+          field: "targetClass",
+          headerName: "Target Class",
+        },
+        {
+          flex: 1,
+          type: "number",
+          field: "overlapScoreTrain",
+          headerName: "Overlap Score",
```
lindsaydbrin, line 84: This is the table header, right? How do you feel about changing this to "Semantic Overlap Score?" I think "class overlap" is a little confusing in a non-specific, vague way - like, it could be interpreted as something related to model confusion - and "semantic overlap" makes it more clear what we're referring to. I think the title will fit fine on 2 lines, but if it's too long, I think I'd lean towards "Semantic Overlap" over "Overlap Score" (i.e., I think adding _semantic_ is more important than keeping _score_).

```suggestion
          headerName: "Semantic Overlap Score",
```

lindsaydbrin: Never mind - we can do this in another PR focusing on improving class analysis docs and in-app communication!

docs/docs/key-concepts/class-similarity.md
```diff
@@ -0,0 +1,74 @@
+# Class Similarity
+
+## What is it?
+
+Class analysis is an aspect of similarity analysis. As such, it is based on the relative locations
+of utterances in embedding space, and only requires training data to be supplied.
```
lindsaydbrin, line 6:
```suggestion
Class analysis is an aspect of similarity analysis, so it is based on the relative locations of utterances in embedding space. Because of this, it only requires training data to be supplied.
```

lindsaydbrin: We need to sort out class similarity, class analysis, and class overlap. In the meantime, I made these sentences technically correct. (I.e., the class analysis table also includes model confusion, so it _includes_ an aspect of similarity analysis, but it is not _just_ an aspect of similarity analysis.)

webapp/src/components/ClassAnalysisTable.tsx
```diff
@@ -0,0 +1,140 @@
...
+          description: "Normalized overlap score (training set)",
```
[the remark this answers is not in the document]
lindsaydbrin: Same as above - we can address this in a separate PR.

azimuth/modules/dataset_analysis/class_overlap.py
```diff
@@ -0,0 +1,84 @@
+from dataclasses import asdict
```
lindsaydbrin, line 1:
```suggestion
# Copyright ServiceNow, Inc. 2021 – 2022
# This source code is licensed under the Apache 2.0 license found in the LICENSE file
# in the root directory of this source tree.
from dataclasses import asdict
```
Looking at other files in the same folder, I think we need to add this, and maybe in other key (like, logic-heavy) class overlap files too?

lindsaydbrin (approved):
Looks good! Thank you - I'm so excited about this!

PR #285, opened 2022-11-01
Nandhini/performance analysis table additional change

nandhinibsn:
Resolve #

## Description:
1. Added "desc" icon to Utterance Count to enable desc by default.

2. Added the tooltip for custom metric columns and ECE.

3. When there is only one pipeline configured...

The "Compare Pipelines" button on the dashboard is instead called "View details" and is still linking to the standalone page.
The "Compare Baseline with" select menu is hidden.

The "Compare Baseline with" select menu has a default value of the first available pipeline.
## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
+++ b/webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
@@ -191,6 +191,14 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
     selectedMetricPerFilterOption,
   ]);
 
+  const fieldWithTooltip = (tooltip: string, label: string) => (
+    <Tooltip title={tooltip}>
+      <Typography variant="caption" fontSize={14}>
+        {label}
+      </Typography>
+    </Tooltip>
+  );
+
   const columns = React.useMemo((): Column<Row>[] => {
     const metricsEntries = Object.entries(metricInfo ?? {});
 
@@ -350,14 +358,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
           ...pipelines.map<Column<Row>>((pipeline) => ({
             ...METRIC_COLUMN,
             field: `${pipeline}${metricName}`,
-            ...groupHeader(
-              pipeline,
-              <Tooltip title={description}>
-                <Typography variant="caption" fontSize={14}>
-                  {metricName}
-                </Typography>
-              </Tooltip>
-            ),
+            ...groupHeader(pipeline, fieldWithTooltip(description, metricName)),
             valueGetter: ({ row: { [pipeline]: metrics } }) =>
               metrics ? metrics.customMetrics[metricName] ?? NaN : undefined,
             renderCell: ({ value }: GridCellParams<number | undefined>) =>
@@ -393,14 +394,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
       ...pipelines.map<Column<Row>>((pipeline) => ({
         ...METRIC_COLUMN,
         field: `${pipeline}ECE`,
-        ...groupHeader(
-          pipeline,
-          <Tooltip title={ECE_TOOLTIP}>
-            <Typography variant="caption" fontSize={14}>
-              ECE
-            </Typography>
-          </Tooltip>
-        ),
+        ...groupHeader(pipeline, fieldWithTooltip(ECE_TOOLTIP, "ECE")),
         valueGetter: ({ row }) => row[pipeline]?.ece,
         renderCell: ({ value }: GridCellParams<number | undefined>) =>
           value !== undefined && (
@@ -437,8 +431,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
   ]);
 
   React.useEffect(() => {
-    config &&
-      config.pipelines &&
+    config?.pipelines &&
       config.pipelines.length > 1 &&
       config.pipelines.map(
         (_, index) =>
--- a/webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
+++ b/webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
@@ -109,11 +109,15 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
   const [selectedMetricPerFilterOption, setSelectedMetricPerFilterOption] =
     React.useState<FilterByViewOption>("label");
 
+  const { data: config } = getConfigEndpoint.useQuery({ jobId });
+
   const [comparedPipeline, setComparedPipeline] = React.useState<
     number | undefined
-  >();
-
-  const { data: config } = getConfigEndpoint.useQuery({ jobId });
+  >(
+    config?.pipelines && config.pipelines.length > 1
+      ? (pipeline.pipelineIndex + 1) % config.pipelines.length
+      : undefined
+  );
 
   const { data: metricInfo } = getCustomMetricInfoEndpoint.useQuery({
     jobId,
@@ -191,11 +195,9 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
     selectedMetricPerFilterOption,
   ]);
 
-  const fieldWithTooltip = (tooltip: string, label: string) => (
+  const reactNodeWithTooltip = (tooltip: string, nodeText: string) => (
     <Tooltip title={tooltip}>
-      <Typography variant="caption" fontSize={14}>
-        {label}
-      </Typography>
+      <Typography variant="inherit">{nodeText}</Typography>
     </Tooltip>
   );
 
@@ -262,7 +264,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
                         display="flex"
                         alignItems="center"
                         justifyContent="center"
-                        zIndex="tooltip"
+                        zIndex={1}
                         gap={1}
                       >
                         {longHeader}
@@ -358,7 +360,10 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
           ...pipelines.map<Column<Row>>((pipeline) => ({
             ...METRIC_COLUMN,
             field: `${pipeline}${metricName}`,
-            ...groupHeader(pipeline, fieldWithTooltip(description, metricName)),
+            ...groupHeader(
+              pipeline,
+              reactNodeWithTooltip(description, metricName)
+            ),
             valueGetter: ({ row: { [pipeline]: metrics } }) =>
               metrics ? metrics.customMetrics[metricName] ?? NaN : undefined,
             renderCell: ({ value }: GridCellParams<number | undefined>) =>
@@ -394,7 +399,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
       ...pipelines.map<Column<Row>>((pipeline) => ({
         ...METRIC_COLUMN,
         field: `${pipeline}ECE`,
-        ...groupHeader(pipeline, fieldWithTooltip(ECE_TOOLTIP, "ECE")),
+        ...groupHeader(pipeline, reactNodeWithTooltip(ECE_TOOLTIP, "ECE")),
         valueGetter: ({ row }) => row[pipeline]?.ece,
         renderCell: ({ value }: GridCellParams<number | undefined>) =>
           value !== undefined && (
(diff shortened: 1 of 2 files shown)


nandhinibsn:
If more than 1 pipeline available:
image
image

If there is only 1 pipeline exists:
image
image

nandhinibsn:
Hey @JosephMarinier, there is one thing I noticed while adding "desc" enabled for the utterance count column on both the base pipeline and compare pipeline comparison scenario, currently, we can only enable 1 column sort option so, to add desc by default for basePipelineUtteranceCount and comparedPipelineUtteranceCount, I need to add some custom logic. But need some thoughts from the team before proceeding with it.

nandhinibsn:
What could be the issue number for this PR here? Please suggest here @JosephMarinier

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -423,6 +431,12 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
   ]);
 
   React.useEffect(() => {
+    config?.pipelines &&
+      config.pipelines.length > 1 &&
+      config.pipelines.map(
+        (_, index) =>
+          index !== pipeline.pipelineIndex && setComparedPipeline(index)
+      );
```
JosephMarinier, line 439: This doesn't work. It prevents selecting any pipeline but the last one in the list. It is not even possible to select `No pipeline`. That is because this `React.useEffect` hook runs every time one of `[pipeline, selectedMetricPerFilterOption, comparedPipeline, metricInfo]` changes. Also, if there are more than two pipelines configured, the `map()` calls `setComparedPipeline()` on every item in the list. Instead, we should call `setComparedPipeline()` only once.

```suggestion
```

Instead, I think it would be reasonable to use `config?.pipelines` at the time of initializing `comparedPipeline`. In practice, `config` will always be defined on the first call (never loading) since it is used in `PipelineCheck` and will prevent the whole page to render if it is not loaded. And if ever that changes on day, it's not too bad to get `undefined` (no pipeline) by default. Then, instead of the `map()`, I see a couple of possibilities to choose the default pipeline:
- we default to the pipeline `0`, or `1` if the base pipeline is `0`; or
- we default to the next pipeline (after the base pipeline), or `0` is the base pipeline is the last one.

This suggestion uses the second approach:
```diff
-  const [comparedPipeline, setComparedPipeline] = React.useState<
-    number | undefined
-  >();
-
   const { data: config } = getConfigEndpoint.useQuery({ jobId });

+  const [comparedPipeline, setComparedPipeline] = React.useState(
+    config?.pipelines && config.pipelines.length > 1
+      ? (pipeline.pipelineIndex + 1) % config.pipelines.length
+      : undefined
+  );
+
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -190,6 +191,14 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
     selectedMetricPerFilterOption,
   ]);
 
+  const fieldWithTooltip = (tooltip: string, label: string) => (
+    <Tooltip title={tooltip}>
+      <Typography variant="caption" fontSize={14}>
```
JosephMarinier, line 196: This is not a caption. Instead, just inherit the parent's style:
```suggestion
      <Typography variant="inherit">
```

JosephMarinier (commented):
Good job! There are a couple of things to polish, but we are getting there. 👍

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -190,6 +191,14 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
     selectedMetricPerFilterOption,
   ]);
 
+  const fieldWithTooltip = (tooltip: string, label: string) => (
```
JosephMarinier, line 194: Can you find better names for the `fieldWithTooltip` function name and the `label` argument name? This is neither a "field" nor a "label". We can also go more general with something like `const typographyWithTooltip = (tooltip: string, typography: string) => (`

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -147,7 +148,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
 
   // Track table sort model to keep 'overall' at top.
   const [sortModel, setSortModel] = React.useState<GridSortModel>([
-    { field: "utteranceCount", sort: "desc" },
+    { field: "basePipelineUtteranceCount", sort: "desc" },
```
JosephMarinier, line 151: That's perfect. I don't think you need

> to add desc by default for basePipelineUtteranceCount and comparedPipelineUtteranceCount

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -253,6 +262,7 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
                         display="flex"
                         alignItems="center"
                         justifyContent="center"
+                        zIndex="tooltip"
```
JosephMarinier, line 265: That is not a tooltip. I'm afraid putting it at the `zIndex` of tooltips might cause conflicts. A simple `1` seems to work. Makes sense?
```suggestion
                        zIndex={1}
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: So, don't we need desc by default for utteranceCount? May I know why we ignore the comparison table?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: My initial choice was to use zIndex={1} but chose to add zIndex="tooltip" because this is specifically for a tooltip.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Okay, to be more generic, I agree to go for a change in the function name. I suggest "reactNodeWithTooltip" as the function name and nodeText for the label arg. The reason I chose this is that we pass this as reactNode. let me know what do you think?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: okay, yes inherit would do the same as we..I can change the variant as suggested. thanks

webapp/src/pages/Dashboard.tsx
```diff
@@ -89,7 +91,12 @@ const Dashboard = () => {
         <PreviewCard
           title="Pipeline Metrics by Data Subpopulation"
           to={`/${jobId}/pipeline_metrics${searchString}`}
-          linkButtonText="Compare pipelines"
+          linkButtonText={
+            (config?.pipelines &&
+              config?.pipelines?.length > 1 &&
+              "Compare pipelines") ||
+            undefined
```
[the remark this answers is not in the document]
nandhinibsn: I prefer both ways are good. But, May I know why the second one is better to be cleaner?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Yes, Agreed. Thanks for this input!

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: No, a [tooltip](https://en.wikipedia.org/wiki/Tooltip) only appears
> when hovering over a screen element or component

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: We can only (easily) sort by one column, and that's all we really need. By sorting on the base pipeline's utterance count by default, we preserve the exact order in which the performance analysis table was on the Dashboard when navigating to pipeline comparison. I think the continuity is the most (and only) important aspect.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -348,8 +360,10 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
           ...pipelines.map<Column<Row>>((pipeline) => ({
             ...METRIC_COLUMN,
             field: `${pipeline}${metricName}`,
-            description,
-            ...groupHeader(pipeline, metricName),
+            ...groupHeader(
+              pipeline,
+              reactNodeWithTooltip(description, metricName)
```
JosephMarinier, line 365: Continuing on https://github.com/ServiceNow/azimuth/pull/285#discussion_r1023030193
I don't like the `reactNode` thing. This function takes a `string` in argument and returns a `Typography` that has a tooltip. It is much more specific than dealing with `ReactNode`s.
Anyway, here is a different solution: I think the fact that it is so hard to find a name hints at the fact we don't need it. I would suggest:
```suggestion
              <Tooltip title={description}>
                <Typography variant="inherit">{metricName}</Typography>
              </Tooltip>
```

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
```diff
@@ -108,11 +109,15 @@ const PerformanceAnalysisTable: React.FC<Props> = ({
   const [selectedMetricPerFilterOption, setSelectedMetricPerFilterOption] =
     React.useState<FilterByViewOption>("label");
 
+  const { data: config } = getConfigEndpoint.useQuery({ jobId });
+
   const [comparedPipeline, setComparedPipeline] = React.useState<
     number | undefined
```
JosephMarinier, line 115: Detail, but as seen in my suggestion [here](https://github.com/ServiceNow/azimuth/pull/285#discussion_r1010987860), this type is redundant, so we can simplify:
```diff
-  const [comparedPipeline, setComparedPipeline] = React.useState<
-    number | undefined
-  >(
+  const [comparedPipeline, setComparedPipeline] = React.useState(
```

webapp/src/pages/Dashboard.tsx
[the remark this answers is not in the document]
JosephMarinier: Because this is exactly what the [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) (`A ? B : C`) is for. Might as well use it. While `(A && B) || C` will have an equivalent behavior in **this situation**, it would not produce the expected result if `B` was to be a falsy value, like `""` or `0`. So, the ternary operator makes it extra clear that the intended output is
`<if there is more than one pipeline> ? <return something> : <return something else>`. Plus it takes one fewer line of code. 😎 Does that make sense?

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
nandhinibsn: Right, but the same lines of code, "Typography with the tooltip," is used in 2 different places in the code at - lines no# 365, 402. So, I would suggest we should keep this function and can decide on a proper name here. 

As you mentioned "reactNode" is not more appropriate here, we can use the names that you have suggested here "const typographyWithTooltip = (tooltip: string, typography: string) => (" earlier.

webapp/src/pages/Dashboard.tsx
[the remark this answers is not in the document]
JosephMarinier: Same thing will happen with the ternary operator (`A ? B : C`). If `config?.pipelines` is falsy, the interpreter will stop evaluation the condition and will directly return `undefined`.

webapp/src/components/Metrics/PerformanceAnalysisTable.tsx
JosephMarinier: I know. I just didn't think repeating 2 times 3 lines of trivial code can be deemed "dangerous" duplication. I would not consider it worth avoiding by creating a 5-line function. But it doesn't hurt either. Let's go with what you have right now. 👍

JosephMarinier (approved):
Here we go! Thank you for taking care of this. 👍

Issue #297, opened 2022-11-21
Ability to import back proposed actions


gabegma:
This would be useful for our users, but also for our demos.

PR #301, opened 2022-11-24
Add support for persistent ID and filtering on ID in exploration space

### LINKED ISSUE #200: The search bar for filtering utterances should work with the id as well
With big datasets, looking for a specific example ID takes a long time. Ideally, typing it in the search bar would directly find it.

gabegma:
Supports #297 and fixes #200

## Description:
You can look at the commits' message for the details of what I have done.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


tests/conftest.py
```diff
@@ -179,11 +179,14 @@ def file_text_config_no_intent(file_text_config_top1):
 
 
 @pytest.fixture
-def clinc_text_config(simple_text_config):
-    clinc_text_config = simple_text_config.copy(deep=True)
-    clinc_text_config.dataset = DATASET_CLINC150_CFG
-    clinc_text_config.rejection_class = "NO_INTENT"
-    clinc_text_config.name = "clinc-test"
+def clinc_text_config(tmp_path):
```
gabegma, line 182: I needed to change this code because the config's dataset could not be loaded before. `DATASET_CLINC150_CFG` was a `Dict` and not a `CustomObject`.

azimuth/dataset_split_manager.py
```diff
@@ -222,14 +236,17 @@ def save_csv(self, table_key=None) -> str:
...
+        ]
 
         # The following allows for new or extra columns to end up here automatically,
         # instead of being lost if we were to hardcode the whole list.
-        omit = {*order, FEATURES, FEATURE_FAISS}
+        omit = {*order_unique, FEATURES, FEATURE_FAISS}
         rest = [c for c in self.get_dataset_split(table_key).column_names if c not in omit]
 
-        columns = order + rest
+        columns = order_unique + rest
```
JosephMarinier, line 249: This `rest` thing is a (second) convoluted mechanism for eliminating duplicates. I think a single `dict.fromkeys()` could take care of both. Great find by the way! So, no need to make `order` unique, keep it as before, then you can get rid of `omit` and `rest` (to verify, but `FEATURES` and `FEATURE_FAISS` don't seem important anymore) and do something like this:
```suggestion
        # dict.fromkeys() eliminates duplicates.
        columns = list(dict.fromkeys([*order, *self.get_dataset_split(table_key).column_names]))
```
While we're here, I would suggest creating a variable `available_columns = self.get_dataset_split(table_key).column_names` to avoid calling `get_dataset_split()` `len(order)` times. All in all, it would end up something like this:
```diff
@@ -236,17 +236,13 @@ class DatasetSplitManager:
             DatasetColumn.neighbors_eval,
             *self._tags,
         ]
-        order_unique = list(dict.fromkeys(order))  # persistent_id can be called row_idx.
-        order_unique = [
-            c for c in order_unique if c in self.get_dataset_split(table_key).column_names
-        ]
+        available_columns = self.get_dataset_split(table_key).column_names
+        order = [c for c in order if c in available_columns]

-        # The following allows for new or extra columns to end up here automatically,
+        # This *available_columns allows for new or extra columns to end up here automatically,
         # instead of being lost if we were to hardcode the whole list.
-        omit = {*order_unique, FEATURES, FEATURE_FAISS}
-        rest = [c for c in self.get_dataset_split(table_key).column_names if c not in omit]
-
-        columns = order_unique + rest
+        # The dict.fromkeys() avoids duplicates.
+        columns = list(dict.fromkeys([*order, *available_columns]))

         # pd.to_csv() instead of HF version to avoid unintended type conversions (list to array)
         df = pd.DataFrame(self.get_dataset_split_with_class_names(table_key)).reindex(
```

azimuth/dataset_split_manager.py
```diff
@@ -153,6 +152,21 @@ def _load_dataset_split(self) -> Optional[Tuple[Dataset, Dataset]]:
...
         return None
 
+    def _load_base_dataset_split(self, dataset_split) -> Tuple[Dataset, Dataset]:
+        base_dataset_split, malformed_dataset = self._split_malformed(dataset_split)
+
+        # Checking if a persistent id was provided.
+        persistent_id = self.config.columns.persistent_id
+        if persistent_id != DatasetColumn.row_idx:  # Default value
+            if persistent_id not in base_dataset_split.column_names:
+                raise ValueError(f"Persistent ID named {persistent_id} not found in the dataset.")
```
JosephMarinier, line 162: I know not everyone is super rigorous on this, but I like to make the distinction:
- lowercase "id" is an abbreviation for "identifier". That's what we have here.
- uppercase "ID" is an acronym, so I reserve it for "Identity Document". That is a document like a passport or a driver's license.

Does that make sense to you?
```suggestion
                raise ValueError(f"Persistent id named {persistent_id} not found in the dataset.")
```

azimuth/dataset_split_manager.py
```diff
@@ -153,6 +152,21 @@ def _load_dataset_split(self) -> Optional[Tuple[Dataset, Dataset]]:
...
+            else:
+                all_persistent_ids = base_dataset_split[persistent_id]
+                if len(all_persistent_ids) > len(set(all_persistent_ids)):
+                    raise ValueError(f"Persistent IDs in {persistent_id} column need to be unique.")
```
JosephMarinier, line 166: No need for the `else`, if you wish to save an indent.
```suggestion

            all_persistent_ids = base_dataset_split[persistent_id]
            if len(all_persistent_ids) > len(set(all_persistent_ids)):
                raise ValueError(f"Persistent ids in {persistent_id} column need to be unique.")
```

tests/conftest.py
JosephMarinier: I'm sorry. You took the time to write a comment, but I don't understand it.
Did `DATASET_CLINC150_CFG` change?

azimuth/dataset_split_manager.py
gabegma: Totally! Thank you!

tests/conftest.py
JosephMarinier: Right, I see. 👍 
But for my personal knowledge: What's new? Why do we need to fix it now? I guess something in the new test, like `load_dataset_from_config()` or `DatasetSplitManager()` needed the real thing, and we were not doing that in any test before. That makes sense.

azimuth/utils/dataset_operations.py
```diff
@@ -62,9 +62,13 @@ def filter_dataset_split(
         and config.columns.text_input in dataset_split.column_names
         and dataset_split.num_rows != 0
     ):
-        by = clean_utterance(filters.utterance)
+        by = filters.utterance
+        by_clean = clean_utterance(by)
```
JosephMarinier, line 66: We can get rid of it. And `by_clean` could simply be
```suggestion
        clean = clean_utterance(by)
```

azimuth/utils/dataset_operations.py
```diff
@@ -62,9 +62,13 @@ def filter_dataset_split(
...
+        # Filter in utterances or if string matches a known row_idx or persistent_id
         dataset_split = dataset_split.filter(
-            lambda x: by in clean_utterance(x[config.columns.text_input])
+            lambda x: by_clean in clean_utterance(x[config.columns.text_input])
+            or by == str(x[DatasetColumn.row_idx])
+            or by == str(x[config.columns.persistent_id])
```
JosephMarinier, line 71:
```suggestion
            lambda x: clean in clean_utterance(x[config.columns.text_input])
            or filters.utterance == str(x[DatasetColumn.row_idx])
            or filters.utterance == str(x[config.columns.persistent_id])
```

JosephMarinier (commented):
You mentioned `by` was not a great variable name... I agree. It's just a detail, but here is a simple change:

JosephMarinier (commented):
You mentioned `by` was not a great variable name... I agree! It's just a detail, but here is a simple change:

JosephMarinier (approved):
That's pretty cool! Thank you!
It should then be very easy to add it to the response of `/utterances`... In another PR?

gabegma:
> It should then be very easy to add it to the response of /utterances... In another PR?

Right @JosephMarinier, maybe it can be included in the PR where you show it on hover in the UI? That way you can name the variable in the API how you prefer. I can also include it in this PR, let me know.

JosephMarinier:
Perfect plan! 👌

PR #302, opened 2022-11-25
Misc fixes and cleanups

JosephMarinier:
## Description:

Here is a compilation of misc small fixes and trivial cleanups. I recommend looking (quickly) at each commit independently.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Analysis/WarningsPreview.tsx
+++ b/webapp/src/components/Analysis/WarningsPreview.tsx
@@ -3,7 +3,6 @@ import Loading from "components/Loading";
 import { WarningPlot } from "components/PlotWrapper";
 import React from "react";
 import { getDatasetWarningsEndpoint } from "services/api";
-import { DatasetWarning } from "types/api";
 import PreviewToggle from "./PreviewToggle";
 import PreviewToggleButton from "./PreviewToggleButton";
 
@@ -15,10 +14,10 @@ const WarningsPreview: React.FC<{ jobId: string }> = ({ jobId }) => {
     isFetching,
   } = getDatasetWarningsEndpoint.useQuery({ jobId });
 
-  const warnings = React.useMemo<(DatasetWarning | undefined)[]>(
+  const warnings = React.useMemo(
     () =>
       datasetWarningGroups?.flatMap(({ warnings }) => warnings) ||
-      Array.from(Array(3), () => undefined),
+      Array.from(Array(4), () => undefined),
     [datasetWarningGroups]
   );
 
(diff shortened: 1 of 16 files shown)


webapp/src/components/Analysis/WarningsPreview.tsx
```diff
@@ -15,10 +14,10 @@ const WarningsPreview: React.FC<{ jobId: string }> = ({ jobId }) => {
     isFetching,
   } = getDatasetWarningsEndpoint.useQuery({ jobId });
 
-  const warnings = React.useMemo<(DatasetWarning | undefined)[]>(
```
nandhinibsn, line 18: Any reason for getting rid of the typing here?

JosephMarinier: No particular reason. Simply because the explicit type had no effect, the automatic type inference is yielding the correct type. In fact, our explicit type `(DatasetWarning | undefined)[]` was not as precise as the inferred type `DatasetWarning[] | undefined[]`, which correctly reflects that all items of the array are of the same type.

Removing it means less imports, less refactor if we rename or change the type, etc.

webapp/src/components/Controls/FilterTextField.tsx
```diff
@@ -53,24 +53,27 @@ const FilterTextField: React.FC<Props> = ({
...
-          value={liveValue ?? ""}
-          onChange={(event) => handleChange(event.target.value || undefined)}
-        />
-      </Box>
+      <Input
+        startAdornment={
+          <InputAdornment position="start">
+            <Search />
+          </InputAdornment>
+        }
```
JosephMarinier, line 61: Precision on
> [Clean up unnecessary `div`](https://github.com/ServiceNow/azimuth/pull/302/commits/7dda02e73b93b1734bcf287d4c5b0e2b05bbf0b0) by using a `startAdornment` for the `Search` icon.

This automatically gives the magnifying glass the right color and spacing, no need to customize. The only noticeable difference is that the underline now goes all the way under the magnifying glass. I like the symmetry of it.

I changed one more thing: I removed 8px of margin that was added between the label "Utterance" and the field. I think it looks perfect by default.

  before this PR   | using `startAdornment` | reduced margin
-- | -- | --
Screen Shot 2022-11-24 at 5 47 34 PM|Screen Shot 2022-11-24 at 5 48 22 PM|Screen Shot 2022-11-24 at 6 11 30 PM

christyler3030 (approved):

PR #332, opened 2022-12-08
Add new start-up tasks

gabegma:
## Description:
On big datasets, the behavioral testing summary and class overlap modules would error out in the UI, while they were being computed. I added them both to the start-up, and that forced me to refactor so we could define what dataset splits to send to a particular start-up task.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/startup.py
+++ b/azimuth/startup.py
@@ -58,7 +58,6 @@ SIMILARITY_TASKS = [
     ),
 ]
 
-
 BMA_PREDICTION_TASKS = [
     Startup(
         "prediction_bma",
--- a/azimuth/startup.py
+++ b/azimuth/startup.py
@@ -41,12 +41,22 @@ class Startup:
     # The names of other start-up tasks to wait for.
     dependency_names: List[str] = field(default_factory=list)
     run_on_all_pipelines: bool = False
+    dataset_split_names: List[DatasetSplitName] = field(
+        default_factory=lambda: [DatasetSplitName.train, DatasetSplitName.eval]
+    )
 
 
 START_UP_THREAD_NAME = "Azimuth_Startup"
 
-SIMILARITY_TASKS = [
-    Startup("neighbors_tags", SupportedModule.NeighborsTagging),
+SIMILARITY_TASKS = [Startup("neighbors_tags", SupportedModule.NeighborsTagging)]
+
+SIMILARITY_TRAIN_TASKS = [
+    Startup(
+        "class_overlap",
+        SupportedModule.ClassOverlap,
+        dependency_names=["prediction", "neighbors_tags"],
+        dataset_split_names=[DatasetSplitName.train],
+    )
 ]
 
 BMA_PREDICTION_TASKS = [
@@ -59,12 +69,37 @@ BMA_PREDICTION_TASKS = [
 ]
 
 PERTURBATION_TESTING_TASKS = [
-    Startup("perturbation_testing", SupportedModule.PerturbationTesting, run_on_all_pipelines=True)
+    Startup(
+        "perturbation_testing",
+        SupportedModule.PerturbationTesting,
+        dependency_names=["prediction"],
+        run_on_all_pipelines=True,
+    ),
+    Startup(
+        "perturbation_testing_summary",
+        SupportedModule.PerturbationTestingSummary,
+        dependency_names=["perturbation_testing"],
+        run_on_all_pipelines=True,
+        dataset_split_names=[DatasetSplitName.all],
+    ),
 ]
 
 PIPELINE_COMPARISON_TASKS = [
     Startup(
-        "prediction_comparison", SupportedModule.PredictionComparison, run_on_all_pipelines=False
+        "prediction_comparison",
+        SupportedModule.PredictionComparison,
+        dependency_names=["prediction"],
+        run_on_all_pipelines=False,
+    )
+]
+
+POSTPROCESSING_EVAL_TASKS = [
+    Startup(
+        "outcome_count_per_threshold",
+        SupportedModule.OutcomeCountPerThreshold,
+        dependency_names=["prediction", "outcome_count"],
+        run_on_all_pipelines=True,
+        dataset_split_names=[DatasetSplitName.eval],
     )
 ]
 
@@ -126,6 +161,7 @@ def make_startup_tasks(
     mod_options: Dict,
     dependencies: List[DaskModule],
     pipeline_index: Optional[int],
+    dataset_split_names: List[DatasetSplitName],
 ) -> Dict[DatasetSplitName, DaskModule]:
     """
     Apply `module` on all indices of all dataset split names and call `save_results` at the end.
@@ -137,6 +173,7 @@ def make_startup_tasks(
         mod_options: Special kwargs for the Module.
         dependencies: Which modules to include as dependency.
         pipeline_index: On which pipeline to run the startup task.
+        dataset_split_names: List of DatasetSplitName on which to run the task.
 
     Returns:
         Named Modules with the tasks.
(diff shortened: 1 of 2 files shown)


gabegma:
I'm waiting on the changes in another incoming PR about including datasets in the hash before removing this from draft.

azimuth/app.py
```diff
@@ -240,12 +240,15 @@ def initialize_managers(azimuth_config, cluster):
...
+        current_config = _task_manager.config
     else:
         task_history = {}
+        current_config = None
 
     _task_manager = TaskManager(azimuth_config, cluster=cluster)
 
-    _task_manager.current_tasks = task_history
+    if _task_manager.config == current_config:
+        _task_manager.current_tasks = task_history
```
gabegma, line 251: When rebasing on #350, these changes should not be needed anymore.

azimuth/startup.py
```diff
@@ -59,12 +69,37 @@ class Startup:
 ]
 
 PERTURBATION_TESTING_TASKS = [
-    Startup("perturbation_testing", SupportedModule.PerturbationTesting, run_on_all_pipelines=True)
+    Startup(
+        "perturbation_testing",
+        SupportedModule.PerturbationTesting,
+        dependency_names=["prediction"],
```
JosephMarinier, line 75: I see you added a bunch of dependencies that make a lot of sense, but I'm curious; why they weren't there already?

azimuth/startup.py
```diff
@@ -146,16 +183,22 @@ def make_startup_tasks(
     dms = {k: v for k, v in dataset_split_managers.items() if v is not None}
     tasks: Dict[DatasetSplitName, DaskModule] = {}
 
-    for dataset_split_name, dm in dms.items():
+    if DatasetSplitName.all not in dataset_split_names:
+        available_datasets_splits = list(set(dataset_split_names) & set(dms))
+    else:
+        available_datasets_splits = [DatasetSplitName.all]
+
+    for dataset_split_name in available_datasets_splits:
```
JosephMarinier, line 191: Hfff, this dms logic is becoming messy, don't you agree? Here is an idea
```suggestion
    for dataset_split_name, dm in {**dms, DatasetSplitName.all: dms}.items():
        if dataset_split_name not in dataset_split_names:
            continue
```
```diff
-            dm = dms[dataset_split_name] if dataset_split_name != DatasetSplitName.all else dms
```

azimuth/startup.py
```diff
@@ -146,16 +183,22 @@ def make_startup_tasks(
...
             dataset_split_name=dataset_split_name,
             dependencies=dependencies,
             mod_options=ModuleOptions(pipeline_index=pipeline_index, **mod_options),
         )
         task = assert_not_none(maybe_task)
-        if task.future is not None:
-            # If we had to launch something, register the callback.
+        task_launched = task.future is not None
+        if task_launched:
+            dm = dms[dataset_split_name] if dataset_split_name != DatasetSplitName.all else dms
```
JosephMarinier, line 201: ... When I use an `else`, I avoid inverting/negating the condition to keep it simple. Otherwise, this is like a double negation. In this case, I would use `==` instead of `!=` and flip.
```suggestion
            dm = dms if dataset_split_name == DatasetSplitName.all else dms[dataset_split_name]
```

azimuth/startup.py
```diff
@@ -146,16 +183,22 @@ def make_startup_tasks(
     dms = {k: v for k, v in dataset_split_managers.items() if v is not None}
     tasks: Dict[DatasetSplitName, DaskModule] = {}
 
-    for dataset_split_name, dm in dms.items():
+    if DatasetSplitName.all not in dataset_split_names:
+        available_datasets_splits = list(set(dataset_split_names) & set(dms))
+    else:
+        available_datasets_splits = [DatasetSplitName.all]
```
JosephMarinier, line 189: Similarly, here I would remove the `not` and flip. Also, there is no `s` to `datasets`.
```suggestion
    if DatasetSplitName.all in dataset_split_names:
        available_dataset_splits = [DatasetSplitName.all]
    else:
        available_dataset_splits = list(set(dataset_split_names) & set(dms))
```
Alternatively though, here is another way with less list->set and set->list conversions:
```suggestion
    available_dataset_splits = [n for n in dataset_split_names if n in dms or n == DatasetSplitName.all]
```

JosephMarinier (commented):
I suspect we will end up not using those lines, but just in case, some small details...

azimuth/startup.py
```diff
@@ -192,23 +235,16 @@ def startup_tasks(
             start_up_tasks += BMA_PREDICTION_TASKS
         if config.pipelines is not None and len(config.pipelines) > 1:
             start_up_tasks += PIPELINE_COMPARISON_TASKS
+        # TODO We only check pipeline_index=0, but we should check all pipelines.
+        if postprocessing_editable(task_manager.config, 0):
+            start_up_tasks += POSTPROCESSING_EVAL_TASKS
     if similarity_available(task_manager.config):
         start_up_tasks += SIMILARITY_TASKS
+        if DatasetSplitName.train in dataset_split_managers:
```
JosephMarinier, line 243: Am I understanding correctly that `start_tasks_for_dms()` still can't handle a `Startup` task if there is no intersection between its `dataset_split_names` and the available dataset splits? I find that unfortunate, especially since here we are duplicating the logic that the `ClassOverlap` module needs the `train` set. Wouldn't it be easy to filter the `start_up_tasks` based on their `dataset_split_names`? No mistake possible, then.

azimuth/startup.py
gabegma: That's a great question, I'm not sure if adding them makes sense. Thanks for pointing this out. 

These are not real dependencies, in the sense that they don't require any information saved in the dataset by other modules. If we call the perturbation module before calling the prediction module, it will not crash because the perturbation module will call the prediction module.  Other modules don't work like that (like the confusion matrix), where it will crash if the prediction were not called because the confusion matrix module relies on the predictions being saved in the dataset.  

However, I thought it would be good to add prediction as dependencies for both perturbation testing and prediction comparison, to avoid the predictions being called twice simultaneously on 2 different workers. For example, if perturbation testing is being sent to one worker, and prediction to the other one, my intuition was that they could be both running the module. I'm not sure if my intuition is right though, Dask may already take care of that.  I thought it would not cost anything to add the dependencies and make sure this doesn't happen. @Dref360 - wdyt?

azimuth/startup.py
gabegma: Absolutely, thanks for taking the time to simplify!

azimuth/startup.py
gabegma: I think you are right. I was almost there with the changes in this PR, but I didn't realize I could now get rid of this. Let me try!

azimuth/startup.py
gabegma: I changed it a bit because `mypy` complained, and I also realized with should not be sending `dms` with `DatasetSplitName.all`, no `dm` is required, so I'm just sending `None`.

azimuth/startup.py
gabegma: I opened a new PR for that. https://github.com/ServiceNow/azimuth/pull/357

azimuth/startup.py
JosephMarinier: Ho cool! Good job!

azimuth/startup.py
```diff
@@ -41,14 +41,24 @@ class Startup:
...
     Startup("neighbors_tags", SupportedModule.NeighborsTagging),
+    Startup(
+        "class_overlap",
+        SupportedModule.ClassOverlap,
+        dependency_names=["neighbors_tags"],
+        dataset_split_names=[DatasetSplitName.train],
+    ),
 ]
 
+
```
JosephMarinier, line 61: I'm nitpicking, but the other spaces are simple, so it was fine.
```suggestion
```

JosephMarinier (approved):
Good job! And thank you, I learned a bunch of things. 👍

PR #333, opened 2022-12-08
Fix webapp crash when no pipeline

### LINKED ISSUE #331: Web app no longer loading when there is no pipeline
The web app no longer loads when there is no pipeline. It can be reproduced using `make CFG_PATH=local_configs/development/clinc_dummy/conf_no_model.json launch.local`, 

Image

JosephMarinier:
Resolve #331

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -162,11 +162,11 @@ const PageHeader = () => {
                   <span className={classes.label}>Project:</span>
                   <span>{datasetInfo.projectName}</span>
                 </Typography>
-                {config && (
+                {config?.pipelines && (
                   <PipelineSelect
                     selectedPipeline={pipeline.pipelineIndex}
                     onChange={setPipeline}
-                    pipelines={config.pipelines!}
+                    pipelines={config.pipelines}
                   />
                 )}
               </>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,5 +20,6 @@ Released changes are shown in the
 
 ### Fixed
 - Showing long utterances fully on hover in similar and perturbed utterances tables.
+- Webapp crash when no pipeline.
 
 ### Security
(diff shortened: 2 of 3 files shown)


webapp/src/components/PageHeader.tsx
```diff
@@ -162,11 +162,11 @@ const PageHeader = () => {
                   <span className={classes.label}>Project:</span>
                   <span>{datasetInfo.projectName}</span>
                 </Typography>
-                {config && (
+                {config?.pipelines && (
                   <PipelineSelect
                     selectedPipeline={pipeline.pipelineIndex}
                     onChange={setPipeline}
-                    pipelines={config.pipelines!}
```
JosephMarinier, line 169: This inappropriate `!` was letting a `null` in, which was crashing `PipelineSelect`.

christyler3030 (approved):
This looks good to me. Just want to confirm that this is the desired behaviour:
No pipelines -> don't render the selector
rather than 
No pipelines -> render empty selector.

webapp/src/components/PageHeader.tsx
```diff
@@ -162,11 +162,11 @@ const PageHeader = () => {
...
+                    pipelines={config.pipelines}
```
JosephMarinier, line 169: > This looks good to me. Just want to confirm that this is the desired behaviour:
> No pipelines -> don't render the selector
> rather than
> No pipelines -> render empty selector.

Good question @christyler3030. 👍 I wanted to do the quickest fix possible so JavaScript stops crashing and it was midnight so I didn't think too much. Let's take the time now, it's not so much of a rush.

I like your alternative suggestion. In fact, that's how it used to work before the bug was introduced [here](https://github.com/ServiceNow/azimuth/pull/228/files#diff-6e6d6777ad0934d6e5d1e1f84de0cc9b59f59595cd83ebaee179d917024e67bcL157-R160), so I'll revert to that. It has two (other) benefits:
1. The place for the select menu is taken while the config loads, as opposed to popping and moving the project name once it loads.
2. This implementation supports `pipelines` being `[]` the same as when it is `null`.
```suggestion
                    pipelines={config?.pipelines ?? []}
```

On that topic, note that the back end currently crashes with an unclear error message when you configure `"pipeline": []`. I created an issue to look into this: #335

christyler3030 (approved):

PR #336, opened 2022-12-09
Remove `as` with more precise typing

JosephMarinier:


nandhinibsn (commented):
can we not use the type"SubconfigKeys" in two other places [line no# 64, 107] where we have declared keyof AzimuthConfig?

webapp/src/types/models.ts
```diff
@@ -68,3 +68,8 @@ export type Bin = {
   name: number;
   type: Outcome;
 };
+
+// Like Pick<Type, Keys>, but with the type of Values instead of the type of Keys.
```
JosephMarinier, line 72: @nandhinibsn

> Is the purpose of adding "PickByValue" to allow only the keys of AzimuthConfig but not some random keys that do not exist in the AzimuthConfig?. But, keyof AzimuthConfig does the similar thing here right?

No. Check out [`Pick`'s documentation](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) if you are not familiar with it. `PickByValue` is similar, as I commented here.

For example, if we have
```ts
type AzimuthConfig = {
  name?: string;
  dataset: components["schemas"]["CustomObject"];
  model_contract?: components["schemas"]["SupportedModelContract"];
};
```
then `PickByValue<Required<AzimuthConfig>, object>` picks only the fields of `AzimuthConfig` that extend `object`, so `name` disappears:
```ts
{
  dataset: components["schemas"]["CustomObject"];
  model_contract: components["schemas"]["SupportedModelContract"];
}
```
So, with `config: SubConfigKeys`, we know that `resultingConfig[config]` will one of those values that extend `object` and therefore can be "spread" with `...`. That makes the error `Spread types may only be created from object types.` disappear.

Does that help?

JosephMarinier: @nandhinibsn

> can we not use the type"SubconfigKeys" in two other places [line no# 64, 107] where we have declared keyof AzimuthConfig?

I don't think the other places should have that restriction. For example, `FIELDS_TRIGGERING_STARTUP_TASKS` should contain `"rejection_class"`, which is a `string`, not an `object`.

nandhinibsn: Ahh!. I was not aware of the pick concept before, so I went and did some reading. But your example explained more about the extend object and how it solved the error message. Thank you for clarifying and for your code suggestion. Can we merge this into my PR after Chris's approval, or should I apply these changes?

nandhinibsn: > don't think the other places should have that restriction. For example, `FIELDS_TRIGGERING_STARTUP_TASKS` should contain `"rejection_class"`, which is a `string`, not an `object.`

Understood now, Thanks! I was not able to understand earlier what the extend object does so asked this question.

JosephMarinier: Of course! My pleasure. We can merge this PR into yours. And yes, Chris had a comment, so let's get his input first.

christyler3030 (approved):

PR #349, opened 2022-12-23
Save config in cache when starting the app and when modifying config

### LINKED ISSUE #345: Save config files in caching folder with json lines
So that user can retrieve their initial config and the subsequent changes that were made.

gabegma:
Resolve #345 

## Description:
- I took the opportunity to do some refactoring so we can skip the validation tasks and start-up tasks in some cases. For now, this is only useful in the `update_config` route, in the `except`  condition. 
- This refactoring triggered some new `mypy` warnings that I have fixed.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/app.py
+++ b/azimuth/app.py
@@ -6,6 +6,7 @@ from threading import Event
 from typing import Dict, Optional
 
 import structlog
+from distributed import SpecCluster
 from fastapi import APIRouter, Depends, FastAPI, HTTPException
 from starlette.middleware.cors import CORSMiddleware
 from starlette.status import HTTP_403_FORBIDDEN, HTTP_404_NOT_FOUND
@@ -13,19 +14,19 @@ from starlette.status import HTTP_403_FORBIDDEN, HTTP_404_NOT_FOUND
 from azimuth import startup
 from azimuth.config import AzimuthConfig, load_azimuth_config
 from azimuth.dataset_split_manager import DatasetSplitManager
-from azimuth.modules.base_classes import Module
+from azimuth.modules.base_classes import DaskModule
 from azimuth.modules.utilities.validation import ValidationModule
 from azimuth.task_manager import TaskManager
 from azimuth.types import DatasetSplitName, ModuleOptions
 from azimuth.utils.cluster import default_cluster
 from azimuth.utils.conversion import JSONResponseIgnoreNan
 from azimuth.utils.logs import set_logger_config
-from azimuth.utils.project import load_dataset_split_managers_from_config
+from azimuth.utils.project import load_dataset_split_managers_from_config, save_config
 from azimuth.utils.validation import assert_not_none
 
 _dataset_split_managers: Dict[DatasetSplitName, Optional[DatasetSplitManager]] = {}
 _task_manager: Optional[TaskManager] = None
-_startup_tasks: Optional[Dict[str, Module]] = None
+_startup_tasks: Optional[Dict[str, DaskModule]] = None
 _azimuth_config: Optional[AzimuthConfig] = None
 _ready_flag: Optional[Event] = None
 
@@ -56,7 +57,7 @@ def get_task_manager() -> Optional[TaskManager]:
     return _task_manager
 
 
-def get_startup_tasks() -> Optional[Dict[str, Module]]:
+def get_startup_tasks() -> Optional[Dict[str, DaskModule]]:
     return _startup_tasks
 
 
@@ -98,10 +99,11 @@ def start_app(config_path, debug=False) -> FastAPI:
     azimuth_config = load_azimuth_config(config_path)
     if azimuth_config.dataset is None:
         raise ValueError("No dataset has been specified in the config.")
+    save_config(azimuth_config)
 
     local_cluster = default_cluster(large=azimuth_config.large_dask_cluster)
 
-    initialize_managers(azimuth_config, local_cluster)
+    run_start_up_tasks(azimuth_config, local_cluster)
     assert_not_none(_task_manager).client.run(set_logger_config, level)
 
     app = create_app()
@@ -227,16 +229,15 @@ def create_app() -> FastAPI:
     return app
 
 
-def initialize_managers(azimuth_config, cluster):
-    """Initialize manager objects.
+def initialize_managers(azimuth_config: AzimuthConfig, cluster: SpecCluster):
+    """Initialize DatasetSplitManagers and TaskManagers.
 
-    Initialize DatasetSplitManagers, TaskManagers, start startup tasks.
 
     Args:
         azimuth_config: Configuration
         cluster: Dask cluster to use.
     """
-    global _task_manager, _dataset_split_managers, _startup_tasks, _azimuth_config, _ready_flag
+    global _task_manager, _dataset_split_managers, _azimuth_config
     _azimuth_config = azimuth_config
     if _task_manager is not None:
         task_history = _task_manager.current_tasks
@@ -248,14 +249,6 @@ def initialize_managers(azimuth_config, cluster):
     _task_manager.current_tasks = task_history
 
     _dataset_split_managers = load_dataset_split_managers_from_config(azimuth_config)
-    # Validate that everything is in order **before** the startup tasks.
-    if _dataset_split_managers.get(DatasetSplitName.train):
-        run_validation(DatasetSplitName.train, _task_manager, azimuth_config)
-    if _dataset_split_managers.get(DatasetSplitName.eval):
-        run_validation(DatasetSplitName.eval, _task_manager, azimuth_config)
-
-    _startup_tasks = startup.startup_tasks(_dataset_split_managers, _task_manager)
-    _ready_flag = Event()
 
 
 def run_validation(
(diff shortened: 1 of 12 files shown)


azimuth/app.py
```diff
@@ -288,3 +282,24 @@ def run_validation_module(pipeline_index=None):
             run_validation_module(pipeline_index)
     task_manager.clear_worker_cache()
     task_manager.restart()
+
+
+def run_start_up_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
```
JosephMarinier, line 287: I think we use `startup` everywhere else. Technically, it should be "start-up", but we can't use an hyphen here.
```suggestion
def run_startup_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
```

azimuth/utils/project.py
```diff
@@ -51,6 +52,18 @@ def load_dataset_from_config(azimuth_config: AzimuthConfig) -> DatasetDict:
...
 
 
+def save_config(azimuth_config: AzimuthConfig):
+    """Append config to configs.jsonl to retrieve past configs."""
+    with jsonlines.open(f"{azimuth_config.get_artifact_path()}/configs.jsonl", mode="a") as f:
+        f.write(azimuth_config.dict())
+
+
+def update_new_config(old_config: AzimuthConfig, change_set: Dict) -> AzimuthConfig:
+    new_config = old_config.copy(update=change_set, deep=True)
```
JosephMarinier, line 62: Detail... I see where `change_set` comes from, but maybe we could get closer to `copy`'s syntax with `update`?
Also, is there a reason why the function is called `update_new_config` and not simply `update_config`?
```suggestion
def update_config(old_config: AzimuthConfig, update: Dict) -> AzimuthConfig:
    new_config = old_config.copy(update=update, deep=True)
```

tests/test_config.py
```diff
@@ -134,3 +136,27 @@ def test_french_defaults_and_override():
...
+
+    change_set_2 = {"dataset_warnings": {"min_num_per_class": 40}}
+    new_config_2 = update_new_config(new_config, change_set_2)
+
+    assert new_config_2.dataset_warnings.min_num_per_class == 40
+
+    with open(f"{new_config_2.get_artifact_path()}/configs.jsonl", "r") as f:
+        lines = f.read().splitlines()
+        assert len(lines) == 2
+        loaded_config_2 = json.loads(lines[-1])
```
JosephMarinier, line 160: If you would like to reuse `jsonlines` here, to be more robust in splitting the file than `splitlines()`, I think you could do something like this, leaving `loaded_config_2` with the value if the last line (not tested):
```suggestion
    count = 0
    with jsonlines.open(f"{new_config_2.get_artifact_path()}/configs.jsonl", "r") as reader:
        for loaded_config_2 in reader:
            count++
    assert count == 2
```

azimuth/utils/project.py
gabegma: No, good idea! And `change_set` was already there, I agree that it wasn't a very good name.

tests/test_config.py
gabegma: Good recommendation! I did it with a list comprehension! I don't think `count++` exists in python.

azimuth/utils/project.py
```diff
@@ -51,6 +52,18 @@ def load_dataset_from_config(azimuth_config: AzimuthConfig) -> DatasetDict:
...
+    save_config(new_config)
```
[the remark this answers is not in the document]
gabegma: Great point! I changed that in my last commit. I also changed the location of the jsonl file so it's in the parent cache folder. This makes more sense and would eventually allow for the command that would run the last config.

tests/test_config.py
```diff
@@ -134,3 +136,30 @@ def test_french_defaults_and_override():
...
+    new_config_2 = update_config(new_config, partial_config_2)
+    assert new_config_2.dataset_warnings.min_num_per_class == 40
+    save_config(new_config_2)
+
+    with jsonlines.open(f"{new_config_2.artifact_path}/configs.jsonl", "r") as reader:
+        all_configs_2 = [config for config in reader]
+
+    assert len(all_configs_2) == 2
+    loaded_config_2 = all_configs_2[-1]
+    assert loaded_config_2 == new_config_2
```
JosephMarinier, line 165: It seems we don't need all those `_2` variables and we can reuse the same variables. I find it makes it obvious for the reader that the second part of the test is independent from the first one.
```suggestion
    with jsonlines.open(f"{new_config.artifact_path}/configs.jsonl", "r") as reader:
        all_configs = list(reader)
    assert len(all_configs) == 1
    assert all_configs[0] == new_config

    # Changing config for a second time
    partial_config = {"dataset_warnings": {"min_num_per_class": 40}}
    new_config = update_config(new_config, partial_config)
    assert new_config.dataset_warnings.min_num_per_class == 40
    save_config(new_config)

    with jsonlines.open(f"{new_config.artifact_path}/configs.jsonl", "r") as reader:
        all_configs = list(reader)

    assert len(all_configs) == 2
    assert all_configs[-1] == new_config
```

gabegma: Ok! All other tests are like that though, I thought it was always best to not reuse variables, so that is good to know!

tests/test_routers/admin/test_admin.py
```diff
@@ -111,14 +112,25 @@ def test_update_config(app: FastAPI, wait_for_startup_after):
...
+    with jsonlines.open(jsonl_file_path, "r") as reader:
+        loaded_configs = [config for config in reader]
+    assert len(loaded_configs) == 2, "Config have been modified once."
+    assert loaded_configs[-1]["model_contract"] == "file_based_text_classification"
+    assert not loaded_configs[-1]["pipelines"]
 
     res = client.patch("/admin/config", json={"model_contract": "potato"})
     assert res.status_code == 400
+    with jsonlines.open(jsonl_file_path, "r") as reader:
+        loaded_configs = [config for config in reader]
```
JosephMarinier, line 131: Similarly
```suggestion
        loaded_configs = list(reader)
    assert len(loaded_configs) == 2, "Config have been modified once."
    assert loaded_configs[-1]["model_contract"] == "file_based_text_classification"
    assert not loaded_configs[-1]["pipelines"]

    res = client.patch("/admin/config", json={"model_contract": "potato"})
    assert res.status_code == 400
    with jsonlines.open(jsonl_file_path, "r") as reader:
        loaded_configs = list(reader)
```

tests/test_config.py
JosephMarinier: [Here](https://github.com/ServiceNow/azimuth/pull/349/files#diff-a6ba4243c32b086df6de7b529feda770ffbdb35b39c4f400a354945e363144edR122-R131) for example, you already reuse the same `loaded_configs` variable.

azimuth/utils/project.py
```diff
@@ -52,6 +53,16 @@ def load_dataset_from_config(azimuth_config: AzimuthConfig) -> DatasetDict:
     return dataset_train_eval
 
 
+def save_config(azimuth_config: AzimuthConfig):
+    """Append config to configs.jsonl to retrieve past configs."""
```
JosephMarinier, line 57: Let's keep a note for if we come across a problem with the file access not being atomic.
```suggestion
    """Append config to configs.jsonl to retrieve past configs."""
    # TODO https://stackoverflow.com/questions/2333872/how-to-make-file-creation-an-atomic-operation
```

azimuth/utils/project.py
```diff
@@ -52,6 +53,16 @@ def load_dataset_from_config(azimuth_config: AzimuthConfig) -> DatasetDict:
...
+    with jsonlines.open(f"{azimuth_config.artifact_path}/configs.jsonl", mode="a") as f:
```
JosephMarinier, line 58: The name of the file... I'm afraid plural `configs.jsonl` could be interpreted as multople configs, whereas this is more like the one config, but recording the past versions. What would you think of `config_history.jsonl`?
```suggestion
    with jsonlines.open(f"{azimuth_config.artifact_path}/config_history.jsonl", mode="a") as f:
```

gabegma: Now that it was moved to the parent cache folder, it can also be multiple different configs though. Do you still prefer `config_history`?

JosephMarinier: Right. So, I feel we are going through a transition right now, so all might not be definitive.

I think we need to tie one `.jsonl` per Azimuth instance. If one day we are to start Azimuth without a config (or almost no config), we need a way to identify which `.jsonl` to load. I think that's the `artifact_path`, and that the `jsonl` is in the correct location. Eventually, if one shuts down Azimuth and restarts it (with the same `artifact_path`), I would expect it to load the `.jsonl` and resume with the last config. In that context, if one runs multiple Azimuth instances, I would expect a different `artifact_path` per instance. Considering you could start 2 Azimuth instances with "no config", there is no other way to tell them which `.jsonl` to look at.


Does that make sense?

As I mentioned before, I am questioning the need for the cache "sub-folders". Even if one changes the pipelines completely (for example one adds a pipeline once a model is trained), the dataset stuff that is cached would still be valid.

JosephMarinier (approved):
Very cool! That paves the way for a lot of cool features. 👍

azimuth/utils/project.py
gabegma: Thanks for further explaining, this makes sense to me!

PR #350, opened 2022-12-23
Exclude fields from cache

### LINKED ISSUE #348: Refactor which fields from the config should impact caching.
We should add a new field 'extra' for all values in the config which should not affect the caching hash, instead of using the 'CommonFieldsConfig'.

JosephMarinier:
Resolve #348

## Description:

Exclude desired fields from cache hash by setting an extra parameter `exclude_from_cache=True`. This allows for more granular control over which config fields (when they are edited) will trigger some modules to recompute. For example, we don't need to recompute the predictions if a pipeline's name changes, so we add `exclude_from_cache=True` to the `name` field.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -13,7 +13,7 @@ from pydantic import BaseModel, BaseSettings, Extra, Field, root_validator, vali
 
 from azimuth.types import AliasModel, DatasetColumn, SupportedModelContract
 from azimuth.utils.conversion import md5_hash
-from azimuth.utils.exclude_fields_with_extra import exclude_fields_with_extra
+from azimuth.utils.exclude_fields_from_cache import exclude_fields_from_cache
 
 log = structlog.get_logger(__file__)
 T = TypeVar("T", bound="ProjectConfig")
@@ -256,7 +256,7 @@ class ProjectConfig(BaseSettings):
         return md5_hash(
             self.dict(
                 include=ProjectConfig.__fields__.keys(),
-                exclude=exclude_fields_with_extra(self, "exclude_from_cache"),
+                exclude=exclude_fields_from_cache(self),
                 by_alias=True,
             )
         )
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -128,7 +128,7 @@ class ThresholdConfig(BaseSettings, CustomObject):
 
 
 class PipelineDefinition(BaseSettings):
-    name: str
+    name: str = Field(exclude_from_cache=True)
     model: CustomObject
     postprocessors: Optional[
         List[Union[TemperatureScaling, ThresholdConfig, CustomObject]]
@@ -233,7 +233,7 @@ class ColumnConfiguration(BaseModel):
 
 class ProjectConfig(BaseSettings):
     # Name of the current project.
-    name: str = Field("New project", env="NAME")
+    name: str = Field("New project", env="NAME", exclude_from_cache=True)
     # Dataset object definition.
     dataset: CustomObject
     # Which model_contract the application is using.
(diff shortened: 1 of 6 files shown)


azimuth/config.py
```diff
@@ -252,7 +273,12 @@ def copy(self: T, *, validate: bool = True, **kwargs: Any) -> T:
         return copy
 
     def to_hash(self):
-        return md5_hash(self.dict(include=ProjectConfig.__fields__.keys(), by_alias=True))
+        return md5_hash(
+            self.dict(
+                exclude=exclude_fields_with_extra(self.__class__, "exclude_from_cache"),
```
gabegma, line 278: I think that we still need `include=ProjectConfig.__fields__.keys()`, since this determines the hash for the parent cache folder, which should just include the project fields.

azimuth/modules/base_classes/module.py
```diff
@@ -45,9 +45,11 @@ def __init__(
...
     def _get_name(self) -> str:
-        options_to_consider = self.mod_options.dict()
-        # Indices are excluded, since the cache for all indices should be in the same folder.
-        options_to_consider = self.mod_options.dict(exclude={"indices"})
+        # indices are excluded, since the cache for all indices should be in the same file.
+        # model_contract_method_name are excluded too because it's already in the task_name.
+        options_to_consider = self.mod_options.dict(
+            exclude={"indices", "model_contract_method_name"}, include=self.allowed_mod_options
+        )
         attributes_to_consider = self.config.dict(
```
gabegma, line 53: This is where you'll want to implement the exclude_from_cache logic.

tests/test_exclude_from_cache.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Salad(BaseModel):
+    sub: Sub
+    excluded_sub_with_excluded_field: SubWithExcludedField = Field(..., exclude_from_cache=True)
+    sub_with_excluded_field: SubWithExcludedField
+    list_of_sub_with_excluded_field: List[SubWithExcludedField]
+
+
+def test_exclude_from_cache():
+    salad = Salad(
```
gabegma, line 25: Can we get rid of the salad, since there are not any potatoes left? :D

gabegma (commented):
Quick review, but I need to look into more details the exclude function so I can understand it :)

Thank you for all your help on this!

azimuth/modules/base_classes/module.py
JosephMarinier: Ah-ha! Wrong hash. Sorry about that! Thank you!

tests/test_exclude_from_cache.py
JosephMarinier: That is a good point. Thanks!

azimuth/utils/exclude_fields_with_extra.py
```diff
@@ -0,0 +1,40 @@
...
+# in the root directory of this source tree.
+
+from typing import Collection, Dict, Mapping, TypeVar
+
+from pydantic import BaseModel
+
+Model = TypeVar("Model", bound=BaseModel)
+
+
+def exclude_fields_with_extra(model: Model, extra: str):
```
JosephMarinier, line 12: We can revisit how "specific" we want this function to be. Right now, it accepts a parameter `extra` and we pass it `"exclude_from_cache"` in 2 places (plus 1 in unit tests). That's useful if one day we have a use case for it that is other than `"exclude_from_cache"`.
Alternatively, we could remove the `extra` parameter and hardcode `"exclude_from_cache"` inside, so we don't have to pass it in 2 places. We should then rename the function and the file to be that specific.
```suggestion
def exclude_fields_from_cache(model: Model):
```
In between, we can give `extra` a default value of `"exclude_from_cache"`.
```suggestion
def exclude_fields_with_extra(model: Model, extra: str = "exclude_from_cache"):
```
What do you think?

gabegma: I'm ok with either - what would be the best practice? Since we only have one use case for now, YAGNI?

JosephMarinier: Oh-ho! So extreme programming terms! 🤟 

Let's think about it...

The `extra` argument is not really an "extra functionality". For example, It's not adding unnecessary lines of code that we would then have to maintain unnecessarily, so it doesn't hurt.

I think the main difference between those is the name of the function.
```py
exclude_fields_with_extra(self.config, "exclude_from_cache")
```
Making the function generic like that allowed me to choose a name I find explains very well what the function does, then the argument like `"exclude_from_cache"` makes it clear for the caller what fields will be excluded. We can see that like separation of concerns; the functionality/logic vs what fields need to be excluded.

I am not able to convey those two pieces of information in one function name, we can't tell what starts where:
```py
exclude_fields_with_exclude_from_cache(self.config)
```
But it might not be necessary. 🤷‍♂️ Maybe the following is clear enough, and the slight benefit of not duplicating the string `"exclude_from_cache"` in two places is worth it.
```py
exclude_fields_from_cache(self.config)
```

JosephMarinier: I think I'll go with
```py
exclude_fields_from_cache(self.config)
```
The function's name gives less detail about how the function works, but I have the opportunity to make the docstring more straightforward and clear:
```diff
-    """Exclude pydantic fields with given extra keyword argument.
+    """Exclude pydantic fields with exclude_from_cache=True.
```

gabegma: I'll let you decide, I think that both are clear.

gabegma (approved):
Fantastic work with this PR!! Thank you for taking my relay on this. I could not have brought it this far.

JosephMarinier:
Of course! Thank you! I enjoyed this one and I learn a bunch about pydantic. 👍

PR #352, opened 2022-12-23
Save modules' metadata in cache

### LINKED ISSUE #344: Save caching metadata for every module for debugging purposes

gabegma:
Resolve #344 

## Description:
- I'm not sure if meta-data is the best name for both the config attributes affecting the modules, and the module options. LMK if you prefer another term.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/caching.py
+++ b/azimuth/modules/base_classes/caching.py
@@ -3,8 +3,10 @@
 # in the root directory of this source tree.
 
 import abc
+import json
 import os
 import time
+from os.path import join as pjoin
 from typing import List, Optional
 
 import h5py
@@ -78,6 +80,10 @@ class HDF5CacheMixin(CachingMechanism):
     name: str
     _cache_lock: str  # Lockfile path
     _cache_file: str  # Cachefile path
+    _cache_metadata: str  # Metadata cachefile
+
+    def get_meta_data(self):
+        raise NotImplementedError
 
     def _store_data_in_cache(self, result: List[ModuleResponse], indices: List[int]):
         """Store `results` in `handle` for some `indices`.
@@ -100,6 +106,10 @@ class HDF5CacheMixin(CachingMechanism):
                 ds = handle.require_dataset(f"{self.name}/{idx}", shape=arr.shape, dtype=arr.dtype)
                 ds[()] = arr
             handle.flush()
+
+        # Save all that affects caching so it can be used for identification and debugging.
+        with open(pjoin(self._cache_metadata), "w+") as f:
+            json.dump(self.get_meta_data().no_alias_dict(), f, indent=4)
 
     @retry(stop_max_attempt_number=5, wait_fixed=0.5)
     def _check_cache_internal(self, indices: Optional[List[int]]):
--- a/azimuth/modules/base_classes/caching.py
+++ b/azimuth/modules/base_classes/caching.py
@@ -6,7 +6,6 @@ import abc
 import json
 import os
 import time
-from os.path import join as pjoin
 from typing import List, Optional
 
 import h5py
@@ -79,10 +78,10 @@ class HDF5CacheMixin(CachingMechanism):
 
     name: str
     _cache_lock: str  # Lockfile path
-    _cache_file: str  # Cachefile path
-    _cache_metadata: str  # Metadata cachefile
+    _cache_file: str  # Cache file path
+    _cache_effective_arguments: str  # Effective arguments cache file
 
-    def get_meta_data(self):
+    def get_effective_arguments(self):
         raise NotImplementedError
 
     def _store_data_in_cache(self, result: List[ModuleResponse], indices: List[int]):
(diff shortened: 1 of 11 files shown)


azimuth/types/general/module_options.py
```diff
@@ -101,3 +101,8 @@ class ModuleOptions(AliasModel):
     cf_reorder_classes: bool = Field(
         True, title="Reorder classes based on reverse Cuthill-Mckee algorithm."
     )
+
+
+class ModuleMetaData(AliasModel):
+    module_options: Dict[str, Any]
+    config_attributes: Dict[str, Any]
```
JosephMarinier, line 108: > I'm not sure if meta-data is the best name for both the config attributes affecting the modules, and the module options.

Right, "metadata" (usually one word) is not wrong, but it's also not very specific. Let's brainstorm...

Correct me if I'm wrong...

First, the `config_attributes` is the config after we apply the "scope" (called `config_scope` in `DaskModule`) and after we exclude the fields that have `exclude_from_cache`.

Similarly, `module_options` is `mod_options` after we exclude fields like `indices` and `model_contract_method_name`.

In both cases, we exclude some fields. We would like to express that in this name, as that's the point of this new class. I'm afraid we won't be able to find a good term that means "with some fields excluded", like "partial", as it says nothing about what's excluded or what remains. However, what remains might be the "effective" configs/options? The remaining fields have an "effect" on the result, and changing them would require recomputing. 🤷‍♂️ It's not suuuper obvious what "effective" means in this context, but with a good docstring, once you get it, I think you will remember.

Now, the union of config+options... Those are the module's "arguments"? Or "parameters"? How about...
[drum roll]...
`ModuleEffectiveArguments`?

Then, we could simply reuse `mod_options` and `config_scope`. Since `ModuleEffectiveArguments` expresses that those contain the "effective" fields, then I think we don't need to repeat it on `mod_options` and `config_scope`.

```suggestion
class ModuleEffectiveArguments(AliasModel):
    mod_options: Dict[str, Any]
    config_scope: Dict[str, Any]
```

To me, the `mod_` is redundant in a `ModuleSomething` class, but we could keep it to be consistent with the rest of the code. I don't have a strong preference.

```suggestion
class ModuleEffectiveArguments(AliasModel):
    options: Dict[str, Any]
    config_scope: Dict[str, Any]
```

What do you think?

azimuth/modules/base_classes/caching.py
```diff
@@ -101,6 +107,10 @@ def _store_data_in_cache(self, result: List[ModuleResponse], indices: List[int])
                 ds[()] = arr
             handle.flush()
 
+        # Save all that affects caching so it can be used for identification and debugging.
+        with open(pjoin(self._cache_metadata), "w+") as f:
```
JosephMarinier, line 111: Detail, but I always hated that habit of renaming `pjoin`. I prefer
```diff
-from os.path import join as pjoin
+import os
```
```suggestion
        with open(os.path.join(self._cache_metadata), "w+") as f:
```

JosephMarinier: Second comment on that line.
This syntax is a nightmare, but the difference between `w` and `w+` is that `w+` gives the ability to read `f`. This is not necessary here, so we can remove the `w`.
```suggestion
        with open(os.path.join(self._cache_metadata), "w") as f:
```
I found a good diagram explaining the syntax, where `truncate` essentially means "empty the file".

azimuth/modules/base_classes/caching.py
```diff
@@ -101,6 +107,10 @@ def _store_data_in_cache(self, result: List[ModuleResponse], indices: List[int])
...
+            json.dump(self.get_meta_data().no_alias_dict(), f, indent=4)
```
JosephMarinier, line 112: In our repo, we format JSON files with 2 indents, so I would suggest
```suggestion
            json.dump(self.get_meta_data().no_alias_dict(), f, indent=2)
```

.gitignore
```diff
@@ -9,6 +9,7 @@ cache
 !poetry.lock
 !yarn.lock
 tests/logs.txt
+webapp/openapi.json
```
JosephMarinier, line 12: You still get issues with that, right? 😞 I'm sorry.

If the pre-commits succeed, then this file should have been be removed. I think you are only left with that file if `openapi-typescript openapi.json` fails. In that case, it would be best to rerun. In other words, we should never commit if this file exists.

That might be extreme, though. I guess it is still annoying. So, here is what I would change in `package.json` to make sure `openapi.json` is deleted, even if the command fails, while still returning the exit code.
```diff
-    "types": "poetry run python ../openapi.py > openapi.json && openapi-typescript openapi.json --output src/types/generated/generatedTypes.ts && rm -f openapi.json"
+    "types": "poetry run python ../openapi.py > openapi.json && openapi-typescript openapi.json --output src/types/generated/generatedTypes.ts; EXIT=$?; rm -f openapi.json; exit $EXIT"
```
```suggestion
```
Would you like me to propose the change in another PR? Or you'll do it here?

azimuth/types/general/module_options.py
gabegma: I like it! Let's go with that, thanks for the brainstorm!!

azimuth/modules/base_classes/caching.py
gabegma: Oh awesome, thank you!! Very helpful!

.gitignore
gabegma: Thank you!! 
It failed on numerous occasions even when there were no changes in the generated types. It seems to fail even more frequently when I do partial commits (with files unstaged). Once, I needed to do it around ~8 times in a row, and in the end, I stashed the unstaged files for it to work. I would often see the `openapi.json` file empty that would be added to the staged files for a few seconds and I was wondering if this could help. But your solution looks better - can you do another PR please so it's clean? Thank you!!

JosephMarinier: Of course!
Thanks for the details.
Here is the PR: #364
Let me know if it helps or not.

JosephMarinier (approved):
Beautiful! 🎩

PR #361, opened 2023-01-05
Support training set only

### LINKED ISSUE #362: Make Azimuth work on just a training set

gabegma:
Resolve #362 

## Description:
* Support only a training set (no eval) for further flexibility. Look at the commits for the details of what was done.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/dataset_analysis/similarity_analysis.py
+++ b/azimuth/modules/dataset_analysis/similarity_analysis.py
@@ -75,22 +75,13 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
     def compute_on_dataset_split(self) -> List[TaggingResponse]:  # type: ignore
         # Get the features from FAISS for all available splits to make sure they are all computed
         # and that we can compute neighbors for all splits.
-        eval_features = (
-            self._get_features_from_faiss(DatasetSplitName.eval)
-            if DatasetSplitName.eval in self.available_dataset_splits
-            else None
-        )
-        train_features = (
-            self._get_features_from_faiss(DatasetSplitName.train)
-            if DatasetSplitName.train in self.available_dataset_splits
-            else None
-        )
+        features_dict = {
+            split: self._get_features_from_faiss(split) for split in self.available_dataset_splits
+        }
 
         dm = self.get_dataset_split_manager()
         indices = self.get_indices()
-        features = assert_not_none(
-            eval_features if self.dataset_split_name == DatasetSplitName.eval else train_features
-        )
+        features = features_dict[self.dataset_split_name]
         neighbors = self.get_neighbors([features[i] for i in indices])
 
         similarity_config: SimilarityOptions = assert_not_none(self.config.similarity)
--- a/azimuth/modules/dataset_analysis/similarity_analysis.py
+++ b/azimuth/modules/dataset_analysis/similarity_analysis.py
@@ -73,26 +73,24 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
     """Compute neighbors for each utterance in a dataset split."""
 
     def compute_on_dataset_split(self) -> List[TaggingResponse]:  # type: ignore
-        def get_dm_and_features(ds_name):
-            if ds_name in self.available_dataset_splits:
-                dm: Optional[DatasetSplitManager] = self.get_dataset_split_manager(ds_name)
-                features: Optional[List[Array]] = self._get_features(ds_name)
-                return dm, features
-            else:
-                return None, None
+        # FAISS is computed for all available splits so neighbors can be computed for all splits.
+        eval_features = (
+            self._get_features_from_faiss(DatasetSplitName.eval)
+            if DatasetSplitName.eval in self.available_dataset_splits
+            else None
+        )
+        train_features = (
+            self._get_features_from_faiss(DatasetSplitName.train)
+            if DatasetSplitName.train in self.available_dataset_splits
+            else None
+        )
 
-        train_dm, train_features = get_dm_and_features(DatasetSplitName.train)
-        eval_dm, eval_features = get_dm_and_features(DatasetSplitName.eval)
-
-        ds_features: List[Array] = assert_not_none(
+        dm = self.get_dataset_split_manager()
+        indices = self.get_indices()
+        features = assert_not_none(
             eval_features if self.dataset_split_name == DatasetSplitName.eval else train_features
         )
-        ds_dm: DatasetSplitManager = assert_not_none(
-            eval_dm if self.dataset_split_name == DatasetSplitName.eval else train_dm
-        )
-
-        indices = self.get_indices()
-        neighbors = self.get_neighbors([ds_features[i] for i in indices])
+        neighbors = self.get_neighbors([features[i] for i in indices])
 
         similarity_config: SimilarityOptions = assert_not_none(self.config.similarity)
 
@@ -106,7 +104,7 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
             if split_name not in self.available_dataset_splits:
                 continue
             conflicting_neighbors_tags[split_name] = self.get_conflicting_neighbors_tag(
-                labels=ds_dm.get_dataset_split().select(indices)[self.config.columns.label],
+                labels=dm.get_dataset_split().select(indices)[self.config.columns.label],
                 neighbors=neighbors[split_name],
                 dm=self.get_dataset_split_manager(split_name),
                 threshold=similarity_config.conflicting_neighbors_threshold,
@@ -159,8 +157,8 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
 
         return results
 
-    def _get_features(self, ds_split_name: DatasetSplitName) -> List[Array]:
-        """Get Sentence embedding features
+    def _get_features_from_faiss(self, ds_split_name: DatasetSplitName) -> List[Array]:
+        """Get Sentence embedding features from FAISS module.
 
         Args:
             ds_split_name: Name of the dataset_split to get
@@ -187,8 +185,8 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
         for col_name in res[0].adds.keys():
             dm.add_column(key=col_name, features=[r.adds[col_name] for r in res])
 
+    @staticmethod
     def get_conflicting_neighbors_tag(
-        self,
         labels: List[int],
         neighbors: List[List[Tuple[int, float]]],
         dm: DatasetSplitManager,
--- a/azimuth/modules/dataset_analysis/similarity_analysis.py
+++ b/azimuth/modules/dataset_analysis/similarity_analysis.py
@@ -73,30 +73,28 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
     """Compute neighbors for each utterance in a dataset split."""
 
     def compute_on_dataset_split(self) -> List[TaggingResponse]:  # type: ignore
-        if DatasetSplitName.train in self.available_dataset_splits:
-            train_dm: Optional[DatasetSplitManager] = self.get_dataset_split_manager(
-                DatasetSplitName.train
-            )
-            # NOTE: Doing this ensure that the FAISS index exists somewhere.
-            # If FAISSModule fails this module will fail instead of blocking.
-            train_features: Optional[List[Array]] = self._get_features(DatasetSplitName.train)
-        else:
-            train_dm = None
-            train_features = None
+        def get_dm_and_features(ds_name):
+            if ds_name in self.available_dataset_splits:
+                dm: Optional[DatasetSplitManager] = self.get_dataset_split_manager(ds_name)
+                features: Optional[List[Array]] = self._get_features(ds_name)
+                return dm, features
+            else:
+                return None, None
 
-        eval_dm = self.get_dataset_split_manager(DatasetSplitName.eval)
-        eval_features = self._get_features(DatasetSplitName.eval)
-        ds_features = assert_not_none(
+        train_dm, train_features = get_dm_and_features(DatasetSplitName.train)
+        eval_dm, eval_features = get_dm_and_features(DatasetSplitName.eval)
+
+        ds_features: List[Array] = assert_not_none(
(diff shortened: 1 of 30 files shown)


christyler3030 (commented):
Frontend stuff looks fine.

azimuth/modules/dataset_analysis/similarity_analysis.py
```diff
@@ -73,30 +73,28 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
     """Compute neighbors for each utterance in a dataset split."""
 
     def compute_on_dataset_split(self) -> List[TaggingResponse]:  # type: ignore
-        if DatasetSplitName.train in self.available_dataset_splits:
-            train_dm: Optional[DatasetSplitManager] = self.get_dataset_split_manager(
-                DatasetSplitName.train
-            )
-            # NOTE: Doing this ensure that the FAISS index exists somewhere.
-            # If FAISSModule fails this module will fail instead of blocking.
```
JosephMarinier, line 81: Just to be sure, is this note not relevant anymore? Or it's just obvious?

azimuth/modules/dataset_analysis/similarity_analysis.py
```diff
@@ -73,30 +73,28 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
...
+
+        train_dm, train_features = get_dm_and_features(DatasetSplitName.train)
+        eval_dm, eval_features = get_dm_and_features(DatasetSplitName.eval)
+
+        ds_features: List[Array] = assert_not_none(
             eval_features if self.dataset_split_name == DatasetSplitName.eval else train_features
         )
+        ds_dm: DatasetSplitManager = assert_not_none(
+            eval_dm if self.dataset_split_name == DatasetSplitName.eval else train_dm
+        )
```
JosephMarinier, line 92: That looks equivalent to something like this, considering `get_dataset_split_manager` already asserts that the dataset split is not `None`. Am I missing something?
```suggestion
        dm = self.get_dataset_split_manager(self.dataset_split_name)
        features = self._get_features(self.dataset_split_name)
```
I'm not sure if we need to wrap `features = assert_not_none(...)`.

azimuth/routers/v1/app.py
```diff
@@ -132,7 +137,9 @@ def custom_metrics_info(
     config: AzimuthConfig = Depends(get_config),
 ) -> Dict[str, MetricInfo]:
     result = {}
-    dm = dataset_split_managers[DatasetSplitName.eval]
+    dm = dataset_split_managers.get(DatasetSplitName.eval) or dataset_split_managers.get(
```
JosephMarinier, line 140: Wanna add an `assert_not_none` here too?

azimuth/utils/project.py
```diff
@@ -43,9 +43,10 @@ def load_dataset_from_config(azimuth_config: AzimuthConfig) -> DatasetDict:
         dataset_train_eval[DatasetSplitName.eval] = dataset_in_config["validation"]
     elif "test" in dataset_in_config and len(dataset_in_config["test"]) > 0:
```
JosephMarinier, line 44: Should we add a ` and len(dataset_in_config["train"]) > 0` for the first `if "train" ...`?

tests/test_loading_resources.py
```diff
@@ -58,11 +58,15 @@ def load_hf_text_classif_pipeline(checkpoint_path: str, azimuth_config: AzimuthC
...
-    # Test has no label
-    ds = {"validation": datasets["validation"].select(np.arange(max_dataset_len))}
+    ds = {}
+    if eval:
+        # Test set has no label
+        datasets["validation"] = datasets["validation"].rename_column("sentence", "utterance")
+        ds.update({"validation": datasets["validation"].select(np.arange(max_dataset_len))})
     if train:
         datasets["train"] = datasets["train"].rename_column("sentence", "utterance")
         ds.update({"train": datasets["train"].select(np.arange(max_dataset_len))})
```
JosephMarinier, line 72: I know it's not new, but there is no need for `update` here.
```suggestion
        ds["train"] = datasets["train"].select(np.arange(max_dataset_len))
```

tests/test_loading_resources.py
```diff
@@ -58,11 +58,15 @@ def load_hf_text_classif_pipeline(checkpoint_path: str, azimuth_config: AzimuthC
...
+) -> DatasetDict:
     datasets = load_dataset("glue", "sst2")
-    datasets["validation"] = datasets["validation"].rename_column("sentence", "utterance")
-    # Test has no label
-    ds = {"validation": datasets["validation"].select(np.arange(max_dataset_len))}
+    ds = {}
+    if eval:
+        # Test set has no label
+        datasets["validation"] = datasets["validation"].rename_column("sentence", "utterance")
+        ds.update({"validation": datasets["validation"].select(np.arange(max_dataset_len))})
```
JosephMarinier, line 69: Same thing here
```suggestion
        ds["validation"] = datasets["validation"].select(np.arange(max_dataset_len))
```

tests/test_modules/test_perturbation_testing/test_perturbation_testing_summary.py
```diff
@@ -39,20 +41,24 @@ def test_perturbation_testing_summary(tiny_text_config):
...
         mod_options=ModuleOptions(pipeline_index=0),
     )
-    [res] = mod.compute_on_dataset_split()
-    assert res.eval_failure_rate >= 0.0 and res.train_failure_rate == 0.0
+    res_dict = mod.compute_on_dataset_split()[0].no_alias_dict()
+    ds_name, other_ds_name = get_tiny_text_config_one_ds_name(tiny_text_config_one_ds)
+    assert (
+        res_dict[f"{ds_name}_failure_rate"] >= 0.0
+        and res_dict[f"{other_ds_name}_failure_rate"] == 0.0
+    )
```
JosephMarinier, line 55: Performance is not that important in tests, but just so you know, if you would like to avoid generating the whole `no_alias_dict()`, you could use the built-in Python `get_attr()` function.
```suggestion
    [res] = mod.compute_on_dataset_split()
    ds_name, other_ds_name = get_tiny_text_config_one_ds_name(tiny_text_config_one_ds)
    assert (
        get_attr(res, f"{ds_name}_failure_rate") >= 0.0
        and get_attr(res, f"{other_ds_name}_failure_rate") == 0.0
    )
```

tests/test_modules/test_perturbation_testing/test_perturbation_testing_summary.py
```diff
@@ -39,20 +41,24 @@ def test_perturbation_testing_summary(tiny_text_config):
...
     mod_sum = PerturbationTestingSummaryModule(
         dataset_split_name=DatasetSplitName.all,
-        config=tiny_text_config_no_train,
+        config=tiny_text_config_one_ds,
         mod_options=ModuleOptions(pipeline_index=0),
     )
 
     res = mod_sum.compute_on_dataset_split()[0]
-    assert all(t.train_count == 0 for t in res.all_tests_summary)
+    assert all(t.no_alias_dict()[f"{other_ds_name}_count"] == 0 for t in res.all_tests_summary)
```
JosephMarinier, line 64: Similarly
```suggestion
    assert all(get_attr(t, f"{other_ds_name}_count") == 0 for t in res.all_tests_summary)
```

tests/test_startup.py
```diff
@@ -69,12 +69,13 @@ def test_on_end(tiny_text_config):
...
-def test_startup_task_no_train(tiny_text_config_no_train, tiny_text_task_manager):
-    dms = load_dataset_split_managers_from_config(tiny_text_config_no_train)
+def test_startup_task_one_ds(tiny_text_config_one_ds, tiny_text_task_manager):
+    dms = load_dataset_split_managers_from_config(tiny_text_config_one_ds)
     assert DatasetSplitName.eval in dms and DatasetSplitName.train in dms
 
     mods = startup_tasks(dms, tiny_text_task_manager)
-    assert all("train" not in k or "eval" in k for k in mods.keys())
+    ds_name, other_ds_name = get_tiny_text_config_one_ds_name(tiny_text_config_one_ds)
+    assert all(other_ds_name not in k or ds_name in k for k in mods.keys())
```
JosephMarinier, line 78: I'm not sure I understand this assertion. Shouldn't it be `and` instead of `or`?
```suggestion
    assert all(other_ds_name not in k and ds_name in k for k in mods.keys())
```

azimuth/modules/dataset_analysis/similarity_analysis.py
gabegma: Thank you for pointing that out. My understanding and changes were completely inadequate. Turns out that the comment was useful. I rewrote it so it's clearer. Let me know what you think.

azimuth/routers/v1/app.py
```diff
@@ -89,17 +91,20 @@ def get_dataset_info(
     task_manager: TaskManager = Depends(get_task_manager),
     config: AzimuthConfig = Depends(get_config),
 ):
-    eval_dm = dataset_split_managers[DatasetSplitName.eval]
+    eval_dm = dataset_split_managers.get(DatasetSplitName.eval)
     training_dm = dataset_split_managers.get(DatasetSplitName.train)
+    dm = assert_not_none(eval_dm if eval_dm is not None else training_dm)
```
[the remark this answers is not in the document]
gabegma: I'm mad at myself for forgetting that one! :)

azimuth/routers/v1/app.py
gabegma: `mypy` is not complaining - you think that it's best to add it regardless?

azimuth/utils/project.py
gabegma: I'm puzzled as to why it wasn't there before! I'll add it!

tests/test_modules/test_perturbation_testing/test_perturbation_testing_summary.py
gabegma: Oh! I'm loving it, thanks!!

JosephMarinier: Hihiii! I'm happy you like it! :) Thanks for the feedback!

azimuth/routers/v1/app.py
JosephMarinier: Ah! It's a good idea to rely on `mypy`. I guess there is no need, then. I wasn't sure where `dm` could be `None` and where not.

azimuth/routers/v1/app.py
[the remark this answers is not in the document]
JosephMarinier: Hahaaa! Don't be! 🐝

tests/test_startup.py
gabegma: It's because some keys have neither because they are computed on `DatasetSplitName.all`. Let me make that explicit. LMK what you think.

azimuth/modules/dataset_analysis/similarity_analysis.py
JosephMarinier: Thank you for taking the time to dig into this and to make sure we understand what is going on. Good job!

tests/test_startup.py
JosephMarinier: Ah-ha! I see. As it was, the ` or ds_name in k` had no effect since `other_ds_name not in k` was always true. Thanks for the explanation.

JosephMarinier: In other words, if the key was `potato1`, `potato2`, etc., the test would have still passed.

tests/test_startup.py
```diff
@@ -69,12 +69,15 @@ def test_on_end(tiny_text_config):
...
-    dms = load_dataset_split_managers_from_config(tiny_text_config_no_train)
+def test_startup_task_one_ds(tiny_text_config_one_ds, tiny_text_task_manager):
+    dms = load_dataset_split_managers_from_config(tiny_text_config_one_ds)
     assert DatasetSplitName.eval in dms and DatasetSplitName.train in dms
 
     mods = startup_tasks(dms, tiny_text_task_manager)
-    assert all("train" not in k or "eval" in k for k in mods.keys())
+    ds_name, other_ds_name = get_tiny_text_config_one_ds_name(tiny_text_config_one_ds)
+    assert all(
+        (DatasetSplitName.all or ds_name in k) and other_ds_name not in k for k in mods.keys()
```
JosephMarinier, line 79: `DatasetSplitName.all or ds_name in k` will always return `DatasetSplitName.all` because it is truthy, so it is not testing anything. The `in` operator has precedence over the `or`, but even if we added some parentheses, `(DatasetSplitName.all or ds_name) in k` would be equivalent to `DatasetSplitName.all in k` since `DatasetSplitName.all` is truthy, which is not what the correct assertion. We need 2 `in`s:
```suggestion
        (DatasetSplitName.all in k or ds_name in k) and other_ds_name not in k for k in mods.keys()
```

gabegma: Oh great catch!! Thank you, I was struggling with how the precedence would work.

azimuth/modules/dataset_analysis/similarity_analysis.py
```diff
@@ -73,30 +73,27 @@ class NeighborsTaggingModule(DatasetResultModule[SimilarityConfig]):
...
+        # and that we can compute neighbors for all splits.
+        eval_features = (
+            self._get_features_from_faiss(DatasetSplitName.eval)
+            if DatasetSplitName.eval in self.available_dataset_splits
+            else None
+        )
+        train_features = (
+            self._get_features_from_faiss(DatasetSplitName.train)
+            if DatasetSplitName.train in self.available_dataset_splits
+            else None
```
JosephMarinier, line 86: Would that work?
```suggestion
        features_dict = {
            split: self._get_features_from_faiss(split) for split in self.available_dataset_splits
        }
```
and
```diff
-        features = assert_not_none(
-            eval_features if self.dataset_split_name == DatasetSplitName.eval else train_features
-        )
+        features = features_dict[self.dataset_split_name]
```

gabegma: Woah, next level! I didn't think of a `Dict`. The tests pass! Awesome thanks!!

JosephMarinier: Awesome! 👍

JosephMarinier (approved):
Let's gooo! 💪

PR #365, opened 2023-01-07
Support no evaluation set

JosephMarinier:
## Description:

Gab had started the work in #361 (against which this PR is made), but it needed some serious refactoring in the front end. I separated the different steps in clear standalone commits.

* [Better manage loading datasetInfo](https://github.com/ServiceNow/azimuth/pull/365/commits/b5f89cf3a170632f106b7a0c42cd5600561ba40f) to remove the need for countless `?`s.
* [Select available dataset by default for behavioral testing preview](https://github.com/ServiceNow/azimuth/pull/365/commits/09b401d758b3a2a299b3be98c4225dbd862b2a52)
* [Select available dataset by default for exploration space](https://github.com/ServiceNow/azimuth/pull/365/commits/4bcae741892aa8d466d5534cf77c5d0ef4171c5a)
* [Manage PreviewCard default content height in the component](https://github.com/ServiceNow/azimuth/pull/365/commits/b89bca8b3522cd180422ec1ba1500a09edaf6b92)
* [Add dataset split in URL for smart tag analysis page](https://github.com/ServiceNow/azimuth/pull/365/commits/31e375b93acea51df12a3d163f1f1c654a331e94)
* [Add dataset split in URL for pipeline performance page](https://github.com/ServiceNow/azimuth/pull/365/commits/057daa03b2ef7734d1739b769b048e1a300ca2fe)
* [Polish error messages](https://github.com/ServiceNow/azimuth/pull/365/commits/80dc07b6c7adcce8d9b582f98918d0fbf8c9d8e1)
* [Add dataset split in URL for class overlap page](https://github.com/ServiceNow/azimuth/pull/365/commits/b72b496c432c4fef119a968daf66ccdb93df93dc) to be consistent with the back end and the other pages.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


webapp/src/components/Analysis/SmartTagsPreviewCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+  defaultDatasetSplitName: DatasetSplitName;
+}> = ({
+  jobId,
+  pipeline,
+  searchString,
+  availableDatasetSplits,
+  defaultDatasetSplitName,
+}) => {
+  const [datasetSplitName, setDatasetSplitName] =
+    React.useState<DatasetSplitName>(defaultDatasetSplitName);
```
JosephMarinier, line 23: Why did I need to create those two new `...PreviewCard` components?
This `React.useState()` can't be called conditionally since it's a hook. However, I need to wait for `datasetInfo` to be loaded before I can choose its initial value `defaultDatasetSplitName`, using `datasetInfo.availableDatasetSplits`. This whole `...PreviewCard` component is rendered conditionally, so there is no issue with the `React.useState` hook.

webapp/src/pages/SmartTags.tsx
```diff
@@ -16,11 +17,18 @@ export const smartTagsDescription = (
 );
 
 const SmartTags = () => {
-  const { jobId } = useParams<{ jobId: string }>();
-  const { pipeline } = useQueryState();
+  const history = useHistory();
```
christyler3030, line 20: Observation: the changes in this file and the previous make me wonder whether there is a generic component we can factor out here. This might be premature however.

christyler3030 (approved):

webapp/src/pages/SmartTags.tsx
JosephMarinier: As discussed offline, I hesitated too. Those `pages` files have some duplication, but it's mostly trivial, like boilerplate stuff. It's sometimes inevitable, so we kind of have to live with it.
In this case, it's not pure duplication, for example, the `setDatasetSplitName` is different, so the new component would need to handle URL magic, but that seems feasible. I say let's see if we get a third component like that and revisit then if it's worth refactoring.

webapp/src/components/Analysis/SmartTagsPreviewCard.tsx
JosephMarinier: We were not happy with that. Here is a refactor: https://github.com/ServiceNow/azimuth/pull/365/commits/e3707e6c9dda51432b833271406f11dc829f2f2d

christyler3030 (approved):

nandhinibsn (commented):
May I know the reason behind changing the default height to auto height for class overlap table?

webapp/src/pages/Dashboard.tsx
```diff
@@ -83,94 +87,100 @@ const Dashboard = () => {
...
             title="Class Overlap"
-            to={`/${jobId}/class_overlap${searchString}`}
+            to={`/${jobId}/dataset_splits/train/class_overlap${searchString}`}
             description={
               <Description
                 text="Assess semantic overlap between class pairs and compare to pipeline confusion."
                 link="user-guide/class-overlap/"
               />
             }
+            autoHeight
```
JosephMarinier, line 104: @nandhinibsn
> May I know the reason behind changing the default height to auto height for class overlap table?

The class overlap preview card has never used the `DEFAULT_PREVIEW_CONTENT_HEIGHT`. That didn't change. I made that the default behavior of `PreviewCard`, so we don't have to pass `DEFAULT_PREVIEW_CONTENT_HEIGHT` to all the majority of `PreviewCards`. The "special" behavior is `autoHeight`, similar to `Table`/`DataGrid`. Does that make sense?

JosephMarinier: The motivation for that change was so I can reuse the `PreviewCard` component in another file (with `DEFAULT_PREVIEW_CONTENT_HEIGHT`), but today I refactored e3707e6c9dda51432b833271406f11dc829f2f2d and that's no longer necessary.
I just noticed that some cards used to set `height` while others used to set `maxHeight`, which was broken with my change, so I just reverted it in d183f7aef25e9dd487ae194fcc352b1c02995630.
Thank you for drawing my attention to that. 👍

nandhinibsn: Okay, I understand. Yes, it makes sense now. Then, we don't require to pass the auto-height property to the preview card component. We can ignore adding it to the class overlap table, right?

JosephMarinier: Yes, I removed it.

nandhinibsn:
The changes look good to me. :)

PR #366, opened 2023-01-09
Improve config page display when values are null

nandhinibsn:
Resolve #

## Description:
Improved null check before displaying fields for not required fields
## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -304,24 +304,36 @@ const Settings: React.FC = () => {
       {displaySectionTitle("General")}
       <FormGroup>
         <Columns columns={3}>
-          {displayReadonlyFields("name", resultingConfig.name)}
-          {displayReadonlyFields(
-            "rejection_class",
-            resultingConfig.rejection_class
+          {resultingConfig.name &&
+            displayReadonlyFields("name", resultingConfig.name)}
+          {resultingConfig.rejection_class &&
+            displayReadonlyFields(
+              "rejection_class",
+              resultingConfig.rejection_class
+            )}
+          {resultingConfig.columns && (
+            <Box display="flex" flexDirection="column">
+              <Typography variant="caption">columns</Typography>
+              <KeyValuePairs>
+                {resultingConfig.columns.text_input && (
+                  <>
+                    <Typography variant="body2">text_input:</Typography>
+                    <Typography variant="body2">
+                      {resultingConfig.columns.text_input}
+                    </Typography>
+                  </>
+                )}
+                {resultingConfig.columns.label && (
+                  <>
+                    <Typography variant="body2">label:</Typography>
+                    <Typography variant="body2">
+                      {resultingConfig.columns.label}
+                    </Typography>
+                  </>
+                )}
+              </KeyValuePairs>
+            </Box>
           )}
-          <Box display="flex" flexDirection="column">
-            <Typography variant="caption">columns</Typography>
-            <KeyValuePairs>
-              <Typography variant="body2">text_input:</Typography>
-              <Typography variant="body2">
-                {resultingConfig.columns?.text_input}
-              </Typography>
-              <Typography variant="body2">label:</Typography>
-              <Typography variant="body2">
-                {resultingConfig.columns?.label}
-              </Typography>
-            </KeyValuePairs>
-          </Box>
         </Columns>
       </FormGroup>
       {displaySectionTitle("Dataset")}
@@ -331,8 +343,10 @@ const Settings: React.FC = () => {
             "class_name",
             resultingConfig.dataset?.class_name
           )}
-          {displayReadonlyFields("remote", resultingConfig.dataset?.remote)}
+          {resultingConfig.dataset?.remote &&
+            displayReadonlyFields("remote", resultingConfig.dataset.remote)}
           {resultingConfig.dataset?.kwargs &&
+            Object.keys(resultingConfig.dataset?.kwargs).length > 0 &&
             displayArgumentsList("kwargs", resultingConfig.dataset.kwargs)}
           {resultingConfig.dataset?.args &&
             resultingConfig.dataset.args.length > 0 &&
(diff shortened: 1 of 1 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -422,15 +440,17 @@ const Settings: React.FC = () => {
...
                     <Columns columns={3}>
-                      {displayReadonlyFields(
-                        "class_name",
-                        postprocessor.class_name
-                      )}
+                      {postprocessor.class_name &&
+                        displayReadonlyFields(
+                          "class_name",
+                          postprocessor.class_name
+                        )}
```
JosephMarinier, line 453: The `class_name` parameter is NOT optional, it must exist.
```suggestion
                      {displayReadonlyFields(
                        "class_name",
                        postprocessor.class_name
                      )}
```

webapp/src/pages/Settings.tsx
```diff
@@ -422,15 +440,17 @@ const Settings: React.FC = () => {
               </FormGroup>
             </FormControl>
             <FormControl>
-              {displayPostprocessorToggleSection(pipelineIndex, pipeline)}
+              {pipeline.postprocessors &&
+                displayPostprocessorToggleSection(pipelineIndex, pipeline)}
```
JosephMarinier, line 444: Wouldn't that defy the purpose of the checkbox? When `pipeline.postprocessors` is `null`, we want the checkbox to be unchecked (as it is already), not hidden.
```suggestion
              {displayPostprocessorToggleSection(pipelineIndex, pipeline)}
```

webapp/src/pages/Settings.tsx
```diff
@@ -304,24 +304,36 @@ const Settings: React.FC = () => {
       {displaySectionTitle("General")}
       <FormGroup>
         <Columns columns={3}>
-          {displayReadonlyFields("name", resultingConfig.name)}
-          {displayReadonlyFields(
-            "rejection_class",
-            resultingConfig.rejection_class
+          {resultingConfig.name &&
+            displayReadonlyFields("name", resultingConfig.name)}
+          {resultingConfig.rejection_class &&
```
JosephMarinier, line 309: I think it would be interesting to show if the `rejection_class` is `null` (instead of hiding the field). This would mean there is no rejection class in the provided dataset, and the user might want to know that. For example, the user might want to edit the config if that was a mistake.

JosephMarinier: Right now, there is an error in the generated types : the `rejection_class` should be `rejection_class?: str | null`. I will fix that in another PR.

webapp/src/pages/Settings.tsx
nandhinibsn: For "TemperatureScaling" and "ThresholdConfig," class_name is declared as optional, whereas in "CustomObject," mentioned "class_name" is required. As we do intersection for all three types, will the class_name be considered "required"?

webapp/src/pages/Settings.tsx
nandhinibsn: Oh right! this is not necessary here. I will revert back to the old change. thanks.

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, I just noticed the generated types. 

So, what do you think we should display in case of null? Is it not only in the case of a rejection class in which the user gets confused? Do we approach this null scenario differently?

JosephMarinier: > Is it not only in the case of a rejection class in which the user gets confused?

No, `null` is a valid use case. I was only giving an example supporting why I think it's important to show it.

I'm not sure how to display it, though. Maybe show `null` in italic? That would differentiate it from a `"null"` string.

When we make the field editable, one day, we could think of a checkbox in the label, disabling the text field if unchecked.

webapp/src/pages/Settings.tsx
JosephMarinier: I see the generated types are a little bit broken here, saying that `postprocessor.class_name` is `class_name?: undefined`. That's confusing. 😝 I'll see if I can do anything about that.
Basically, when defining the config, one can define a `TemperatureScaling` or a `ThresholdConfig` for which the `class_name` has a default value. When getting the config, it will always have the `class_name` defined.

webapp/src/pages/Settings.tsx
JosephMarinier: > `rejection_class` should be `rejection_class?: str | null`

That's ready for review: https://github.com/ServiceNow/azimuth/pull/373

webapp/src/pages/Settings.tsx
JosephMarinier: I played with different possibilities and I won't be able to fix the generated type of `postprocessor.class_name`. If it becomes a problem, let me know and we can instead overwrite the generated type manually.

webapp/src/pages/Settings.tsx
nandhinibsn: > > Is it not only in the case of a rejection class in which the user gets confused?
> 
> No, `null` is a valid use case. I was only giving an example supporting why I think it's important to show it.
> 
> I'm not sure how to display it, though. Maybe show `null` in italic? That would differentiate it from a `"null"` string.
> 
> When we make the field editable, one day, we could think of a checkbox in the label, disabling the text field if unchecked.

So, I would conditionally hide the fields if they are undefined. Otherwise, I will show it as it has from the API response.

webapp/src/pages/Settings.tsx
JosephMarinier: Turns out I found a solution! You can search for `postprocessors?:` in 48513d284700d320f5ed19a11de1e952e9dbe224 to see the result.

webapp/src/pages/Settings.tsx
JosephMarinier: > So, I would conditionally hide the fields if they are undefined.

`rejection_class` cannot be `undefined`. It can be `null`. When it is `null`, we would like to see that. I propose showing _null_ in italic to different it from a `"null"` string.

PR #367, opened 2023-01-09
 Display config args as vertical list

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -200,7 +200,7 @@ const Settings: React.FC = () => {
           overflow="hidden"
           textOverflow="ellipsis"
         >
-          {value}
+          {index + 1}. {value}
         </Typography>
       ))}
     </Box>
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -164,9 +164,9 @@ const Settings: React.FC = () => {
     />
   );
 
-  const displayKeyArgumentsList = (
+  const displayKeywordArguments = (
     name: string,
-    argEntries: Record<string, any> | any[]
+    argEntries: Record<string, any>
   ) => (
     <Box display="grid">
       <Typography variant="caption">{name}</Typography>
@@ -190,20 +190,18 @@ const Settings: React.FC = () => {
     </Box>
   );
 
-  const displayArgumentsList = (name: string, argEntries: string[] | any[]) => (
+  const displayArgumentsList = (name: string, argEntries: any[]) => (
     <Box display="flex" flexDirection={"column"}>
       <Typography variant="caption">{name}</Typography>
       {argEntries.map((value, index) => (
-        <React.Fragment key={index}>
-          <Typography
-            variant="body2"
-            whiteSpace="nowrap"
-            overflow="hidden"
-            textOverflow="ellipsis"
-          >
-            {value}
-          </Typography>
-        </React.Fragment>
+        <Typography
+          variant="body2"
+          whiteSpace="nowrap"
+          overflow="hidden"
+          textOverflow="ellipsis"
+        >
+          {value}
+        </Typography>
       ))}
     </Box>
   );
@@ -351,7 +349,7 @@ const Settings: React.FC = () => {
           )}
           {displayReadonlyFields("remote", resultingConfig.dataset?.remote)}
           {resultingConfig.dataset?.kwargs &&
-            displayKeyArgumentsList("kwargs", resultingConfig.dataset.kwargs)}
+            displayKeywordArguments("kwargs", resultingConfig.dataset.kwargs)}
           {resultingConfig.dataset?.args &&
             resultingConfig.dataset.args.length > 0 &&
             displayArgumentsList("args", resultingConfig.dataset.args)}
@@ -432,7 +430,7 @@ const Settings: React.FC = () => {
                   )}
                   {displayReadonlyFields("remote", pipeline.model.remote)}
                   {pipeline.model.kwargs &&
-                    displayKeyArgumentsList("kwargs", pipeline.model.kwargs)}
+                    displayKeywordArguments("kwargs", pipeline.model.kwargs)}
                   {pipeline.model.args &&
                     pipeline.model.args.length > 0 &&
                     displayArgumentsList("args", pipeline.model.args)}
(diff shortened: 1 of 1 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -190,6 +190,24 @@ const Settings: React.FC = () => {
     </Box>
   );
 
+  const displayArgumentsList = (name: string, argEntries: string[] | any[]) => (
```
JosephMarinier, line 193: If you put ` | any[]`, the `string[]` is useless.

webapp/src/pages/Settings.tsx
```diff
@@ -190,6 +190,24 @@ const Settings: React.FC = () => {
...
+    <Box display="flex" flexDirection={"column"}>
+      <Typography variant="caption">{name}</Typography>
+      {argEntries.map((value, index) => (
+        <React.Fragment key={index}>
```
JosephMarinier, line 197: No need for a fragment here, there is only one component inside it.

webapp/src/pages/Settings.tsx
```diff
@@ -164,7 +164,7 @@ const Settings: React.FC = () => {
     />
   );
 
-  const displayArgumentsList = (
+  const displayKeyArgumentsList = (
     name: string,
     argEntries: Record<string, any> | any[]
```
JosephMarinier, line 169: I guess you want to allow for objects only, now?
```suggestion
    argEntries: Record<string, any>
```

webapp/src/pages/Settings.tsx
```diff
@@ -164,7 +164,7 @@ const Settings: React.FC = () => {
     />
   );
 
-  const displayArgumentsList = (
+  const displayKeyArgumentsList = (
```
JosephMarinier, line 167: Since we are not using this function for lists anymore, we can remove `List` from its name.
Also, in Python, `kwargs` stands for "Keyword Arguments", not "Key Arguments", so I suggest being consistent here:
```suggestion
  const displayKeywordArguments = (
```

webapp/src/pages/Settings.tsx
nandhinibsn: yes, should allow only objects for now.

webapp/src/pages/Settings.tsx
nandhinibsn: In generated types, arg's type is declared as [key: string]: any. whereas args are always supposed to be a list of strings. can we change the type to string[] for it?

webapp/src/pages/Settings.tsx
JosephMarinier: Perfect, thanks! By the way, I think you meant "from now on" and not "for now".

webapp/src/pages/Settings.tsx
JosephMarinier: You need to add the `key` to the `Typography`, though.

webapp/src/pages/Settings.tsx
JosephMarinier: Right, the generated types don't support that well. I think it's fine to use `any[]` for now, and we'll revisit if it poses problem later.

webapp/src/pages/Settings.tsx
nandhinibsn: Keys are just indexes here. As I discussed with @gabegma, we display a list of values for args. Please correct me if I am wrong.  Do we need to show the indexes as well?

nandhinibsn: Added indexes to args.

image

webapp/src/pages/Settings.tsx
```diff
@@ -190,6 +190,22 @@ const Settings: React.FC = () => {
...
+    <Box display="flex" flexDirection={"column"}>
+      <Typography variant="caption">{name}</Typography>
+      {argEntries.map((value, index) => (
+        <Typography
+          variant="body2"
+          whiteSpace="nowrap"
+          overflow="hidden"
+          textOverflow="ellipsis"
+        >
+          {index + 1}. {value}
```
JosephMarinier, line 203: What's the thinking behind adding `{index + 1}. `?

The result would be very similar to before (when using `displayKeywordArguments()`) as the `Object.entries(argEntries)` was taking care of generating this index, plus the values were aligned, with proper spacing, thanks to the grid. Was there any issue with that?

nandhinibsn: Right, I wanted to display a list with serial no's. This PR is made for that purpose.

JosephMarinier: I am not following.
Before this PR, we already had the index on the left
image
In this PR, you removed the index, then in the latest commit, you added it back, but starting from `1` instead of `0`, and with a `.` instead of a `:`. Was that the objective of this PR?

JosephMarinier: This PR introduces two different alignment bugs:
1. the values are not aligned anymore, for example, `100. ` is much wider than `1.`; and
2. the parent grid doesn't work properly anymore.

image

webapp/src/pages/Settings.tsx
```diff
@@ -190,6 +190,22 @@ const Settings: React.FC = () => {
     </Box>
   );
 
+  const displayArgumentsList = (name: string, argEntries: any[]) => (
+    <Box display="flex" flexDirection={"column"}>
```
JosephMarinier, line 194: You can fix issue 2 of https://github.com/ServiceNow/azimuth/pull/367/files#r1068648552 by changing
```suggestion
    <Box display="grid">
```

webapp/src/pages/Settings.tsx
```diff
@@ -190,6 +190,22 @@ const Settings: React.FC = () => {
...
+      <Typography variant="caption">{name}</Typography>
+      {argEntries.map((value, index) => (
```
JosephMarinier, line 196: In JavaScript, the term "entries" usually refers to an array of key-value pairs. I would avoid that term here.

webapp/src/pages/Settings.tsx
JosephMarinier: I asked to add `key={index}` to the `Typography` to resolve the error you probably see in the JavaScript console.

webapp/src/pages/Settings.tsx
nandhinibsn: So, do you suggest changing the parameter name "argEntries" here?

JosephMarinier: Yes

JosephMarinier (approved):
Perfect! 👍

PR #368, opened 2023-01-09
Add better support for csv files

gabegma:
## Description:
* Refactor the syntax tagging module to get rid of the tokenizer for counting tokens. I switched to counting words instead of sub-words, since this is what spacy provides, and it also makes sense. I lowered a bit the default value for a `long_sentence` to account for that.
* Add loading function for csv. 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -279,7 +279,7 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
         return [
             DatasetWarning(
                 name=f"Length mismatch " f"(>{max_delta_mean_words}±{max_delta_std_words} words)",
-                description=f"Delta between the number of words of a "
+                description=f"Delta between the number of words per utterance for a "
                 f"given class in the evaluation set vs the train set is "
                 f"above {max_delta_mean_words}±{max_delta_std_words}.",
                 columns=["mean", "std"],
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -14,7 +14,7 @@ from azimuth.plots.dataset_warnings import (
     class_imbalance_plot,
     class_representation,
     min_nb_samples_plot,
-    nb_tokens_plot,
+    word_count_plot,
 )
 from azimuth.types import DatasetColumn, DatasetSplitName
 from azimuth.types.dataset_warnings import (
@@ -214,47 +214,47 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
         # Get token length and other (see screen 3b)
         train_df = pd.DataFrame(
             {
-                DatasetColumn.token_count: train_ds[DatasetColumn.token_count],
+                DatasetColumn.word_count: train_ds[DatasetColumn.word_count],
                 "label": train_ds[self.config.columns.label],
             }
         )
         eval_df = pd.DataFrame(
             {
-                DatasetColumn.token_count: eval_ds[DatasetColumn.token_count],
+                DatasetColumn.word_count: eval_ds[DatasetColumn.word_count],
                 "label": eval_ds[self.config.columns.label],
             }
         )
 
         # overall stats
-        train_overall = np.mean(cast(List[int], train_ds[DatasetColumn.token_count])), np.std(
-            cast(List[int], train_ds[DatasetColumn.token_count])
+        train_overall = np.mean(cast(List[int], train_ds[DatasetColumn.word_count])), np.std(
+            cast(List[int], train_ds[DatasetColumn.word_count])
         )
-        eval_overall = np.mean(cast(List[int], eval_ds[DatasetColumn.token_count])), np.std(
-            cast(List[int], eval_ds[DatasetColumn.token_count])
+        eval_overall = np.mean(cast(List[int], eval_ds[DatasetColumn.word_count])), np.std(
+            cast(List[int], eval_ds[DatasetColumn.word_count])
         )
 
         # Get mean and std token lengths
         train_desc = train_df.groupby("label").agg(["mean", "std"])
         eval_desc = eval_df.groupby("label").agg(["mean", "std"])
         # Compute divergence
-        divergence = (train_desc - eval_desc)[DatasetColumn.token_count].apply(np.abs)
+        divergence = (train_desc - eval_desc)[DatasetColumn.word_count].apply(np.abs)
         divergence = fill_at_missing_indices(divergence, find_missing_rows(divergence, cls_index))
         divergence_mean = divergence["mean"].to_numpy()
         divergence_std = divergence["std"].to_numpy()
 
         # Make alerts from values.
-        max_delta_mean_tokens = int(self.config.dataset_warnings.max_delta_mean_tokens)
-        max_delta_std_tokens = int(self.config.dataset_warnings.max_delta_std_tokens)
-        alert_norm_mean = divergence_mean > max_delta_mean_tokens
-        alert_norm_std = divergence_std > max_delta_std_tokens
+        max_delta_mean_words = int(self.config.dataset_warnings.max_delta_mean_words)
+        max_delta_std_words = int(self.config.dataset_warnings.max_delta_std_words)
+        alert_norm_mean = divergence_mean > max_delta_mean_words
+        alert_norm_std = divergence_std > max_delta_std_words
         alert_norm = np.any([alert_norm_mean, alert_norm_std], axis=0)
 
         # Get counts, this will have N columns with count and LABEL rows.
         train_count = (
-            train_df.groupby(["label", DatasetColumn.token_count]).size().unstack(fill_value=0)
+            train_df.groupby(["label", DatasetColumn.word_count]).size().unstack(fill_value=0)
         )
         eval_count = (
-            eval_df.groupby(["label", DatasetColumn.token_count]).size().unstack(fill_value=0)
+            eval_df.groupby(["label", DatasetColumn.word_count]).size().unstack(fill_value=0)
         )
 
         train_missing_cols = find_missing_rows(train_count, cls_index)
@@ -278,11 +278,10 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
 
         return [
             DatasetWarning(
-                name=f"Length mismatch "
-                f"(>{max_delta_mean_tokens}±{max_delta_std_tokens} tokens)",
-                description=f"Delta between the number of tokens of a "
+                name=f"Length mismatch " f"(>{max_delta_mean_words}±{max_delta_std_words} words)",
+                description=f"Delta between the number of words of a "
                 f"given class in the evaluation set vs the train set is "
-                f"above {max_delta_mean_tokens}±{max_delta_std_tokens}.",
+                f"above {max_delta_mean_words}±{max_delta_std_words}.",
                 columns=["mean", "std"],
                 format=FormatType.Decimal,
                 comparisons=[
@@ -300,11 +299,11 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
                     )
                     for i, _ in enumerate(train_dist)
                 ],
-                plots=nb_tokens_plot(
+                plots=word_count_plot(
                     train_count_filled,
                     test_count_filled,
-                    max_delta_mean_tokens,
-                    max_delta_std_tokens,
+                    max_delta_mean_words,
+                    max_delta_std_words,
                     train_desc,
                     eval_desc,
                     train_overall,
--- a/azimuth/modules/dataset_analysis/syntax_tagging.py
+++ b/azimuth/modules/dataset_analysis/syntax_tagging.py
@@ -48,7 +48,8 @@ class SyntaxTaggingModule(DatasetResultModule[SyntaxConfig]):
 
         # Always use the BERT tokenizer to ensure this feature works regardless of model
         tokenizer = self.artifact_manager.get_tokenizer()
-        inputs = tokenizer(utterances)
+        # truncate to max model size so it doesn't crash if the utterances are too long.
+        inputs = tokenizer(utterances, truncation=True)
         batch_tokens = [tokenizer.convert_ids_to_tokens(tokens) for tokens in inputs["input_ids"]]
         batch_tokens = [
             [tok for tok in batch_token if tok not in [tokenizer.cls_token, tokenizer.sep_token]]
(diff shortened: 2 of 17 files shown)


azimuth/modules/dataset_analysis/syntax_tagging.py
```diff
@@ -48,7 +48,8 @@ def compute(self, batch: Dataset) -> List[TaggingResponse]:  # type: ignore
 
         # Always use the BERT tokenizer to ensure this feature works regardless of model
         tokenizer = self.artifact_manager.get_tokenizer()
-        inputs = tokenizer(utterances)
+        # truncate to max model size so it doesn't crash if the utterances are too long.
```
lindsaydbrin, line 51: I have a couple thoughts and questions here. I'm assuming this just truncates the utterance (drops the rest). 

1. Should we warn the user when this happens? It's probably worth a note in the docs, but besides that, results will look pretty weird if this happens a bunch in the datasetf - e.g. the token count distribution plot will all be stacked at the max. 
2. The tokenizer is only used for token count (i.e., not syntax tagging). If we want to get an accurate count, maybe we could loop through utterance sections of the correct length and sum the counts?

gabegma: 1. Yes, good point, we should add it to the doc! 
2. Yes, there is probably a clever way to do it, maybe we can add it to our backlog? Right now I at least wanted to make sure that it is not crashing.

JosephMarinier: 2. Ho cool idea! Is that regex chanllenge? 🤩 😝 
```py
sections_of_at_most_12_characters = re.compile("(.{1,12})(?: +|$)")

sections_of_at_most_12_characters.findall("loooooong short short rest")
# ['loooooong', 'short short', 'rest']
```

azimuth_shr/loading_resources.py
```diff
@@ -67,3 +67,32 @@ def load_CLINC150_data(full_path, python_loader, train=True) -> Dataset:
     if not train:
         dst.pop("train")
     return dst
+
+
+def align_labels(ds_dict: DatasetDict, azimuth_config: AzimuthConfig) -> DatasetDict:
+    features: Features = [v.features for v in ds_dict.values()][0]
```
JosephMarinier, line 73: I know this is not new, but while we are here, we can grab the first value in `ds_dict` without doing the rest of the iteration, and without creating a new comprehension list:
```suggestion
    features: Features = next(iter(ds_dict.values())).features
```

azimuth_shr/loading_resources.py
```diff
@@ -67,3 +67,32 @@ def load_CLINC150_data(full_path, python_loader, train=True) -> Dataset:
...
+    if not isinstance(features[azimuth_config.columns.label], ClassLabel):
+        # Get all classes from both set and apply the same mapping to every dataset.
+        classes = sorted(
+            list(set(sum([ds[azimuth_config.columns.label] for ds in ds_dict.values()], [])))
+        )
+        ds_dict = ds_dict.class_encode_column(azimuth_config.columns.label)
+        return ds_dict.align_labels_with_mapping(
+            {class_name: i for i, class_name in enumerate(classes)}, azimuth_config.columns.label
+        )
```
JosephMarinier, line 82: `load_file_dataset()` used to return `ds_dict` no matter if we were getting inside this `if` or not. Now, `align_labels()` (and therefore `load_file_dataset()`) will return `None` if we don't get inside this `if`. The type in the function signature doesn't allow for `None`, so I guess that's a mistake?

You could not return in the `if` and always return `ds_dict`:
```suggestion
    if not isinstance(features[azimuth_config.columns.label], ClassLabel):
        # Get all classes from both set and apply the same mapping to every dataset.
        classes = sorted(
            list(set(sum([ds[azimuth_config.columns.label] for ds in ds_dict.values()], [])))
        )
        ds_dict = ds_dict.class_encode_column(azimuth_config.columns.label)
        ds_dict = ds_dict.align_labels_with_mapping(
            {class_name: i for i, class_name in enumerate(classes)}, azimuth_config.columns.label
        )
    return ds_dict
```

Or do an early return here and unindent the rest:
```suggestion
    if isinstance(features[azimuth_config.columns.label], ClassLabel):
        return ds_dict

    # Get all classes from both set and apply the same mapping to every dataset.
    classes = sorted(
        list(set(sum([ds[azimuth_config.columns.label] for ds in ds_dict.values()], [])))
    )
    ds_dict = ds_dict.class_encode_column(azimuth_config.columns.label)
    return ds_dict.align_labels_with_mapping(
        {class_name: i for i, class_name in enumerate(classes)}, azimuth_config.columns.label
    )
```

azimuth/modules/dataset_analysis/syntax_tagging.py
lindsaydbrin: Re (2)...hmm, your call. My feeling is that it's not intuitive to expect truncation, so it will be easily missed and thus misleading to users. Maybe they'll see a lot of 512s (I *think* that's the limit but I haven't confirmed!) and start asking questions, but it feels buggy. Although it's maybe fine if we note it in the docs, put it in the backlog, and make a point of picking it up soonish. 

In any case, I'm wondering if there is a quick and clever option. I looked at the docs and don't see a way within the tokenizer to alter the behavior of truncation to either return the truncated part or do it in chunks. 

It's a little complex to do exactly outside of the tokenizer because you don't have the tokens yet, but the one thing I saw that might be useful would be to (a) try/except the tokenizer, and if there's an error, (b) use a parser (spaCy) to split into sentences (or maybe there's a max word option that should be safely below the `BertTokenizer` limit), and then run the tokenizer on each sentence.

lindsaydbrin: Ha, Joseph, I had reloaded before you posted yours! Pretty sure it's about tokens rather than characters, but another way to approach it could be to make some assumption about token length (as opposed to sentences or word count).

JosephMarinier: Right, the absolute safest we can be is max 512 characters, then there can't possibly be more than 512 tokens. But indeed, we could probably optimize.

lindsaydbrin: @JosephMarinier what are your thoughts on the need to address this before merging vs. adding it to the backlog?

If we decide it's necessary, I can take it if it'd be helpful or if you're less familiar with the code, @gabegma.

tests/test_loading_resources.py
```diff
@@ -73,24 +74,14 @@ def load_intent_data(train_path, test_path, python_loader) -> Dataset:
     return load_dataset(python_loader, data_files={"train": train_path, "test": test_path})
 
 
-def load_file_dataset(data_files: Dict[str, str], azimuth_config):
+def load_file_dataset(data_files: Dict[str, str], azimuth_config) -> DatasetDict:
     # Load a file dataset and cast the label column as a ClassLabel.
     # Train and test need to be loaded separately because they don't always share the same columns.
     # Train sometimes don't have predictions. HF will complain if we load both together.
```
lindsaydbrin, line 80:
```suggestion
    # Train sometimes doesn't have predictions. HF will complain if we load both together.
```

azimuth_shr/loading_resources.py
```diff
@@ -67,3 +67,32 @@ def load_CLINC150_data(full_path, python_loader, train=True) -> Dataset:
...
+
+def load_csv(azimuth_config, train_path=None, validation_path=None) -> DatasetDict:
+    data_files = dict()
+    if train_path:
+        data_files["train"] = train_path
+    if validation_path:
+        data_files["validation"] = validation_path
+    ds_dict = load_dataset(path="csv", data_files=data_files)
+    ds_dict = align_labels(ds_dict, azimuth_config)
+    # Remove empty utterances that can exist
```
lindsaydbrin, line 93: I find this kind of confusing...can we do one of these?
```suggestion
    # Remove any empty utterances (which may exist in csv input)
```

or just

```suggestion
    # Remove any empty utterances
```

azimuth/modules/dataset_analysis/syntax_tagging.py
gabegma: As discussed offline, we will change the histogram to count words instead of tokens, and remove the usage of the tokenizer altogether. In the module, the variable `tokens_doc` is already a list of all the words present in the utterance.

azimuth_shr/loading_resources.py
```diff
@@ -67,3 +67,32 @@ def load_CLINC150_data(full_path, python_loader, train=True) -> Dataset:
...
         dst.pop("train")
     return dst
+
+
+def align_labels(ds_dict: DatasetDict, azimuth_config: AzimuthConfig) -> DatasetDict:
+    features: Features = [v.features for v in ds_dict.values()][0]
+    if not isinstance(features[azimuth_config.columns.label], ClassLabel):
+        # Get all classes from both set and apply the same mapping to every dataset.
+        classes = sorted(
+            list(set(sum([ds[azimuth_config.columns.label] for ds in ds_dict.values()], [])))
```
JosephMarinier, line 77: Again, not new code, but this way of concatenating lists with `sum([...], [])` is not super readable and it's apparently very inefficient (I say this article https://mathieularose.com/how-not-to-flatten-a-list-of-lists-in-python).
Instead, we can directly "flatten" the lists in the set comprehension:
```suggestion
            list(set(label for ds in ds_dict.values() for label in ds[azimuth_config.columns.label]))
```

azimuth_shr/loading_resources.py
gabegma: Oh, awesome hack, thanks!

azimuth_shr/loading_resources.py
gabegma: Totally, my mistake! Thanks!

azimuth/modules/dataset_analysis/syntax_tagging.py
JosephMarinier: Cool!
Just for reference, or if you are curious @lindsaydbrin
> if we split by chars, we could break a token, whereas if we split by words, we're safe

The solution I suggested does split by **words** (it splits on spaces), but based on a max number of **characters**.

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -278,11 +278,10 @@ def get_syntactic_warnings(
 
         return [
             DatasetWarning(
-                name=f"Length mismatch "
-                f"(>{max_delta_mean_tokens}±{max_delta_std_tokens} tokens)",
-                description=f"Delta between the number of tokens of a "
+                name=f"Length mismatch " f"(>{max_delta_mean_words}±{max_delta_std_words} words)",
+                description=f"Delta between the number of words of a "
```
lindsaydbrin, line 282:
```suggestion
                description=f"Delta between the number of words per sample for a "
```
or "per utterance" or "per example," if not "per sample?" I realize that phrasing isn't new, but if you don't mind updating, I think this would be more clear!

azimuth_shr/loading_resources.py
gabegma: Thank you, yeah the `sum` was puzzling me!

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: I picked utterance because I don't think we use 'sample' often in the UI.

gabegma:
@lindsaydbrin @JosephMarinier - I made some big refactoring to the syntax tagging module - you can look at that commit 9061507 individually to assess that it makes sense.

azimuth/modules/dataset_analysis/syntax_tagging.py
```diff
@@ -42,56 +40,50 @@ def compute(self, batch: Dataset) -> List[TaggingResponse]:  # type: ignore
...
-            do = self.spacy_pipeline(utterance)
-            adds[DatasetColumn.token_count] = len(tokens)
+            doc = spacy_model(clean_utterance(utterance))
+            # Remove punctuation for word count and smart tags
+            tokens = [token.text for token in doc if not token.is_punct]
 
-            if len(list(do.sents)) > 1:
-                tag[SmartTag.multi_sent] = True
+            # Some issues occur with other languages such as french if using doc.sents directly.
+            # Hence, we use an english sentencizer that seems to work better for similar languages.
```
lindsaydbrin, line 61:
```suggestion
            # Hence, we use an English sentencizer that seems to work better for similar languages.
```
I know it's an uncapitalized anglais :)

azimuth/modules/dataset_analysis/syntax_tagging.py
```diff
@@ -42,56 +40,50 @@ def compute(self, batch: Dataset) -> List[TaggingResponse]:  # type: ignore
...
                 smart_tag: False
                 for family in [SmartTagFamily.extreme_length, SmartTagFamily.partial_syntax]
                 for smart_tag in SMART_TAGS_FAMILY_MAPPING[family]
             }
-            adds = {}
-            syntax_options: SyntaxOptions = assert_not_none(self.config.syntax)
 
-            do = self.spacy_pipeline(utterance)
-            adds[DatasetColumn.token_count] = len(tokens)
+            doc = spacy_model(clean_utterance(utterance))
```
lindsaydbrin, line 56: Stylistic choice, so totally ok if you disagree, but I feel like it would be useful to still keep an indication that this is tokenized at this point. (It was `tokens_doc` - maybe there's something better - or maybe you have logic I'm missing?)

azimuth/plots/dataset_warnings.py
```diff
@@ -575,8 +575,8 @@ def nb_tokens_plot(
     """Create the plot with the dropdown for all labels.
 
     Args:
-        train_per_class_count: Training set count of nb of tokens.
-        eval_per_class_count: Evaluation set count of nb of tokens.
+        train_per_class_count: Training set word count.
+        eval_per_class_count: Evaluation set word count.
```
lindsaydbrin, line 579:
```suggestion
        train_per_class_count: Training set word count per class.
        eval_per_class_count: Evaluation set word count per class.
```
I thought maybe I was just being finicky, but `train_per_class_stats` includes "per class" too... again, lack of "per class" isn't your code, but while you're here, if you don't mind!

azimuth/modules/dataset_analysis/syntax_tagging.py
gabegma: I can add it if you prefer, but that is directly the definition of doc : "A Doc is a sequence of [Token](https://spacy.io/api/token) objects". From https://spacy.io/api/doc.

lindsaydbrin: Oh cool, I didn't know that - sorry I didn't look it up! Thanks. Ignore my comment, then!

lindsaydbrin (approved):
Looks great! Big fan of the syntax tagging refactoring. 😁 Thanks!

azimuth/modules/dataset_analysis/syntax_tagging.py
gabegma: Np! I didn't know either before refactoring :)

JosephMarinier (approved):
Beautiful! 👍

PR #376, opened 2023-01-12
Fix nullable generated types

JosephMarinier:
## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -9,7 +9,7 @@ from os.path import join as pjoin
 from typing import Any, Dict, List, Literal, Optional, TypeVar, Union
 
 import structlog
-from pydantic import BaseModel, BaseSettings, Extra, Field, root_validator, validator
+from pydantic import BaseSettings, Extra, Field, root_validator, validator
 
 from azimuth.types import AliasModel, DatasetColumn, SupportedModelContract
 from azimuth.utils.conversion import md5_hash
@@ -75,8 +75,23 @@ class AzimuthValidationError(Exception):
     pass
 
 
-# Mypy does not like a variable named kwargs.
-class CustomObject(BaseModel):  # type: ignore
+class AzimuthBaseSettings(BaseSettings):
+    class Config:
+        @staticmethod
+        def schema_extra(schema):
+            # For Union types, openapi-typescript understands oneOf better than anyOf.
+            for field in schema["properties"].values():
+                if field.get("type") == "array":
+                    field = field["items"]
+                if "anyOf" in field:
+                    field["oneOf"] = field.pop("anyOf")
+
+            # pydantic considers fields with default values to be optional, but when the API returns
+            # an object, all the default values are set, so the fields are always present.
+            schema["required"] = list(schema["properties"].keys())
+
+
+class CustomObject(AzimuthBaseSettings):
     class_name: str = Field(..., title="Class name to load.")
     args: List[Any] = []
     kwargs: Dict[str, Any] = {}
@@ -90,7 +105,7 @@ class CustomObject(BaseModel):  # type: ignore
 CustomObject.update_forward_refs()
 
 
-class MetricDefinition(CustomObject):  # type: ignore
+class MetricDefinition(CustomObject):
     additional_kwargs: Dict = Field(
         default_factory=dict,
         title="Additional kwargs",
@@ -98,7 +113,7 @@ class MetricDefinition(CustomObject):  # type: ignore
     )
 
 
-class TemperatureScaling(CustomObject, BaseSettings):
+class TemperatureScaling(CustomObject):
     class_name: Literal[
         "azimuth.utils.ml.postprocessing.TemperatureScaling"
     ] = "azimuth.utils.ml.postprocessing.TemperatureScaling"
@@ -114,7 +129,7 @@ class TemperatureScaling(CustomObject, BaseSettings):
         return values
 
 
-class ThresholdConfig(BaseSettings, CustomObject):
+class ThresholdConfig(CustomObject):
     class_name: Literal[
         "azimuth.utils.ml.postprocessing.Thresholding"
     ] = "azimuth.utils.ml.postprocessing.Thresholding"
@@ -129,7 +144,7 @@ class ThresholdConfig(BaseSettings, CustomObject):
         return values
 
 
-class PipelineDefinition(BaseSettings):
+class PipelineDefinition(AzimuthBaseSettings):
     name: str = Field(exclude_from_cache=True)
     model: CustomObject
     postprocessors: Optional[
@@ -160,7 +175,7 @@ class PipelineDefinition(BaseSettings):
         return None
 
 
-class DatasetWarningsOptions(BaseModel):
+class DatasetWarningsOptions(AzimuthBaseSettings):
     min_num_per_class: int = 20
     max_delta_class_imbalance: float = 0.5
     max_delta_representation: float = 0.05
@@ -168,7 +183,7 @@ class DatasetWarningsOptions(BaseModel):
     max_delta_std_words: float = 3.0
 
 
-class SyntaxOptions(BaseModel):
+class SyntaxOptions(AzimuthBaseSettings):
     short_sentence_max_word: int = 3
     long_sentence_min_word: int = 12
     spacy_model: SupportedSpacyModels = SupportedSpacyModels.use_default  # Language-based default
@@ -176,28 +191,29 @@ class SyntaxOptions(BaseModel):
     obj_tags: List[str] = []  # Language-based dynamic default value
 
 
-class NeutralTokenOptions(BaseModel):
+class NeutralTokenOptions(AzimuthBaseSettings):
     threshold: float = 1
     suffix_list: List[str] = []  # Language-based default value
     prefix_list: List[str] = []  # Language-based default value
 
 
-class PunctuationTestOptions(BaseModel):
+class PunctuationTestOptions(AzimuthBaseSettings):
     threshold: float = 1
 
 
-class FuzzyMatchingTestOptions(BaseModel):
+class FuzzyMatchingTestOptions(AzimuthBaseSettings):
     threshold: float = 1
 
 
-class TypoTestOptions(BaseModel):
+class TypoTestOptions(AzimuthBaseSettings):
     threshold: float = 1
     # Ex: if nb_typos_per_utterance = 2, this will create both tests with 1 typo and 2 typos per
     # utterance.
     nb_typos_per_utterance: int = 1
 
 
-class BehavioralTestingOptions(BaseModel):
+# TODO Change to AzimuthBaseSettings once the front end knows the default values.
+class BehavioralTestingOptions(BaseSettings):
     neutral_token: NeutralTokenOptions = NeutralTokenOptions()
     punctuation: PunctuationTestOptions = PunctuationTestOptions()
     fuzzy_matching: FuzzyMatchingTestOptions = FuzzyMatchingTestOptions()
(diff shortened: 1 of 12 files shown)


docs/docs/reference/custom-objects/index.md
```diff
@@ -25,16 +25,15 @@ from pydantic import BaseModel, Field
...
-    remote: Optional[str] = None  # (1)
+    remote: Optional[str] = None # (2)
 ```
 
-1. Absolute path to `class_name`.
-
-`class_name` is the name of the function or class that is located in `remote`. `args` and `kwargs`
+1. Name of the function or class that is located in `remote`. `args` and `kwargs`
 will be sent to the function/class.
+2. Relative path to class. `class_name` needs to be accessible from this path.
```
gabegma, line 36: Relative? Isn't that the absolute path?

docs/docs/reference/configuration/project.md
```diff
@@ -67,13 +67,14 @@ explained in [:material-link: Defining Dataset](../custom-objects/dataset.md).
     from pydantic import BaseModel, Field
 
     class CustomObject(BaseModel):
-        class_name: str = Field(..., title="Class name to load")
+        class_name: str
```
gabegma, line 70:
```suggestion
        class_name: str # (1)
```

gabegma (approved):
Cool!
Small details only.
For the env var - have we tested it to make sure that it works to omit them? I just want to make sure it's supported in our pydantic version.

JosephMarinier:
@gabegma

> For the env var - have we tested

Yes, I have tested locally with `SIMILARY=null`.

docs/docs/reference/custom-objects/index.md
JosephMarinier: Good question. I wanted to bring it to your attention, but I got distracted, sorry.
That's the `description` from the `Field` definition:
https://github.com/ServiceNow/azimuth/blob/94ec5d0a0e99d561b738860bf6ea719119202899/azimuth/config.py#L85
At least one of the two is wrong. 😝 I can dig to find which one if you're not sure.
In the docs, I see
```md
1. A package on Pypi
     1. Example: `remote: "torchvision"`
2. A folder with `requirements.txt`
3. A folder with `setup.py`
```

docs/docs/reference/configuration/project.md
JosephMarinier: Exactly! Thank you!

JosephMarinier: Also, as you have probably noticed by now, I am proposing to remove the pydantic `Field` details from the doc. There were not consistently represented in the docs (for example, we decided to not show the `exclude_from_cache` ), so instead of adding everything, I opted for removing everything. I think the point of this doc is to show the type, not how we deal with cache and front-end type generation. Also, the `description` or `title` is often reused for the ` # (1)` thing, so no need to duplicate it.
Does that make sense to you?

gabegma: Yeah totally! It's just that right now there is a legend for `(1) `on line 76.

docs/docs/reference/custom-objects/index.md
gabegma: In our configs, those are absolute paths. Either `/azimuth_shr` when on Docker, or we do m`ake local_configs` to have the absolute path on our computer. Interesting that we have relative in our code though! Perhaps I'm missing something.

docs/docs/reference/configuration/project.md
gabegma: Oh I wasn't seeing your 'exactly! Thank you!` for some reason. So you want to remove it?

JosephMarinier: Sorry, I meant "Exactly! Thank you!" as in "That's exactly what I was trying to do! I forgot the ` # (1)`. Good catch! Thank you!". I will add it. Thanks again!

tests/utils.py
```diff
@@ -34,7 +30,6 @@
 _SAMPLE_VOCAB_DIR = _SAMPLE_DATA_DIR / "distilbert-tokenizer-files"
 _CHECKPOINT_PATH = str(_SAMPLE_VOCAB_DIR)
 SIMPLE_PERTURBATION_TESTING_CONFIG = BehavioralTestingOptions(
-    neutral_suffix=NeutralTokenOptions(suffix_list=["pls", "thanks"], prefix_list=["pls", "hello"]),
```
JosephMarinier, line 37: This had no effect because of a typo. It should have been `neutral_token` instead of `neutral_suffix`. Since I changed `BehavioralTestingOptions`'s parent class from `BaseModel` to `BaseSettings`, in which extra arguments are forbidden by default instead of ignored, this was now crashing.
Alternatively, I could fix the typo to `neutral_token` and adapt the two tests that depend on it. Any preference?

webapp/src/pages/Settings.tsx
```diff
@@ -49,7 +49,11 @@ const CONFIG_SUB_FIELDS: Partial<AzimuthConfig> = {
     max_delta_std_words: 0,
     min_num_per_class: 0,
   },
-  similarity: { conflicting_neighbors_threshold: 0, no_close_threshold: 0 },
+  similarity: {
```
nandhinibsn, line 52: May I know why the default values for "conflicting_neighbors_threshold" and "no_close_threshold " are set to 0.9 and 0.5? 

Can we also set default values for temperature/threshold similarly?

webapp/src/pages/Settings.tsx
```diff
@@ -304,53 +313,40 @@ const Settings: React.FC = () => {
       {displaySectionTitle("General")}
       <FormGroup>
         <Columns columns={3}>
-          {resultingConfig.name !== undefined &&
-            displayReadonlyFields("name", resultingConfig.name)}
-          {resultingConfig.rejection_class !== undefined &&
+          {displayReadonlyFields("name", resultingConfig.name)}
+          {resultingConfig.rejection_class !== null &&
```
nandhinibsn, line 317: I hope the condition is for "undefined" here. We are supposed to display if it's null for the rejection class.

JosephMarinier: Sorry, you hope?
There is no need to check for `undefined` as `rejection_class` cannot be `undefined`.
Indeed, in your PR, I suggested handling the `null` case and showing it. My PR was mainly about the types and adapting the front end so it compiles.

webapp/src/pages/Settings.tsx
JosephMarinier: Those are the default values defined in the back end. I think it's not ideal to duplicate the default values, but I don't have an easy solution.

nandhinibsn: Yes right, it would be better if the back-end send us the default values.

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, it is based on the discussion on the null case scenario; we display fields if it's null. 

Right!. I understood the part where we don't worry about undefined from now on. I just wanted to clarify why did you hide the fields if it's null on this PR.

JosephMarinier: Because `displayReadonlyFields()` doesn't support `null`, and never did. In this PR, I fixed the type of `rejection_class` by adding `nullable=True`, so now I am forced to handle the `null` case.

JosephMarinier: Ok, I added support for `null`, in italic, like we discussed in your PR: df69ff3155ad4ae8f73df3806ffa4df02419229d

nandhinibsn: Changes looks good to me with italic font style if its null.

azimuth/config.py
```diff
@@ -75,29 +75,50 @@ class AzimuthValidationError(Exception):
     pass
 
 
-# Mypy does not like a variable named kwargs.
-class CustomObject(BaseModel):  # type: ignore
-    class_name: str = Field(..., title="Class name to load.")
+class AzimuthBaseSettings(BaseSettings):
```
gabegma, line 78: Should this be used for other objects in the future too? Like `ModuleOptions` maybe? Should we track it somewhere?

gabegma (commented):
This looks good to me, at least for the back end. I haven't reviewed the front end. I do have a few questions though :)

azimuth/config.py
```diff
@@ -159,44 +180,45 @@ def temperature(self) -> Optional[float]:
...
+class TypoTestOptions(AzimuthBaseSettings):
     threshold: float = 1
     # Ex: if nb_typos_per_utterance = 2, this will create both tests with 1 typo and 2 typos per
     # utterance.
     nb_typos_per_utterance: int = 1
 
 
-class BehavioralTestingOptions(BaseModel):
+# TODO Change to AzimuthBaseSettings once the front end knows the default values.
+class BehavioralTestingOptions(BaseSettings):
```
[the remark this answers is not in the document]
JosephMarinier: That's the only section of the config that is nullable (so the config page has a checkbox for it) AND that the front end doesn't know about its fields (they are not on the config page, unlike `uncertainty`). Therefore, if one selects the checkbox, we are currently sending an empty object `{}` to the back end, so it takes the default values. When we add the fields to the config page, and the front end knows the default values (so it can initialize the fields correctly), then we can make this change.
Do you think I should add more detail to the comment? Like this message?

azimuth/config.py
gabegma: Yes, thank you for the explanation!

azimuth/config.py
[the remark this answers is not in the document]
gabegma: I think the comment is fine. Isn't similarity analysis the same though?

webapp/src/types/generated/generatedTypes.ts
```diff
@@ -745,13 +811,25 @@ export interface components {
...
-      threshold?: number;
-      nb_typos_per_utterance?: number;
+      threshold: number;
+      nb_typos_per_utterance: number;
     };
+    /**
+     * Base class for settings, allowing values to be overridden by environment variables.
+     *
+     * This is useful in production for secrets you do not wish to save in code, it plays nicely with docker(-compose),
+     * Heroku and any 12 factor app design.
```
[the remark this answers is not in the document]
JosephMarinier: Yes. That happened to all the classes that were `BaseModel` instead of `BaseSettings`. The `BaseSettings` class had some docstrings, and it expectedly ends up here.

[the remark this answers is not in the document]
gabegma: Ah I see, it's because of that change. Thanks!

azimuth/config.py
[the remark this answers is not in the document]
JosephMarinier: Sorry, in my response, I said `uncertainty` where I meant `similarity`. I edited now. Does that answer your question?

webapp/src/types/generated/generatedTypes.ts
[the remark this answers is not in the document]
JosephMarinier: By the way, switching to `BaseSettings` allowed to remove two `# type: ignore` as well as catching this ["silent error"](https://github.com/ServiceNow/azimuth/pull/376/files/cbd9f29ddda1fc575e79462cd3ed1eda6c01778c#diff-9e08e8c7769db7b58089fd7659d75ca96df40bb7b5d50299d3a30abd54da9ad6L37).

azimuth/config.py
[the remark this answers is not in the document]
gabegma: Yes thanks!

tests/utils.py
JosephMarinier: Resolved in 7db40b04aab5e53ba11aafb0db2fa0b446258992

azimuth/config.py
JosephMarinier: Well, it wasn't long before I also needed the same fix elsewhere! 😝 You had a good instinct!
https://github.com/ServiceNow/azimuth/pull/384

PR #370, opened 2023-01-10
Add units to config fields

nandhinibsn:
Resolve #

## Description:
Added units to the following config sub fields:
Dataset warnings:
  max_delta_class_imbalance
  max_delta_mean_tokens
  max_delta_representation
  max_delta_std_tokens
  min_num_per_class
Syntax
  short_sentence_max_token
  long_sentence_min_token
Similarity:
  conflicting_neighbors_threshold
  no_close_threshold
Uncertainty:
  iterations
  high_epistemic_threshold
Postprocesors:
 temperature
 threshold

## Checklist:

You should check all boxes before the PR is ready. If a box does not apply, check it to acknowledge it.

* [x] **ISSUE NUMBER.** You linked the issue number (Ex: Resolve #XXX).
* [x] **PRE-COMMIT.** You ran pre-commit on all commits, or else, you
  ran `pre-commit run --all-files` at the end.
* [x] **USER CHANGES.** The changes are added to CHANGELOG.md and the documentation, if they impact
  our users.
* [x] **DEV CHANGES.**
    * Update the documentation if this PR changes how to develop/launch on the app.
    * Update the `README` files and our wiki for any big design decisions, if relevant.
    * Add unit tests, docstrings, typing and comments for complex sections.

PR #378, opened 2023-01-16
Refactor config units

JosephMarinier:
I recommend reviewing the commits one by one. The commit **_[Move utility functions outside Settings](https://github.com/ServiceNow/azimuth/pull/378/commits/5a3f7b23744114c36ddf296db694dd6d40913d9a) as much as possible to avoid unnecessarily redefining them on every render_** is just moving code around, without changing anything.

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -80,19 +80,104 @@ const Columns: React.FC<{ columns?: number }> = ({ columns = 1, children }) => (
   </Box>
 );
 
+const displaySectionTitle = (section: string) => (
+  <Typography variant="subtitle2" marginY={1.5}>
+    {section}
+  </Typography>
+);
+
 const KeyValuePairs: React.FC = ({ children }) => (
   <Box display="grid" gridTemplateColumns="max-content auto" columnGap={1}>
     {children}
   </Box>
 );
 
+const displayKeywordArguments = (name: string, kwargs: Record<string, any>) => (
+  <Box display="grid">
+    <Typography variant="caption">{name}</Typography>
+    <KeyValuePairs>
+      {Object.entries(kwargs).map(([field, value], index) => (
+        <React.Fragment key={index}>
+          <Typography variant="body2">{field}:</Typography>
+          <Tooltip title={value}>
+            <Typography
+              variant="body2"
+              whiteSpace="nowrap"
+              overflow="hidden"
+              textOverflow="ellipsis"
+            >
+              {value}
+            </Typography>
+          </Tooltip>
+        </React.Fragment>
+      ))}
+    </KeyValuePairs>
+  </Box>
+);
+
+const displayArgumentsList = (name: string, args: any[]) => (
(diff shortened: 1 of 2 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -86,13 +74,6 @@ const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof AzimuthConfig)[] = [
   "metrics",
 ];
 
-const ANALYSES_CUSTOMIZATION_IGNORE_FIELDS: string[] = [
-  "spacy_model",
-  "subj_tags",
-  "obj_tags",
-  "faiss_encoder",
-];
-
```
JosephMarinier, line 95: I am now assuming that all fields are declared in `FIELDS` and I check `field in FIELDS` instead of `!ANALYSES_CUSTOMIZATION_IGNORE_FIELDS.includes(field)`.

nandhinibsn (commented):
Looks more clean and reusable :) Good to go for merge.

webapp/src/pages/Settings.tsx
```diff
@@ -26,38 +27,31 @@ import { getConfigEndpoint, updateConfigEndpoint } from "services/api";
...
+  no_close_threshold: PERCENTAGE,
+  min_num_per_class: { ...INT, units: "samples" },
+  max_delta_class_imbalance: PERCENTAGE,
+  max_delta_representation: PERCENTAGE,
+  max_delta_mean_words: { ...FLOAT, units: "words" },
+  max_delta_std_words: { ...FLOAT, units: "words" },
+  short_sentence_max_word: { ...INT, units: "words" },
+  long_sentence_min_word: { ...INT, units: "words" },
+  temperature: FLOAT,
+  threshold: PERCENTAGE,
```
JosephMarinier, line 50: Hey @gabegma! Yesterday, I did some fixes on the units.
- the `iterations` field doesn't have units (it had been set to "tokens")
- the rest of the "tokens" fields became "words"
- the `temperature` field is NOT a percentage anymore, just a decimal value.

I think you also wanted to make `high_epistemic_threshold` not a percentage? It's cosine similarity, right? So -1 to 1... by steps of 0.1 to 0.01?

JosephMarinier: It's by bad, sorry, I think I went too quickly on my comment in https://github.com/ServiceNow/azimuth/pull/370#discussion_r1066561665
> The rest of the fields plus `high_epistemic_threshold`, the ones with `max: 1`, are ratios (like percentages) so they don't have units. Alternatively, we could consider displaying them in percentages (multiplying by 100 and showing "%"). What do you think @gabegma?

Do you see more of these fields for which percentages might not be appropriate?

JosephMarinier: We synced offline and we made those adjustments: https://github.com/ServiceNow/azimuth/pull/370/commits/e4ba44f7e1970e56faaa02c64e7c8c77714d02c8

PR #380, opened 2023-01-17
Refactor Syntactic Warnings

gabegma:
Part of #383

## Description:
First pass at refactoring dataset warnings to eventually allow them to run on just one dataset split.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -1,6 +1,7 @@
 # Copyright ServiceNow, Inc. 2021 – 2022
 # This source code is licensed under the Apache 2.0 license found in the LICENSE file
 # in the root directory of this source tree.
+from collections import defaultdict
 from typing import Dict, List, Set
 
 import numpy as np
@@ -205,12 +206,10 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
 
         """
         class_index = np.arange(dm.get_num_classes(labels_only=True))
-        value_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, float]] = {
-            split: {} for split in self.available_dataset_splits
-        }
-        value_per_label_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, np.ndarray]] = {
-            split: {} for split in self.available_dataset_splits
-        }
+        value_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, float]] = defaultdict(dict)
+        value_per_label_per_agg_per_split: Dict[
+            DatasetSplitName, Dict[Agg, np.ndarray]
+        ] = defaultdict(dict)
         hist_per_label_per_split = {}
 
         for split in self.available_dataset_splits:
@@ -248,14 +247,12 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
             Agg.mean: self.config.dataset_warnings.max_delta_mean_words,
             Agg.std: self.config.dataset_warnings.max_delta_std_words,
         }
-        for agg in thresholds_per_agg.keys():
+        for agg, threshold in thresholds_per_agg.items():
             divergence_per_label_per_agg[agg] = np.abs(
                 value_per_label_per_agg_per_split[DatasetSplitName.train][agg]
                 - value_per_label_per_agg_per_split[DatasetSplitName.eval][agg]
             )
-            alert_per_label_per_agg[agg] = (
-                divergence_per_label_per_agg[agg] > thresholds_per_agg[agg]
-            )
+            alert_per_label_per_agg[agg] = divergence_per_label_per_agg[agg] > threshold
         alert_per_label = np.any(
             [alert_per_label_per_agg[Agg.mean], alert_per_label_per_agg[Agg.std]], axis=0
         )
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -248,10 +248,8 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
         }
         for agg in thresholds_per_agg.keys():
             divergence_per_label_per_agg[agg] = np.abs(
-                (
-                    agg_per_label_per_split[DatasetSplitName.train][agg]
-                    - agg_per_label_per_split[DatasetSplitName.eval][agg]
-                )
+                agg_per_label_per_split[DatasetSplitName.train][agg]
+                - agg_per_label_per_split[DatasetSplitName.eval][agg]
             )
             alert_per_label_per_agg[agg] = (
                 divergence_per_label_per_agg[agg] > thresholds_per_agg[agg]
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -205,10 +205,10 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
 
         """
         class_index = np.arange(dm.get_num_classes(labels_only=True))
-        agg_per_split: Dict[DatasetSplitName, Dict[Agg, float]] = {
+        value_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, float]] = {
             split: {} for split in self.available_dataset_splits
         }
-        agg_per_label_per_split: Dict[DatasetSplitName, Dict[Agg, np.ndarray]] = {
+        value_per_label_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, np.ndarray]] = {
             split: {} for split in self.available_dataset_splits
         }
         hist_per_label_per_split = {}
@@ -223,13 +223,15 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
                 ]
             ).to_pandas()
 
-            agg_per_split[split][Agg.mean] = df["word_count"].mean()
-            agg_per_split[split][Agg.std] = df["word_count"].std()
+            value_per_agg_per_split[split][Agg.mean] = df["word_count"].mean()
+            value_per_agg_per_split[split][Agg.std] = df["word_count"].std()
 
             # Get mean and std word count per label
             stats_per_label = df.groupby("label").agg(["mean", "std"]).reindex(class_index)
             for agg in [Agg.mean, Agg.std]:
-                agg_per_label_per_split[split][agg] = stats_per_label["word_count"][agg].to_numpy()
+                value_per_label_per_agg_per_split[split][agg] = stats_per_label["word_count"][
+                    agg
+                ].to_numpy()
 
             # Get word count histogram per label. Columns are word counts and rows are labels.
             hist_per_label_per_split[split] = (
@@ -248,8 +250,8 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
         }
         for agg in thresholds_per_agg.keys():
             divergence_per_label_per_agg[agg] = np.abs(
-                agg_per_label_per_split[DatasetSplitName.train][agg]
-                - agg_per_label_per_split[DatasetSplitName.eval][agg]
+                value_per_label_per_agg_per_split[DatasetSplitName.train][agg]
+                - value_per_label_per_agg_per_split[DatasetSplitName.eval][agg]
             )
             alert_per_label_per_agg[agg] = (
                 divergence_per_label_per_agg[agg] > thresholds_per_agg[agg]
@@ -303,8 +305,8 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
                 ],
                 plots=word_count_plot(
                     hist_filled_per_split_per_label,
-                    agg_per_split,
-                    agg_per_label_per_split,
+                    value_per_agg_per_split,
+                    value_per_label_per_agg_per_split,
                     divergence_per_label_per_agg,
                     class_names,
                 ),
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -213,13 +213,9 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
         hist_per_label_per_split = {}
 
         for split in self.available_dataset_splits:
-            ds = self.get_dataset_split(split).remove_columns([])
+            ds = self.get_dataset_split(split)
             df = ds.remove_columns(
-                [
-                    column
-                    for column in ds.column_names
-                    if column not in [DatasetColumn.word_count, self.config.columns.label]
-                ]
+                list(set(ds.column_names) - {DatasetColumn.word_count, self.config.columns.label})
             ).to_pandas()
 
             value_per_agg_per_split[split][Agg.mean] = df["word_count"].mean()
--- a/azimuth/modules/dataset_analysis/dataset_warnings.py
+++ b/azimuth/modules/dataset_analysis/dataset_warnings.py
@@ -205,7 +205,7 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
             List of syntactic warnings.
 
         """
-        class_index = np.arange(dm.get_num_classes(labels_only=True))
+        class_indices = np.arange(dm.get_num_classes(labels_only=True))
         value_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, float]] = defaultdict(dict)
         value_per_label_per_agg_per_split: Dict[
             DatasetSplitName, Dict[Agg, np.ndarray]
@@ -226,8 +226,8 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
             value_per_agg_per_split[split][Agg.std] = df["word_count"].std()
 
             # Get mean and std word count per label
-            stats_per_label = df.groupby("label").agg(["mean", "std"]).reindex(class_index)
-            for agg in [Agg.mean, Agg.std]:
+            stats_per_label = df.groupby("label").agg(list(Agg)).reindex(class_indices)
+            for agg in Agg:
                 value_per_label_per_agg_per_split[split][agg] = stats_per_label["word_count"][
                     agg
                 ].to_numpy()
@@ -237,17 +237,17 @@ class DatasetWarningsModule(ComparisonModule[DatasetWarningConfig]):
                 df.groupby(["label", "word_count"])
                 .size()
                 .unstack(fill_value=0)
-                .reindex(class_index, fill_value=0)
+                .reindex(class_indices, fill_value=0)
             )
 
         # Compute divergence and alerts
-        divergence_per_label_per_agg = dict()
-        alert_per_label_per_agg = dict()
-        thresholds_per_agg = {
+        divergence_per_label_per_agg = {}
+        alert_per_label_per_agg = {}
+        threshold_per_agg = {
             Agg.mean: self.config.dataset_warnings.max_delta_mean_words,
             Agg.std: self.config.dataset_warnings.max_delta_std_words,
         }
-        for agg, threshold in thresholds_per_agg.items():
+        for agg, threshold in threshold_per_agg.items():
             divergence_per_label_per_agg[agg] = np.abs(
                 value_per_label_per_agg_per_split[DatasetSplitName.train][agg]
                 - value_per_label_per_agg_per_split[DatasetSplitName.eval][agg]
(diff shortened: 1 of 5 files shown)


azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -44,6 +43,8 @@ def compute_on_dataset_split(self) -> List[DatasetWarningsResponse]:  # type: ig
         train_mng = self.get_dataset_split_manager(DatasetSplitName.train)
         eval_mng = self.get_dataset_split_manager(DatasetSplitName.eval)
 
+        dm = train_mng or eval_mng
```
gabegma, line 46: I'll refactor this part once I refactor the generic warnings.

azimuth/plots/dataset_warnings.py
```diff
@@ -554,82 +512,72 @@ def create_histogram_mean_std(
...
-    )
-    train_per_class_stats, eval_per_class_stats = map(
-        lambda data: np.nan_to_num(np.array(data, dtype=float)),
-        (train_per_class_stats, eval_per_class_stats),
-    )
+    hist_per_split = dict()
+    for split in hist_per_label_per_split.keys():
+        # Sanitize values
+        agg_per_label = agg_per_label_per_split[split]
+        for agg in agg_per_label.keys():
```
JosephMarinier, line 560: Is the key an `agg` or a `label` in `agg_per_label`?

azimuth/plots/dataset_warnings.py
```diff
@@ -432,113 +434,69 @@ def class_representation(
...
         Plot for one label.
 
     """
     fig = go.Figure()
 
-    train_count_norm = train_count / max(sum(train_count), 1)
-    eval_count_norm = eval_count / max(sum(eval_count), 1)
+    hist_normalized = dict()
+    for split in hist_per_split.keys():
+        hist_normalized[split] = hist_per_split[split] / max(sum(hist_per_split[split]), 1)
```
JosephMarinier, line 457: Also using a dict comprehension
```suggestion
    hist_normalized = {split: hist / max(sum(hist), 1) for split, hist in hist_per_split.items()}
```

azimuth/plots/dataset_warnings.py
```diff
@@ -432,113 +434,69 @@ def class_representation(
...
+    max_y = max([value.max() for value in hist_normalized.values()])
+
+    y_per_split = {DatasetSplitName.eval: 1.14, DatasetSplitName.train: 1.07}
+    opacity_per_split = {DatasetSplitName.eval: 1, DatasetSplitName.train: 0.5}
+    # reversed because the color overlay looks better when train is first
+    for split in reversed(hist_normalized.keys()):
+        if hist_normalized[split].max() != 0:
+            fig.add_bar(
+                x=list(range(1, len(hist_normalized[split]) + 1)),
+                y=hist_normalized[split],
```
JosephMarinier, line 469:
```suggestion
    for split, hist in reversed(hist_normalized.items()):
        if hist.max() != 0:
            fig.add_bar(
                x=list(range(1, len(hist) + 1)),
                y=hist,
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +194,121 @@ def get_general_warnings(
...
+        for agg in thresholds_per_agg.keys():
+            divergence_per_label_per_agg[agg] = np.abs(
+                (
+                    agg_per_label_per_split[DatasetSplitName.train][agg]
+                    - agg_per_label_per_split[DatasetSplitName.eval][agg]
+                )
+            )
+            alert_per_label_per_agg[agg] = (
+                divergence_per_label_per_agg[agg] > thresholds_per_agg[agg]
+            )
```
JosephMarinier, line 258: Also, here, it seems there are some redundant parentheses.
```suggestion
        for agg, threshold in thresholds_per_agg.items():
            divergence_per_label_per_agg[agg] = np.abs(
                agg_per_label_per_split[DatasetSplitName.train][agg]
                - agg_per_label_per_split[DatasetSplitName.eval][agg]
            )
            alert_per_label_per_agg[agg] = divergence_per_label_per_agg[agg] > threshold
```

JosephMarinier (commented):
I think we can simplify a couple of loops by using of `.items()` instead of `.keys()`. In fact, it's quite rare that we need `.keys()`, we almost always also need the values.

azimuth/plots/dataset_warnings.py
```diff
@@ -554,82 +512,72 @@ def create_histogram_mean_std(
...
-    train_per_class_stats, eval_per_class_stats = map(
-        lambda data: np.nan_to_num(np.array(data, dtype=float)),
-        (train_per_class_stats, eval_per_class_stats),
-    )
+    hist_per_split = dict()
+    for split in hist_per_label_per_split.keys():
+        # Sanitize values
+        agg_per_label = agg_per_label_per_split[split]
+        for agg in agg_per_label.keys():
+            agg_per_label[agg] = np.nan_to_num(agg_per_label[agg])
```
JosephMarinier, line 561: Here, we are modifying the `agg_per_label_per_split` dict that the function (`word_count_plot()`) receives. It is good practice to have the function do only one thing (in this case, return a plot), and not have any "side effects" like that. We could do a copy instead, say `sanitized_hist_per_label_per_split`. That avoids potential confusion caused to the caller, who sees its parameters change unexpectedly. That being said, it might be pointless in this context. You tell me.

azimuth/plots/dataset_warnings.py
```diff
@@ -554,82 +512,72 @@ def create_histogram_mean_std(
...
+
+        hist_per_split[split] = hist_per_label_per_split[split].sum(axis=0)
```
JosephMarinier, line 563: In a similar fashion as before, instead of creating an empty `dict()` and then adding stuff to it in a loop, we can do all that in a dict comprehension:
```diff
-    hist_per_split = dict()
```
```suggestion
        hist_per_split = {split: h.sum(axis=0) for split, h in hist_per_label_per_split.items()}
```

azimuth/utils/plots.py
```diff
@@ -46,6 +48,14 @@ class Colors(str, Enum):
     Text = "#2a3d40"
 
 
+COLORS_PER_SPLIT = {DatasetSplitName.eval: Colors.DataViz1, DatasetSplitName.train: Colors.DataViz2}
+
+SPLIT_PRETTY_NAMES = {
```
JosephMarinier, line 53: Oh I like the "arrangement" you used for `SPLIT_PRETTY_NAMES`. I think it could be applied to `SPLIT_COLORS` too. I think we could even go super explicit with `DATASET_SPLIT_COLORS`, and `DATASET_SPLIT_PRETTY_NAMES` (I have that exact same name in the front end), especially since "split" could be a verb without the context of "dataset split". What do you think?
```suggestion
DATASET_SPLIT_COLORS = {DatasetSplitName.eval: Colors.DataViz1, DatasetSplitName.train: Colors.DataViz2}

DATASET_SPLIT_PRETTY_NAMES = {
```

azimuth/plots/dataset_warnings.py
```diff
@@ -554,82 +512,72 @@ def create_histogram_mean_std(
...
-            threshold_std,
-            train_per_class_stats[label_id],
-            eval_per_class_stats[label_id],
-            divergence_mean[label_id],
-            divergence_std[label_id],
-            class_names[label_id],
+            {split: value[label_id] for split, value in hist_per_label_per_split.items()},
+            {
+                split: {agg: value[label_id] for agg, value in per_split_value.items()}
+                for split, per_split_value in agg_per_label_per_split.items()
```
JosephMarinier, line 578: The way I read the name `agg_per_label_per_split`, I would expect to loop on it in the following fashion. I'm not sure I understand what an `agg` is. Can you help me understand?
```suggestion
                split: {label: agg[label_id] for label, agg in agg_per_label.items()}
                for split, agg_per_label in agg_per_label_per_split.items()
```

azimuth/plots/dataset_warnings.py
```diff
@@ -432,113 +434,69 @@ def class_representation(
...
-        eval_count: Histogram values for the selected label in the evaluation set.
-        threshold_mean: Alert value.
-        threshold_std: Alert value.
-        train_stats: Mean and std dev of the train.
-        eval_stats: Mean and std dev of the evaluation set.
-        divergence_mean: Difference in the mean. None means one distribution is empty.
-        divergence_std: Difference in the std dev. None means one distribution is empty.
-        label_name: Name of the label, when sending 'None', the traces are generated for 'all'.
+        hist_per_split: Histogram values for the selected label per split.
+        agg_per_split: Mean and std dev for the selected label per split.
```
JosephMarinier, line 445: So isn't that an `agg_per_label_per_split`?
```suggestion
        agg_per_label_per_split: Mean and std dev for the selected label per split.
```

azimuth/plots/dataset_warnings.py
gabegma: Right.. I'm not used to dict comprehension, and I forgot I could use them that way..  🤦‍♀️

azimuth/plots/dataset_warnings.py
gabegma: Right, this was stupid. I had a fixation on `keys()` for some reason.

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: Totally for the first one, but the second ones are added by the pre-commits.

azimuth/plots/dataset_warnings.py
gabegma: I'll create a new variable!

azimuth/modules/dataset_analysis/dataset_warnings.py
JosephMarinier: Of course, but by changing `thresholds_per_agg[agg]` to `threshold` (thanks to `.items()`), it now fits on one line.

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
+                    if column not in [DatasetColumn.word_count, self.config.columns.label]
+                ]
+            ).to_pandas()
+
+            value_per_agg_per_split[split][Agg.mean] = df["word_count"].mean()
+            value_per_agg_per_split[split][Agg.std] = df["word_count"].std()
+
+            # Get mean and std word count per label
+            stats_per_label = df.groupby("label").agg(["mean", "std"]).reindex(class_index)
+            for agg in [Agg.mean, Agg.std]:
```
JosephMarinier, line 230: Here, you could loop on `Agg`, instead of repeating the values.
```suggestion
            stats_per_label = df.groupby("label").agg(list(Agg)).reindex(class_index)
            for agg in Agg:
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
-        eval_count = fill_at_missing_indices(eval_count, eval_missing_cols)
-        # Add to mean too.
-        train_desc = fill_at_missing_indices(
-            train_desc, train_missing_cols, fill_value=float("nan")
+        # Fill columns so both splits have the same values from 1 to N.
+        N = max(
+            [
+                max(hist_per_label_per_split[split].columns)
+                for split in self.available_dataset_splits
+            ]
```
JosephMarinier, line 265: You shouldn't need the list comprehension here, `max()` accepts a generator directly.
```suggestion
            max(hist_per_label_per_split[split].columns) for split in self.available_dataset_splits
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
-        # Get counts, this will have N columns with count and LABEL rows.
-        train_count = (
-            train_df.groupby(["label", DatasetColumn.word_count]).size().unstack(fill_value=0)
-        )
-        eval_count = (
-            eval_df.groupby(["label", DatasetColumn.word_count]).size().unstack(fill_value=0)
+        # Compute divergence and alerts
+        divergence_per_label_per_agg = dict()
+        alert_per_label_per_agg = dict()
+        thresholds_per_agg = {
```
JosephMarinier, line 246: One threshold per agg, right?
```suggestion
        threshold_per_agg = {
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
 
-        # Get counts, this will have N columns with count and LABEL rows.
-        train_count = (
-            train_df.groupby(["label", DatasetColumn.word_count]).size().unstack(fill_value=0)
-        )
-        eval_count = (
-            eval_df.groupby(["label", DatasetColumn.word_count]).size().unstack(fill_value=0)
+        # Compute divergence and alerts
+        divergence_per_label_per_agg = dict()
+        alert_per_label_per_agg = dict()
```
JosephMarinier, line 245: Nitpick, but for consistency, it seems that we use `{}` more frequently than the `dict()` in our repo.
```suggestion
        divergence_per_label_per_agg = {}
        alert_per_label_per_agg = {}
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
-        train_count_filled = np.zeros([len(train_count), N], dtype=int)
-        test_count_filled = np.zeros([len(eval_count), N], dtype=int)
-
-        # Fill with token count (-1 because columns start at 1)
-        train_count_filled[:, np.array(train_count.columns) - 1] = train_count.to_numpy()
-        test_count_filled[:, np.array(eval_count.columns) - 1] = eval_count.to_numpy()
-
+        hist_filled_per_split_per_label = dict()
+        for split in self.available_dataset_splits:
+            hist_per_label = hist_per_label_per_split[split]
```
JosephMarinier, line 269: One line!
```suggestion
        for split, hist_per_label in hist_per_label_per_split.items():
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
-        divergence_mean = divergence["mean"].to_numpy()
-        divergence_std = divergence["std"].to_numpy()
-
-        # Make alerts from values.
-        max_delta_mean_words = int(self.config.dataset_warnings.max_delta_mean_words)
-        max_delta_std_words = int(self.config.dataset_warnings.max_delta_std_words)
-        alert_norm_mean = divergence_mean > max_delta_mean_words
-        alert_norm_std = divergence_std > max_delta_std_words
-        alert_norm = np.any([alert_norm_mean, alert_norm_std], axis=0)
+        class_index = np.arange(dm.get_num_classes(labels_only=True))
```
JosephMarinier, line 208: Since it's a list, should we make the name plural? `class_indices`? Just `classes`?

azimuth/plots/dataset_warnings.py
```diff
@@ -432,113 +434,69 @@ def class_representation(
...
+                marker=dict(color=DATASET_SPLIT_COLORS[split]),
+            )
 
-    at_least_one_distribution = min(train_count_norm.max(), eval_count_norm.max()) > 0
+    at_least_one_distribution = min([value.max() for value in hist_normalized.values()]) > 0
 
-    if divergence_mean and divergence_std and at_least_one_distribution:
+    if divergence_per_agg and at_least_one_distribution:
         fig.add_annotation(
             dict(
```
JosephMarinier, line 487: In the two places we call `add_annotation`, we can remove the `dict(...)` and unindent the arguments.
```suggestion
```

azimuth/plots/dataset_warnings.py
```diff
@@ -432,113 +434,69 @@ def class_representation(
...
+                name=f"{split}_mean_std",
+                error_x=dict(
+                    type="constant", value=np.round(value_per_agg_per_split[split][Agg.std], 2)
+                ),
+                hoverinfo="x",
+                marker=dict(color=DATASET_SPLIT_COLORS[split]),
+            )
 
-    at_least_one_distribution = min(train_count_norm.max(), eval_count_norm.max()) > 0
+    at_least_one_distribution = min([value.max() for value in hist_normalized.values()]) > 0
```
JosephMarinier, line 483: In practice, you wouldn't see a real difference in performance, but here we could use `any` and move the check `> 0` inside the generator, so that it breaks our of the loop as soon as it finds one, instead of generating the whole list every time.
```suggestion
    at_least_one_distribution = any(value > 0 for h in hist_normalized.values() for value in h)
```

JosephMarinier (approved):
What a breath of fresh air! Good job!
I only had some small comments about cosmetic changes, for readability. The architecture and the logic look perfect! 👍

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: Right thanks! I wanted to verify how to loop on an `Enum`, but forgot to check.

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: Oh awesome!

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,117 @@ def get_general_warnings(
...
                         name=class_names[i],
-                        alert=alert_norm[i],
+                        alert=alert_per_label[i],
                         data=[
                             DatasetDistributionComparisonValue(
-                                value=divergence_mean[i], alert=alert_norm_mean[i]
+                                value=divergence_per_label_per_agg[Agg.mean][i],
+                                alert=alert_per_label_per_agg[Agg.mean][i],
                             ),
                             DatasetDistributionComparisonValue(
```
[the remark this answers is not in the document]
gabegma: Oh good catch!

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: Yeah I like `class_indices`, I was also thinking of changing it to that.

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: I fixed 3 others like that.

azimuth/plots/dataset_warnings.py
```diff
@@ -432,99 +434,83 @@ def class_representation(
...
 
     Returns:
         Plot for one label.
 
     """
     fig = go.Figure()
 
-    train_count_norm = train_count / max(sum(train_count), 1)
-    eval_count_norm = eval_count / max(sum(eval_count), 1)
+    hist_normalized = {split: hist / max(sum(hist), 1) for split, hist in hist_per_split.items()}
```
JosephMarinier, line 455: I have an idea! We can filter out the empty histograms from `hist_normalized`. That also solves the `max(..., 1)` hack to avoid divisions by 0.
```suggestion
    hist_normalized = {split: hist / sum(hist) for split, hist in hist_per_split.items() if any(hist)}
```
Then, there is no need to for this `if` inside the loop anymore:
```diff
-        if hist.max() != 0:
```
And the two checks become easy:
```diff
-    two_distributions = all(any(value > 0 for value in h) for h in hist_normalized.values())
- 
-    if divergence_per_agg and two_distributions:
+    if divergence_per_agg and len(hist_normalized) == 2:
```
```diff
-    empty_distribution = max(value.max() for value in hist_normalized.values()) == 0
-    if empty_distribution:
+    if len(hist_normalized) == 0:
```
I think there is only one more thing to adjust for this to work; avoiding the error in `max()` if there are no non-empty histograms:
```diff
     # Helps position the error bars.
-    max_y = max(value.max() for value in hist_normalized.values())
+    max_y = max((value.max() for value in hist_normalized.values()), default=0)
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,110 @@ def get_general_warnings(
...
-        alert_norm = np.any([alert_norm_mean, alert_norm_std], axis=0)
+        class_indices = np.arange(dm.get_num_classes(labels_only=True))
+        value_per_agg_per_split: Dict[DatasetSplitName, Dict[Agg, float]] = defaultdict(dict)
+        value_per_label_per_agg_per_split: Dict[
+            DatasetSplitName, Dict[Agg, np.ndarray]
+        ] = defaultdict(dict)
+        hist_per_label_per_split = {}
+
+        for split in self.available_dataset_splits:
+            ds = self.get_dataset_split(split).remove_columns([])
```
JosephMarinier, line 216: The following `remove_columns()` already returns
> A copy of the dataset object without the columns to remove.

So it seems unnecessary to call `remove_columns()` twice. Or am I missing something?
```suggestion
            ds = self.get_dataset_split(split)
```

azimuth/modules/dataset_analysis/dataset_warnings.py
```diff
@@ -193,135 +195,110 @@ def get_general_warnings(
...
+            df = ds.remove_columns(
+                [
+                    column
+                    for column in ds.column_names
+                    if column not in [DatasetColumn.word_count, self.config.columns.label]
+                ]
```
JosephMarinier, line 222: No big difference, but you might like that:
```suggestion
                list(set(ds.column_names) - {DatasetColumn.word_count, self.config.columns.label})
```

azimuth/modules/dataset_analysis/dataset_warnings.py
gabegma: Oh yes, this was a mistake, good catch.

azimuth/plots/dataset_warnings.py
gabegma: Good idea

PR #388, opened 2023-01-24
Remove generic tokenizer and support multiple languages for the word cloud.

gabegma:
## Description:
I investigated if the generic tokenizer was still useful. It turns out it wasn't really, so I made the following changes:
- There are no more dummy saliencies (equal to 0) that are computed when saliency is not available. It was counter-productive to compute them since they were not used in the UI, and for the top words, the words were converted to sub-tokens, and then back to words. 
- Because of that, I added saliency as an optional startup task.
- Furthermore, I needed to refactor the top words code, so it no longer calls the dummy saliency method when saliency is not available. This turned out to be an opportunity: I now use the Spacy model (which, as we know, can be defined per language), which already has its own list of punctuation chars (`is_punct`) and stop words (`is_stop`). The chars are a bit different from the ones we had, but as you can see in the tests, the differences seem ok to me. Plus, now it will work for french :).
   - However, I needed to add a char replacement for this to work. The English Spacy model did not recognize "`" as a punctuation sign. I believe this also caused problems that we weren't seeing when counting tokens or performing POS tagging. When replacing it by "'", this slightly changed the values in the dataset warnings (token count).  
       - Example from sst2 with the  "`":
       >  designed to provide a mix of smiles and tears , `` crossroads '' instead provokes a handful of unintentional howlers and numerous yawns .
  - I needed to change the scope of `TopWordsModule` from `ModelContractConfig` to `AzimuthConfig`, since it now relies on both the model or the syntax config, depending if saliency is available. `AzimuthConfig` is a bit too broad, but since this module computes fast, I think that it is ok. 
 
## Checklist:

  our users.
* [x] **DEV CHANGES.**


azimuth/utils/utterance.py
```diff
@@ -13,5 +13,5 @@ def clean_utterance(utterance: str) -> str:
         utterance_clean: Cleaned utterance.
 
     """
-    utterance_clean = utterance.replace("’", "'").lower()  # Quote->apostrophe (unicode 8217 -> 39)
+    utterance_clean = utterance.replace("’", "'").replace("`", "'").lower()
```
JosephMarinier, line 16: I think it's fine to repeat `replace(..., "'")` twice, but if it bothers you, or if you suspect we might need to add more, we could replace them all at one call using a regular expression.
```py
import re

re_irregular_apostrophes = re.compile("[’`]")
```
```suggestion
    utterance_clean = re_irregular_apostrophes.sub("'", utterance).lower()
```

azimuth/modules/word_analysis/top_words.py
```diff
@@ -19,21 +19,20 @@
...
 from azimuth.utils.dataset_operations import get_predictions_from_ds
-from azimuth.utils.ml.third_parties.stop_words import STOP_WORDS
 from azimuth.utils.project import saliency_available
+from azimuth.utils.utterance import clean_utterance
 
 MIN_SALIENCY = 0.01
 
 
-class TopWordsModule(FilterableModule[ModelContractConfig]):
+class TopWordsModule(FilterableModule[AzimuthConfig]):
```
JosephMarinier, line 28: Do you think we should comment on why this scope is as broad as `AzimuthConfig`? You already have something...
> relies on both the model or the syntax config, depending if saliency is available. `AzimuthConfig` is a bit too broad, but since this module computes fast, I think that it is ok.

azimuth/utils/utterance.py
JosephMarinier: Also, the name `re_irregular_apostrophes` serves as a comment. Maybe `re_unusual_apostrophes` is better? Or `re_atypical_apostrophes`? 🤷‍♂️

azimuth/modules/word_analysis/top_words.py
gabegma: Good idea!

azimuth/utils/utterance.py
gabegma: Oh cool!! I really need to learn about regular expressions..

tests/test_modules/test_word_analysis/test_top_words.py
```diff
@@ -55,34 +55,25 @@ def test_top_words_with_saliency(simple_text_config, apply_mocked_startup_task,
...
-    for word in ["i", "is", "the", "."]:
-        assert word not in top_word_all
+    assert all(
+        word in top_word_all for word in ["phone", "potato", "pizza", "forest", "park", "turning"]
+    )
+    assert all(word not in top_word_all for word in ["to", "the", "not", "with", "on"])
+    assert all(word in top_word_right for word in ["phone", "potato", "pizza", "forest"])
+    assert all(word not in top_word_right for word in ["turning", "park"])
+    assert all(word in top_word_errors for word in ["phone", "turning", "park"])
+    assert all(word not in top_word_errors for word in ["potato", "pizza", "forest"])
```
JosephMarinier, line 79: Here, you could leverage some sets to make the assertions more readable.
```suggestion
    assert {"phone", "turning", "park"}.issubset(top_word_errors)
    assert {"potato", "pizza", "forest"}.isdisjoint(top_word_errors)
```

If you change the comprehension lists to sets like so
```py
    top_word_all = {top_words_result.word for top_words_result in json_output.all}
    top_word_right = {top_words_result.word for top_words_result in json_output.right}
    top_word_errors = {top_words_result.word for top_words_result in json_output.errors}
```
you could even use `<=` (instead of `issubset()`) or `==` (if it applies here).
```suggestion
    assert {"phone", "turning", "park"} <= top_word_errors
    assert {"potato", "pizza", "forest"}.isdisjoint(top_word_errors)
```

azimuth/modules/word_analysis/top_words.py
```diff
@@ -102,49 +101,57 @@ def compute_on_dataset_split(self) -> List[TopWordsResponse]:  # type: ignore
...
+            for idx, words in important_words_per_idx.items()
+            for word in words
+            if not is_error[idx]
+        ]
+        important_words_errors = [
+            word
+            for idx, words in important_words_per_idx.items()
+            for word in words
+            if is_error[idx]
+        ]
```
JosephMarinier, line 154: In this case, I didn't hate the single loop that populates the 3 lists. It seems less repetitive. That said, I don't have a string preference. What do you think?
```suggestion
        important_words_all = []
        important_words_right = []
        important_words_errors = []
        for idx, words in important_words_per_idx.items():
            important_words_all.extend(words)
            if is_error[idx]:
                important_words_errors.extend(words)
            else:
                important_words_right.extend(words)
```

JosephMarinier (approved):
Cool cool cool! 😚👌 I only have some small comments.

tests/test_modules/test_word_analysis/test_top_words.py
gabegma: Ok cool! `sets` are still a bit of a mystery to me.

azimuth/modules/word_analysis/top_words.py
gabegma: Yeah, you are right. list comprehensions are not always the best when you are looping over the same items.

azimuth/utils/utterance.py
```diff
@@ -13,5 +14,6 @@ def clean_utterance(utterance: str) -> str:
         utterance_clean: Cleaned utterance.
 
     """
-    utterance_clean = utterance.replace("’", "'").lower()  # Quote->apostrophe (unicode 8217 -> 39)
+    re_irregular_apostrophes = re.compile("[’`]")
```
JosephMarinier, line 17: Oh, I wasn't clear. You can "compile" the regular expression at the top of the file, say right after the imports, to avoid compiling it every time this function is called.

azimuth/modules/base_classes/artifact_manager.py
```diff
@@ -117,11 +116,6 @@ def get_model(self, config: AzimuthConfig, pipeline_idx: int):
 
         return self.models_mapping[config_key][pipeline_idx]
 
-    def get_tokenizer(self):
```
lindsaydbrin, line 120: Hooray, nice job finishing the work to make this unnecessary! :)

azimuth/modules/word_analysis/top_words.py
```diff
@@ -19,21 +19,25 @@
...
 
 
-class TopWordsModule(FilterableModule[ModelContractConfig]):
-    """Returns the most important words in terms of their saliency value or frequency."""
+class TopWordsModule(FilterableModule[AzimuthConfig]):
+    """Returns the most important words in terms of their saliency value or frequency.
+
+    Note: The config scope is AzimuthConfig because the module relies on both the pipeline or the
+    syntax config, depending if saliency is available. AzimuthConfig is a bit too broad, but it
+    should not be a problem since this module computes fast.
```
lindsaydbrin, line 33: So minor and picky, feel free to take or leave. Super helpful docstring note, though.
```suggestion
    Note: The config scope is AzimuthConfig because the module relies on both the pipeline or the
    syntax config, depending on whether saliency is available. AzimuthConfig is a bit too broad,
    but it should not be a problem since this module computes quickly.
```

azimuth/modules/word_analysis/top_words.py
```diff
@@ -102,44 +106,48 @@ def compute_on_dataset_split(self) -> List[TopWordsResponse]:  # type: ignore
...
-        # Saliencies will be 0 if saliency maps are not available.
-        words_saliencies = self.get_words_saliencies(self.get_indices())
-
-        important_words_all = []
-        important_words_right = []
-        important_words_errors = []
+        important_words_per_idx: Dict[int, List[str]] = {}
+        if importance_criteria == TopWordsImportanceCriteria.salient:
+            words_saliencies = self.get_words_saliencies(self.get_indices())
+            tokenizer = self.get_model().tokenizer
```
lindsaydbrin, line 112: 💯

azimuth/modules/base_classes/artifact_manager.py
gabegma: Exactly!

gabegma:
@JosephMarinier @lindsaydbrin actually, I realized that I could create a specific config scope for TopWords. See my last commit. Does that make sense to you?

azimuth/modules/word_analysis/top_words.py
```diff
@@ -102,44 +101,48 @@ def compute_on_dataset_split(self) -> List[TopWordsResponse]:  # type: ignore
...
+                        self.mod_options.th_importance * max(record.saliency), MIN_SALIENCY
+                    )
+                    important_words_per_idx[idx] = [
+                        word
+                        for word, _ in filter(
+                            lambda s: s[1] > importance_saliency, zip(words, saliencies)
+                        )
+                    ]
+        # If saliency is not available, we proxy important words as any word that is neither
+        # punctuation or a stop word.
```
lindsaydbrin, line 128:
```suggestion
        # punctuation nor a stop word.
```

lindsaydbrin:
> @JosephMarinier @lindsaydbrin actually, I realized that I could create a specific config scope for TopWords. See my last commit. Does that make sense to you?

If it works, seems like a more direct approach! It seems like if we do this a lot, we could have a tangled mess of custom config pairings, but we can deal with that if it happens. But anyway, generally, I'd defer to @JosephMarinier on this.

docs/docs/user-guide/exploration-space/prediction-overview.md
```diff
@@ -58,5 +58,4 @@ time will clear that search filter, so does the ✗ in the `Search utterances` f
     saliency values is greater than 60% of the largest saliency value in the utterance.
 
     If saliency maps are not available, the word clouds only show the most **frequent words** which
-    are not part of a pre-defined list of **stop words**
-    (from [nltk.corpus](https://github.com/nltk/nltk)).
+    are neither punctuation or stop words, according to the spacy model defined in the config.
```
lindsaydbrin, line 61:
```suggestion
    are neither punctuation nor stop words, according to the spacy model defined in the config.
```
```suggestion
    are neither punctuation or stop words, according to the spacy model defined in the config.
```
```suggestion
    are neither punctuation or stop words, according to the spacy model defined in the config.
```

azimuth/modules/model_contracts/hf_text_classification.py
```diff
@@ -116,7 +116,7 @@ def saliency(self, batch: Dataset) -> List[SaliencyResponse]:
 
         """
         if self.saliency_layer is None:
-            return self.empty_saliency_from_batch(batch)
+            raise ValueError("This method should not be called when saliency_layer is not defined.")
```
lindsaydbrin, line 119: I'm thinking about who this message is meant for. When would you get here? If `config.saliency_layer` is `None`, then the saliency module wouldn't be called in the startup tasks, right? So is this for development, or can it ever show up for a user? (I think I'm wondering about wording in that context, but also trying to understand more generally.)

lindsaydbrin (approved):
Looks great - thanks for taking care of this, and for such helpful `Description` context! Some small edits on comments, one small question mostly for my understanding, and I'll let Joseph comment on the config scope change. It was already approved, but here's another anyhow. 😆

azimuth/modules/model_contracts/hf_text_classification.py
gabegma: It could probably be deleted. It was already there so I left it. It's meant for developers. If from a module, a developer calls the saliency method without checking if it's available, it would simply crash later in this code. This is probably fine too.

lindsaydbrin: Oh, I think that's reasonable, then! I guess it's kind of unnecessary lines of code, but I think it's not so bad to give information that will help a future dev (or self) who might run into trouble.

azimuth/config.py
```diff
@@ -393,13 +393,17 @@ class SyntaxConfig(CommonFieldsConfig):
     syntax: SyntaxOptions = SyntaxOptions()
 
 
+class TopWordsConfig(SyntaxConfig, ModelContractConfig):
```
JosephMarinier, line 396: Oh cool! Very good idea!
Indeed, it shouldn't matter if multiple sub-configs all inherit from the same classes, in this case, `ModelContractConfig`. Let's leverage that flexibility! 👍

PR #392, opened 2023-01-25
Enforce mypy in pre-commit

gabegma:
## Description:
`mypy` was never enforced in the pre-commit because, at a certain time, we couldn't put the errors to 0 easily. This is no longer the case, and I don't think we should be able to commit with `mypy` errors. 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,6 +63,6 @@ repos:
         name: mypy
         language: system
         verbose: true
-        entry: bash -c 'poetry run mypy "azimuth" || true' --
+        entry: bash -c 'poetry run mypy "azimuth"'
         files: azimuth
         pass_filenames: false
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -63,6 +63,6 @@ repos:
         name: mypy
         language: system
         verbose: true
-        entry: bash -c 'poetry run mypy "azimuth" || true' --
+        entry: bash -c 'poetry run mypy "azimuth"'
```
JosephMarinier, line 66: Without the `||`, we don't need `bash` anymore.
```suggestion
        entry: poetry run mypy azimuth
```

JosephMarinier (approved):
Oh! Very good catch! Thank you!
I just have a small comment to clean it up a bit.

PR #396, opened 2023-01-25
Add default config to config page UI

nandhinibsn:
Resolve #

## Description:

This PR is to call the default config API from the back end to display the default config subfields in scenarios like postprocessor or similarity being null
## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -227,7 +227,8 @@ const Settings: React.FC = () => {
                 {
                   ...pipeline,
                   postprocessors: checked
-                    ? config.pipelines![pipelineIndex].postprocessors ?? []
+                    ? config.pipelines![pipelineIndex].postprocessors ??
+                      defaultConfig.pipelines![0].postprocessors
                     : null,
                 },
                 ...resultingConfig.pipelines!.slice(pipelineIndex + 1),
@@ -251,6 +252,7 @@ const Settings: React.FC = () => {
     <NumberField
       label={field}
       value={value}
+      disabled={!resultingConfig.pipelines![pipelineIndex].postprocessors}
       onChange={(newValue) =>
         setPartialConfig({
           ...partialConfig,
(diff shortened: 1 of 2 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -179,15 +174,16 @@ const NumberField: React.FC<
...
   const [updateConfig] = updateConfigEndpoint.useMutation();
 
   const [partialConfig, setPartialConfig] = React.useState<
     Partial<AzimuthConfig>
   >({});
 
-  // If config was undefined, PipelineCheck would not even render the page.
-  if (config === undefined) return null;
+  // If config or defaultConfig was undefined, PipelineCheck would not even render the page.
+  if (config === undefined || defaultConfig === undefined) return null;
```
JosephMarinier, line 186: That's not true for `defaultConfig`. We should revert that change and handle loading/errors with `defaultConfig`. We'll need `isLoading` and `error` from
```diff
-  const { data: defaultConfig } = getDefaultConfigEndpoint.useQuery({ jobId });
+  const { data: defaultConfig, isLoading, error } = getDefaultConfigEndpoint.useQuery({ jobId });
```
and
```suggestion
  // If config was undefined, PipelineCheck would not even render the page.
  if (config === undefined) return null;

  if (isLoading) {
    return <Loading />;
  } else if (error || defaultConfig === undefined) {
    return (
      <Box alignItems="center" display="grid" justifyItems="center">
        <img src={noData} width="50%" alt="No confusion matrix data" />
        <Typography>{error?.message || UNKNOWN_ERROR}</Typography>
      </Box>
    );
  }
```

nandhinibsn: Okay, Yes, I agree with the importance of it. Thanks

JosephMarinier: Don't forget to revert the comment as I suggested.

nandhinibsn: Okay, Sorry I missed out that part. I will add it.Thanks

JosephMarinier (approved):
Good job! Works well! 👍

nandhinibsn:
Thank you!!

PR #397, opened 2023-01-26
Add "Status" service API tag to trigger start up task on config update

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [ ] **DEV CHANGES.**

### DIFF
--- a/webapp/src/services/api.ts
+++ b/webapp/src/services/api.ts
@@ -314,6 +314,11 @@ export const api = createApi({
           tags.add("Status");
           tags.add("SimilarUtterances");
         }
+        ["uncertainty", "pipelines", "metrics", "syntax"].map((config) => {
+          if (config in partialConfig) {
+            tags.add("Status");
+          }
+        });
         return [...tags];
       },
       async onQueryStarted(


webapp/src/services/api.ts
```diff
@@ -314,6 +314,11 @@ export const api = createApi({
           tags.add("Status");
           tags.add("SimilarUtterances");
         }
+        ["uncertainty", "pipelines", "metrics", "syntax"].map((config) => {
+          if (config in partialConfig) {
+            tags.add("Status");
```
JosephMarinier, line 319: Good catch!

However, there are a lot of other "tags" (routes' caches) that should be invalidated so we get the updated responses based on the new config, without the need for reloading the app. I think trying to define in the front end the logic for what fields of the config should invalidate what tags would be error-prone (it has proven to be with this bug) and would duplicate the logic of the back end. Therefore, I propose invalidating all the tags. What do you think?

```diff
-      invalidatesTags: (...[, , { body: partialConfig }]) => {
-        const tags = new Set<Tag>();
-        // Invalidate DatasetInfo only after the query is fulfilled,
-        // otherwise the response is not up to date or even fails.
-        if ("behavioral_testing" in partialConfig) {
-          tags.add("Status");
-          tags.add("PerturbationTestingSummary");
-          tags.add("PerturbedUtterances");
-        }
-        if ("similarity" in partialConfig) {
-          tags.add("Status");
-          tags.add("SimilarUtterances");
-        }
-        return [...tags];
-      },
+      invalidatesTags: () => tagTypes,
```

JosephMarinier (approved):

nandhinibsn:
Thank you @JosephMarinier for your time in reviewing it and providing suggestions.

PR #399, opened 2023-01-27
Add a config scope for MetricsPerFilterModule

gabegma:
## Description:
* When dataset warnings thresholds will be modified, this will no longer affect `MetricsPerFilter`.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -390,13 +390,16 @@ class TopWordsConfig(SyntaxConfig, ModelContractConfig):
     pass
 
 
+class MetricPerFilterConfig(
+    PerturbationTestingConfig, SimilarityConfig, SyntaxConfig, LanguageConfig, MetricConfig
+):
+    pass
+
+
 class AzimuthConfig(
-    MetricConfig,
-    PerturbationTestingConfig,
-    SimilarityConfig,
-    DatasetWarningConfig,
-    LanguageConfig,
+    MetricPerFilterConfig,
     TopWordsConfig,
+    DatasetWarningConfig,
     extra=Extra.forbid,
 ):
     # Reminder: If a module depends on an attribute in AzimuthConfig, the module will be forced to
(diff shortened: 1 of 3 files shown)


azimuth/config.py
```diff
@@ -390,13 +390,16 @@ class TopWordsConfig(SyntaxConfig, ModelContractConfig):
     pass
 
 
+class MetricPerFilterConfig(
```
lindsaydbrin, line 393: I'm wondering if this is the best name for this layer... I get why it's called `MetricPerFilter` specifically, because that's the only module that calls this level of config, but I'm wondering if it would be more descriptive and accurate to indicate something about "per filter" without "metric"...? (Am I thinking about it correctly? It's about data subpopulations/filters, but happens to be useful for the metrics per filter module?) I also might be solving a problem that isn't a problem yet, and we can address it down the road if we have something else that is at the level of the filters. Curious what you think.

Anyhow, if we do keep it, I'd suggest `Metric` -> `Metrics` (and also for `MetricConfig`, actually), both because it feels like it makes more sense for some reason, and because then it's consistent with `MetricsPerFilterModule`.
```suggestion
class MetricsPerFilterConfig(
```

lindsaydbrin (approved):
Good refactoring thought! Thanks!

Just a question and suggestion about naming.

azimuth/config.py
gabegma: - Good catch for the 's'! 
- For the naming, I had the same thought, until I realized that this module also needs the `MetricsConfig`, and that is only specific to `MetricsPerFilterModule`. If one day, we have another module that depends on the smart tags without the metrics, we could create a new object for both, and add metrics in `MetricsPerFilterConfig` only. Does that make sense?

lindsaydbrin: Yes, perfect! Good logic.

JosephMarinier (approved):
Nice! 👍

PR #403, opened 2023-01-27
Add start-up task completion message before the application is loaded

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [ ] **DEV CHANGES.**


webapp/src/components/PipelineCheck.tsx
```diff
@@ -23,7 +24,15 @@ const PipelineCheck: React.FC<Props> = ({ children }) => {
   }
 
   if (!config) {
```
JosephMarinier, line 26: This has nothing to do with the startup tasks. This check will happen too late and is not checking the right thing. We don't need to change this file.

I think you are looking for the `StatusCheck` component instead. The message should happen when `!status.startupTasksReady` but there is no `"pending"` item in `status.startupTasksStatus`. Does that make sense?

JosephMarinier (changes requested):

webapp/src/components/StatusCheck.tsx
```diff
@@ -81,6 +81,14 @@ const StatusCheck: React.FC<Props> = ({ children }) => {
               </Box>
             ))}
           </Box>
+          {Object.values(status.startupTasksStatus).every(
+            (taskStatus) => taskStatus !== "pending"
+          ) && (
+            <Typography variant="h6" align="center">
+              The startup tasks are completed and getting finalized at the back
+              end. Please wait while the application is loaded.
+            </Typography>
```
JosephMarinier, line 90: I think we could polish the visual a bit. Also, semantically, I don't think this should be a header/title (`h6`).
Screen Shot 2023-01-27 at 3 07 21 PM
Here is a quick try with a `Paper` to draw attention, and the same spinner as for the startup tasks. What do you think @nandhinibsn and @gabegma?
Screen Shot 2023-01-30 at 9 25 57 AM
Here are the code change:
```suggestion
            <Paper>
              <Box display="flex" alignItems="center" gap={1} margin={2}>
                {PENDING}
                <Typography>
                  The results are being finalized. Hang on for a few more
                  seconds...
                </Typography>
              </Box>
            </Paper>
```
with
```diff
+const PENDING = <CircularProgress size={16} sx={{ margin: "2px" }} />;
+
 const STATUS_ICONS: Record<string, React.ReactElement> = {
   finished: <DoneIcon color="success" />,
   not_started: <DoneIcon color="success" />, // Happens when the task was already computed.
-  pending: <CircularProgress size={16} sx={{ margin: "2px" }} />,
+  pending: PENDING,
   error: <ErrorIcon color="error" />,
   lost: <ErrorIcon color="error" />,
 };
```

nandhinibsn: I like this design with loading. :)

JosephMarinier: The text should be inside the `<Typography>`, as shown in my suggestion.

JosephMarinier (approved):
👌

nandhinibsn:
Thank you, @JosephMarinier, for your time in reviewing it.

PR #406, opened 2023-01-27
Support empty list for pipelines

### LINKED ISSUE #335: Handle `"pipelines": []` in config
When you configure `"pipelines": []`, the back end currently crashes with an `IndexError: list index out of range` in `postprocessing_known()`.

I suspect users might try `"pipelines": []` if they start with a config with a pipeline and then adapt it to a new dataset, for example, and remove the pipeline.

I see two options:
1. We forbid `"pipelines": []`, but we handle it clearly in a config validator.
2. We support `"pipelines": []` just like if it was ``"pipelines": null``.

gabegma:
Resolve #335 

## Description:
That was tiny enough!

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -317,8 +317,8 @@ class ModelContractConfig(CommonFieldsConfig):
 
     @validator("pipelines", pre=True)
     def check_pipeline_names(cls, pipeline_definitions):
-        if pipeline_definitions is None:
-            return pipeline_definitions
+        if not pipeline_definitions:
+            return None
         pipeline_definitions = [
             pipeline_def.dict() if isinstance(pipeline_def, PipelineDefinition) else pipeline_def
             for pipeline_def in pipeline_definitions
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -317,6 +317,7 @@ class ModelContractConfig(CommonFieldsConfig):
 
     @validator("pipelines", pre=True)
     def check_pipeline_names(cls, pipeline_definitions):
+        # We support both [] and None (null in JSON), and we standardize it to None.
         if not pipeline_definitions:
             return None
         pipeline_definitions = [
--- a/azimuth/utils/project.py
+++ b/azimuth/utils/project.py
@@ -117,7 +117,7 @@ def perturbation_testing_available(config: PerturbationTestingConfig) -> bool:
 def postprocessing_known(config: ModelContractConfig, pipeline_index) -> bool:
     # Sometimes the postprocessing cannot be edited but we still want show the values in the FE.
     # Example: file-based
-    if config.pipelines is None:
+    if not config.pipelines:
         return False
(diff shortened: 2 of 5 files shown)


azimuth/utils/project.py
```diff
@@ -117,7 +117,7 @@ def perturbation_testing_available(config: PerturbationTestingConfig) -> bool:
 def postprocessing_known(config: ModelContractConfig, pipeline_index) -> bool:
     # Sometimes the postprocessing cannot be edited but we still want show the values in the FE.
     # Example: file-based
-    if config.pipelines is None:
+    if not config.pipelines:
```
JosephMarinier, line 120: I see `config.pipelines is( not)? None` (regex) in a dozen more places. For example:
https://github.com/ServiceNow/azimuth/blob/5ebf352b1575e12ada04ffac9d531e2295a719ee/azimuth/utils/project.py#L99-L100
We'll have to check all of them, right?
Alternatively, do you think we should check for `pipelines == []` in `@validator("pipelines"` and change it to `[]` so the rest of the code can remain as is?

gabegma: Oh right, I tested quickly, but it's true that I didn't test properly with different model contracts.

> and change it to [] so the rest of the code can remain as is.

Do you mean change it to `None`?

JosephMarinier: Right, `None`, sorry. Edited.

gabegma: Sounds good, I can do that!

JosephMarinier: If you do that, you can also rebase my commit out, it won't be necessary.

JosephMarinier: But it doesn't hurt to keep it either. Actually, I'd keep it.

azimuth/config.py
```diff
@@ -317,8 +317,8 @@ class ModelContractConfig(CommonFieldsConfig):
 
     @validator("pipelines", pre=True)
     def check_pipeline_names(cls, pipeline_definitions):
-        if pipeline_definitions is None:
-            return pipeline_definitions
+        if not pipeline_definitions:
+            return None
```
gabegma, line 321: @JosephMarinier is that what you had in mind?

gabegma: Good idea!

azimuth/config.py
```diff
@@ -317,8 +317,8 @@ class ModelContractConfig(CommonFieldsConfig):
 
     @validator("pipelines", pre=True)
     def check_pipeline_names(cls, pipeline_definitions):
-        if pipeline_definitions is None:
-            return pipeline_definitions
+        if not pipeline_definitions:
```
gabegma, line 320:
```suggestion
        # We support both [] and None (null in JSON), and we standardize it to None.
        if not pipeline_definitions:
```

PR #413, opened 2023-01-31
Add required_mod_options

### LINKED ISSUE #114: Add `required_mod_options` and raise when `mod_options` are missing them
`model_contract_method_name`
`pipeline_idx`

gabegma:
Resolve #114 

## Description:
- When calling a module without mandatory mod_options, it would sometimes fail silently, or the error message would be hard to understand. This proposal makes it clear when specific mod_options are required.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/README.md
+++ b/azimuth/modules/README.md
@@ -54,7 +54,8 @@ inherit from. Different options are available, with different particularities.
 
 - This is the parent module. No Module should directly inherit from it. It includes a lot of the
   basic methods that apply to all modules.
-- All module options are ignored by default (`allowed_mod_options` is empty) in this module.
+- All module options are ignored by default (`allowed_mod_options` and `required_mod_options` are
+  empty) in this module.
 
 ### Indexable Module
 
@@ -76,9 +77,8 @@ inherit from. Different options are available, with different particularities.
 - It inherits from Dataset Result Module, and is a module that wraps models.
 - These modules handle all the tasks which are in need of direct access to the model in order to
   compute their results (e.g. prediction, saliency).
-- `allowed_mod_options` include some default options that can affect models, such as the `threshold`
-  and the `pipeline_index`. The mod_options `model_contract_method_name` is mandatory as it
-  specifies which method should be called in the module.
+- `allowed_mod_options` include default options that can affect models, such as the `threshold`.
+- `required_mod_options` include both `model_contract_method_name` and `pipeline_index`.
 
 ### Aggregation Module
 - This module inherits from `Module`, and so the same methods and attributes are available.
@@ -88,6 +88,8 @@ inherit from. Different options are available, with different particularities.
 - This module inherits from `AggregationModule` and is similar in all points
 - However, `allowed_mod_options` include `filters`, which mean the dataset can be filtered using `DatasetFilters` (as opposed to regular modules which use indices only).
 - As such, `.get_dataset_split()` returns a subset of the `dataset_split`, based on `filters`.
+- `required_mod_options` include `pipeline_index` since all filterable modules require filtering on the dataset, which involve predictions.
+
 
 #### Comparison Module
 - This module inherits from `AggregationModule` and is similar in all points. The only difference it that it works on multiple dataset split at the same time. As such, it only accepts `DatasetSplitName.all`.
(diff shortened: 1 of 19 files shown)


azimuth/modules/base_classes/module.py
```diff
@@ -31,11 +32,11 @@ def __init__(
...
         mod_options = mod_options or ModuleOptions()
         self.mod_options = mod_options
-        if diff := (
-            set(self.mod_options.no_alias_dict(exclude_defaults=True).keys())
-            - self.allowed_mod_options
-        ):
+        defined_mod_options = set(self.mod_options.no_alias_dict(exclude_defaults=True).keys())
+        if diff := (defined_mod_options - self.allowed_mod_options - self.required_mod_options):
             raise ValueError(f"Unexpected mod_options {diff} for {self.__class__.__name__}.")
+        if self.required_mod_options and self.required_mod_options.isdisjoint(defined_mod_options):
```
JosephMarinier, line 38: I think we should check if ALL options in `required_mod_options` are in `required_mod_options`. We can use `issubset` or `<=`.
```suggestion
        if not self.required_mod_options.issubset(defined_mod_options):
```
The `isdisjoint` will pass as soon as ONE option in `required_mod_options` is in `required_mod_options`. For example, here, `2` is missing.
```python
>>> {1, 2}.isdisjoint({1})
False
```

gabegma: Oh, right! Thank you!

azimuth/modules/base_classes/module.py
```diff
@@ -52,7 +53,8 @@ def get_effective_arguments(self) -> ModuleEffectiveArguments:
         # model_contract_method_name are excluded too because it's already in the task_name.
         return ModuleEffectiveArguments(
             mod_options=self.mod_options.dict(
-                exclude={"indices", "model_contract_method_name"}, include=self.allowed_mod_options
+                exclude={"indices", "model_contract_method_name"},
+                include=self.allowed_mod_options.union(self.required_mod_options),
```
JosephMarinier, line 57: I see in most places you used the operator `|` for the union. I don't have a preference for one or for the other, but I guess we should keep it consistent.
```suggestion
                include=self.allowed_mod_options | self.required_mod_options,
```

JosephMarinier (approved):
Good catch! 👍 Thank you!

azimuth/modules/base_classes/module.py
gabegma: Right!! I forgot about it. Thanks!

azimuth/modules/README.md
```diff
@@ -76,9 +77,8 @@ inherit from. Different options are available, with different particularities.
 - It inherits from Dataset Result Module, and is a module that wraps models.
 - These modules handle all the tasks which are in need of direct access to the model in order to
   compute their results (e.g. prediction, saliency).
-- `allowed_mod_options` include some default options that can affect models, such as the `threshold`
-  and the `pipeline_index`. The mod_options `model_contract_method_name` is mandatory as it
-  specifies which method should be called in the module.
+- `allowed_mod_options` include default options that can affect models, such as the `threshold`.
```
JosephMarinier, line 80: I'm not sure I get what we mean by "default" options.

azimuth/modules/base_classes/module.py
```diff
@@ -22,6 +22,7 @@ class Module(DaskModule[ConfigScope]):
     with the dataset and the pipelines."""
 
     allowed_mod_options: Set[str] = set()
```
JosephMarinier, line 24: To make it clear that those don't also include the required options (in other words, they are disjoint), do you think `optional_mod_options` would be better? Or would it be a pain to rename?

gabegma: Yeah good idea, I can rename.

azimuth/modules/README.md
gabegma: I just removed it in my new PR.

azimuth/modules/base_classes/module.py
gabegma: See PR #416

PR #417, opened 2023-02-01
Refactor flaky test

gabegma:
## Description:
- This test sometimes fails, and I don't understand why. I cleaned it up, but I'm not sure if my code is equivalent and could improve anything. Opening this PR to brainstorm. Any ides @JosephMarinier @Dref360 ? 
Screen Shot 2023-01-31 at 5 48 19 PM

## Checklist:
  our users.
* [ ] **DEV CHANGES.**

### DIFF
--- a/tests/test_modules/test_model_contracts/test_hf_text_classification.py
+++ b/tests/test_modules/test_model_contracts/test_hf_text_classification.py
@@ -122,13 +122,14 @@ def test_rejection_class(simple_text_config):
     out = cast(List[PredictionResponse], mod.compute_on_dataset_split())
 
     # At least 1 out of the 10 predictions will have confidence below the threshold.
-    rej_cls_idx = mod.get_dataset_split_manager().rejection_class_idx
-    rejected_items = [item for item in out if item.postprocessed_output.preds[0] == rej_cls_idx]
+    rejected_items = [
+        item for item in out if item.postprocessed_output.preds[0] != item.model_output.preds[0]
+    ]
     assert len(rejected_items) > 0
     assert all(
         np.max(pred_response.model_output.probs[0]) <= simple_text_config.pipelines[0].threshold
         for pred_response in rejected_items
-    )
+    ), rejected_items
 
 
 def test_mc_dropout(simple_text_config):
--- a/tests/test_modules/test_model_contracts/test_hf_text_classification.py
+++ b/tests/test_modules/test_model_contracts/test_hf_text_classification.py
@@ -120,11 +120,11 @@ def test_rejection_class(simple_text_config):
         ),
     )
     out = cast(List[PredictionResponse], mod.compute_on_dataset_split())
-    # At least 1 out of the 10 predictions will have confidence < 0.99
-    assert set(k.postprocessed_output.preds[0] != k.model_output.preds[0] for k in out) == {True}
-    rejected_items = filter(
-        lambda k: k.postprocessed_output.preds[0] != k.model_output.preds[0], out
-    )
+
+    # At least 1 out of the 10 predictions will have confidence below the threshold.
+    rej_cls_idx = mod.get_dataset_split_manager().rejection_class_idx
+    rejected_items = [item for item in out if item.postprocessed_output.preds[0] == rej_cls_idx]
+    assert len(rejected_items) > 0
     assert all(
         np.max(pred_response.model_output.probs[0]) <= simple_text_config.pipelines[0].threshold
         for pred_response in rejected_items


tests/test_modules/test_model_contracts/test_hf_text_classification.py
```diff
@@ -120,11 +120,11 @@ def test_rejection_class(simple_text_config):
         ),
     )
     out = cast(List[PredictionResponse], mod.compute_on_dataset_split())
-    # At least 1 out of the 10 predictions will have confidence < 0.99
-    assert set(k.postprocessed_output.preds[0] != k.model_output.preds[0] for k in out) == {True}
```
JosephMarinier, line 124: This seemed to assert that ALL the post-processed predictions and different than the model predictions, otherwise it would ` == {True, False}`. Because of temperature? Is it relevant to assert that here? In other words, it should be equivalent to
```diff
-    assert set(k.postprocessed_output.preds[0] != k.model_output.preds[0] for k in out) == {True}
+    assert all(k.postprocessed_output.preds[0] != k.model_output.preds[0] for k in out)
```

tests/test_modules/test_model_contracts/test_hf_text_classification.py
```diff
@@ -120,11 +120,11 @@ def test_rejection_class(simple_text_config):
...
-        lambda k: k.postprocessed_output.preds[0] != k.model_output.preds[0], out
-    )
+
+    # At least 1 out of the 10 predictions will have confidence below the threshold.
+    rej_cls_idx = mod.get_dataset_split_manager().rejection_class_idx
+    rejected_items = [item for item in out if item.postprocessed_output.preds[0] == rej_cls_idx]
+    assert len(rejected_items) > 0
     assert all(
         np.max(pred_response.model_output.probs[0]) <= simple_text_config.pipelines[0].threshold
         for pred_response in rejected_items
```
JosephMarinier, line 130: We could add an assertion message to see what the value is next time it happens. Would that work?
```diff
-    )
+    ), rejected_items
```

tests/test_modules/test_model_contracts/test_hf_text_classification.py
```diff
@@ -120,11 +120,11 @@ def test_rejection_class(simple_text_config):
...
     out = cast(List[PredictionResponse], mod.compute_on_dataset_split())
-    # At least 1 out of the 10 predictions will have confidence < 0.99
-    assert set(k.postprocessed_output.preds[0] != k.model_output.preds[0] for k in out) == {True}
-    rejected_items = filter(
-        lambda k: k.postprocessed_output.preds[0] != k.model_output.preds[0], out
-    )
+
+    # At least 1 out of the 10 predictions will have confidence below the threshold.
+    rej_cls_idx = mod.get_dataset_split_manager().rejection_class_idx
+    rejected_items = [item for item in out if item.postprocessed_output.preds[0] == rej_cls_idx]
```
JosephMarinier, line 126: Cool, I like this `[... if ...]` instead of the `filter(...`. 👍 
Is it possible the model may predict the rejection class?
If so, the old condition `k.postprocessed_output.preds[0] != k.model_output.preds[0]` would filter them out, but not the new condition `item.postprocessed_output.preds[0] == rej_cls_idx`.

tests/test_modules/test_model_contracts/test_hf_text_classification.py
```diff
@@ -120,11 +120,11 @@ def test_rejection_class(simple_text_config):
...
+    assert len(rejected_items) > 0
     assert all(
         np.max(pred_response.model_output.probs[0]) <= simple_text_config.pipelines[0].threshold
```
JosephMarinier, line 129: Where is the condition for thresholding defined? What happens if the confidence is equal to the threshold? Should this be `<` instead of `<=`?

tests/test_modules/test_model_contracts/test_hf_text_classification.py
gabegma: Exactly ok, I was confused by the comment, so I thought that maybe the `False` was removed automatically in a `set` for some reason I didn't know.

tests/test_modules/test_model_contracts/test_hf_text_classification.py
gabegma: Great idea!!

tests/test_modules/test_model_contracts/test_hf_text_classification.py
gabegma: It's not possible in this case with `simple_text_config`, but I could change it back if we think it's more robust.

tests/test_modules/test_model_contracts/test_hf_text_classification.py
gabegma: No, the function is `post_processing_io.probs.max(axis=-1) > threshold`.

tests/test_modules/test_model_contracts/test_hf_text_classification.py
gabegma: I added a `assert len(rejected_items) > 0` to replace this though, so it's no longer needed right?

JosephMarinier: So we go with "at least 1 rejection" from the comment, and not the "all rejections" from the original code? I think it makes more sense. 👍 Yes, in that case, `assert len(rejected_items) > 0` works. 👌

tests/test_modules/test_model_contracts/test_hf_text_classification.py
JosephMarinier: Perfect, that makes sense. 👍

JosephMarinier (approved):
As you said, I'm not sure we solved the problem, but we made the code easier to understand and with the added assertion print we should be able to understand the error next time it happens. 👍

PR #425, opened 2023-02-07
Simplify add missing behavioral testing fields

JosephMarinier:


webapp/src/pages/Settings.tsx
```diff
@@ -550,44 +546,21 @@ const Settings: React.FC = () => {
...
-                      <Typography variant="body2">
-                        {Array.isArray(value) ? value.join(", ") : value}
-                      </Typography>
-                    </React.Fragment>
-                  ))}
-                </KeyValuePairs>
-              </Box>
-            )
+        {(
+          ["neutral_token", "punctuation", "fuzzy_matching", "typo"] as const
```
nandhinibsn, line 550: Okay, It's a bit clean. But is the list going to be constant for all configs?

nandhinibsn: Ohh! right. It makes sense now. I think we can apply your simplified changes.

PR #421, opened 2023-02-06
Add behavioral testing fields to config page

nandhinibsn:
Resolve #

## Checklist:

You should check all boxes before the PR is ready. If a box does not apply, check it to acknowledge it.

* [x] **ISSUE NUMBER.** You linked the issue number (Ex: Resolve #XXX).
* [x] **PRE-COMMIT.** You ran pre-commit on all commits, or else, you
  ran `pre-commit run --all-files` at the end.
* [x] **USER CHANGES.** The changes are added to CHANGELOG.md and the documentation, if they impact
  our users.
* [x] **DEV CHANGES.**
    * Update the documentation if this PR changes how to develop/launch on the app.
    * Update the `README` files and our wiki for any big design decisions, if relevant.
    * Add unit tests, docstrings, typing and comments for complex sections.

PR #426, opened 2023-02-07
Add all missing analyses config fields

JosephMarinier:
## Description:

Add to the config page all the fields that were missing from the syntax, similarity, and behavioral testing sections.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -556,7 +556,7 @@ const Settings: React.FC = () => {
       {getAnalysesCustomization("syntax")}
       {displayToggleSectionTitle("similarity", "Similarity")}
       {getAnalysesCustomization("similarity")}
-      {displayToggleSectionTitle("behavioral_testing", "Perturbation Testing")}
+      {displayToggleSectionTitle("behavioral_testing", "Behavioral Testing")}
       {getAnalysesCustomization("behavioral_testing")}
     </>
   );
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -93,20 +93,20 @@ const KeyValuePairs: React.FC = ({ children }) => (
 );
 
 const displayKeywordArguments = (name: string, kwargs: Record<string, any>) => (
-  <Box display="grid">
+  <Box key={name} display="flex" flexDirection="column">
     <Typography variant="caption">{name}</Typography>
     <KeyValuePairs>
       {Object.entries(kwargs).map(([field, value], index) => (
         <React.Fragment key={index}>
           <Typography variant="body2">{field}:</Typography>
-          <Tooltip title={value}>
+          <Tooltip title={Array.isArray(value) ? value.join(", ") : value}>
             <Typography
               variant="body2"
               whiteSpace="nowrap"
               overflow="hidden"
               textOverflow="ellipsis"
             >
-              {value}
+              {Array.isArray(value) ? value.join(", ") : value}
             </Typography>
           </Tooltip>
         </React.Fragment>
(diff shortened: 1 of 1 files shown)


nandhinibsn:
True. this is simpler. I think we can apply these changes.

webapp/src/pages/Settings.tsx
```diff
@@ -551,6 +557,7 @@ const Settings: React.FC = () => {
       {displayToggleSectionTitle("similarity", "Similarity")}
       {getAnalysesCustomization("similarity")}
       {displayToggleSectionTitle("behavioral_testing", "Perturbation Testing")}
+      {getAnalysesCustomization("behavioral_testing")}
```
nandhinibsn, line 560: Do we not hide the section if behavioral testing is null? Do we display default config values if it's null?

webapp/src/pages/Settings.tsx
```diff
@@ -551,6 +557,7 @@ const Settings: React.FC = () => {
       {displayToggleSectionTitle("similarity", "Similarity")}
       {getAnalysesCustomization("similarity")}
       {displayToggleSectionTitle("behavioral_testing", "Perturbation Testing")}
```
nandhinibsn, line 559: Display name should be changed to "Behavioral Testing"

JosephMarinier: Good catch! I'll change it in this PR. Thank you!

webapp/src/pages/Settings.tsx
JosephMarinier: Right, I was waiting for your input [here](https://github.com/ServiceNow/azimuth/pull/421/files#r1099033775) before changing that. What do you think?

webapp/src/pages/Settings.tsx
```diff
@@ -93,20 +93,20 @@ const KeyValuePairs: React.FC = ({ children }) => (
 );
 
 const displayKeywordArguments = (name: string, kwargs: Record<string, any>) => (
-  <Box display="grid">
+  <Box key={name} display="flex" flexDirection="column">
```
JosephMarinier, line 96: @nandhinibsn
> Changing it to flex won't shrink the content when we scroll?

With the `display="grid"`, I was getting this:
image

With `display="flex" flexDirection="column"`, I get:
image

Can you reproduce any issue with `flex`?

webapp/src/pages/Settings.tsx
nandhinibsn: Okay, I'm so sorry I did not respond there. Yes, I do prefer hiding the content; but we disable in similarity and postprocessor.

webapp/src/pages/Settings.tsx
nandhinibsn: Thank you so much!

webapp/src/pages/Settings.tsx
JosephMarinier: No worries!
Right, showing the default values, but grayed out, is also interesting. You have done it for similarity and postprocessor fields, except it doesn't work for read-only fields.
image
We could fix that. Or we can hide. Maybe we can ask the team in standup tomorrow?

webapp/src/pages/Settings.tsx
nandhinibsn: Hey @JosephMarinier, I tried later with adding flex, and it did not shrink. So, it should be fine. Thanks.

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, sure, we can ask for clarification. True, disable feature is not added to the read-only part.

webapp/src/pages/Settings.tsx
JosephMarinier: Awesome, thank you!

PR #427, opened 2023-02-08
Refactor dataset manager

gabegma:
Part of #297

## Description:
* Refactor `get_tags` so the output is the same as the input format for `add_tags`. It felt odd to me that they were slightly different (list vs dict). Let me know if you disagree, I can change it back to a list. 
   * I removed the pandas conversion and the method is now faster. 
   * I refactored the test which was messy and didn't test much the `get_tags` method.
* Add a new method to convert `persistent_id` to `row_idx`. This will be useful for my next PR, where we modify the route to update the tags.  

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/dataset_split_manager.py
+++ b/azimuth/dataset_split_manager.py
@@ -310,7 +310,7 @@ class DatasetSplitManager:
 
     def get_tags(
         self, indices: Optional[List[int]] = None, table_key: Optional[PredictionTableKey] = None
-    ) -> List[Dict[str, bool]]:
+    ) -> Dict[int, Dict[Tag, bool]]:
         """Get tags from the dataset split.
 
         Args:
@@ -318,17 +318,18 @@ class DatasetSplitManager:
             table_key: Predictions table to gather prediction tags.
 
         Returns:
-            List of records per index.
-
+             Value of tags per row_idx.
         """
         if indices is not None and len(indices) > 0:
             ds = self.get_dataset_split(table_key).select(indices)
         else:
             ds = self.get_dataset_split(table_key)
         available_tags = self._tags if table_key is None else self._tags + self._prediction_tags
-        df = pd.DataFrame({t: ds[t] for t in available_tags})
-        tags: List[Dict[str, bool]] = df.to_dict(orient="records")
-        return tags
+
+        return {
+            row_idx: {tag: ds[tag][idx] for tag in available_tags}
+            for idx, row_idx in enumerate(ds["row_idx"])
+        }
 
     def class_distribution(self, labels_only=False):
         """Compute the class distribution for a dataset_split.
--- a/azimuth/dataset_split_manager.py
+++ b/azimuth/dataset_split_manager.py
@@ -319,12 +319,11 @@ class DatasetSplitManager:
             table_key: Predictions table to gather prediction tags.
 
         Returns:
-             Value of tags per row_idx.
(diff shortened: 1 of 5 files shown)


azimuth/dataset_split_manager.py
```diff
@@ -305,25 +311,26 @@ def add_tags(
...
         """Get tags from the dataset split.
 
         Args:
             indices: Set of indices to select.
             table_key: Predictions table to gather prediction tags.
 
         Returns:
-            List of records per index.
-
+             Value of tags per row_idx.
```
JosephMarinier, line 322:
```suggestion
            Value of tags per row_idx.
```

azimuth/modules/model_performance/outcome_count.py
```diff
@@ -84,7 +84,7 @@ def get_outcome_count_per_tag(
             indices=assert_is_list(ds[DatasetColumn.row_idx]), table_key=self._get_table_key()
         )
         for utterance_tags, outcome in zip(
-            all_tags, get_outcomes_from_ds(ds, self.mod_options.without_postprocessing)
+            all_tags.values(), get_outcomes_from_ds(ds, self.mod_options.without_postprocessing)
```
JosephMarinier, line 87: Oh cool! PyCharm was confused, thinking `dict` was equivalent to `dict.keys()` instead of `dict.values()`, right? Good idea!

tests/test_dataset_manager.py
```diff
@@ -407,6 +410,28 @@ def test_custom_persistent_id(simple_text_config, a_text_dataset):
...
+        DatasetSplitName.eval,
+        simple_text_config_persistent_id,
+        initial_tags=ALL_STANDARD_TAGS,
+        dataset_split=a_text_dataset,
+    )
+    ds = dm.get_dataset_split()
+
+    persistent_ids = list(reversed(ds[simple_text_config.columns.text_input]))
+    row_indices = dm.get_row_indices_from_persistent_id(persistent_ids)
+    assert row_indices == list(reversed(range(len(ds))))
```
JosephMarinier, line 427: That's clever. 😚👌

azimuth/dataset_split_manager.py
```diff
@@ -305,25 +311,26 @@ def add_tags(
...
         """
         if indices is not None and len(indices) > 0:
             ds = self.get_dataset_split(table_key).select(indices)
         else:
             ds = self.get_dataset_split(table_key)
```
JosephMarinier, line 327: I know it's not new code, but I couldn't find anything in your new code so... 😅
```suggestion
        ds = self.get_dataset_split(table_key)
        if indices is not None and len(indices) > 0:
            ds = ds.select(indices)
```

JosephMarinier (approved):
🚢 🇮🇹

azimuth/modules/model_performance/outcome_count.py
gabegma: That is actually the behavior of python, it loops over the `keys` instead of the `values`. I was surprised.  🤷‍♀️

JosephMarinier: Oh right! So that change is because `all_tags` comes from `dm.get_tags()`, which changed from a list to a `dict`. I am following now. 😅 Thanks for the explanation!

PR #428, opened 2023-02-09
Add language field to config page

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -240,7 +240,7 @@ const Settings: React.FC = () => {
         behavioral_testing: resultingConfig.behavioral_testing && {
           ...resultingConfig.behavioral_testing,
           neutral_token: {
-            ...resultingConfig.behavioral_testing?.neutral_token,
+            ...resultingConfig.behavioral_testing.neutral_token,
             suffix_list:
               defaultConfig.behavioral_testing!.neutral_token.suffix_list,
             prefix_list:
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -13,8 +13,11 @@ import {
   InputAdornment,
   InputBaseComponentProps,
   inputClasses,
+  InputLabel,
   inputLabelClasses,
+  MenuItem,
   Paper,
+  Select,
   TextField,
   TextFieldProps,
   Tooltip,
@@ -63,7 +66,7 @@ type SubConfigKeys = keyof PickByValue<AzimuthConfig, object | null>;
 
 const CUSTOM_METRICS: string[] = ["Accuracy", "Precision", "Recall", "F1"];
 const ADDITIONAL_KWARGS_CUSTOM_METRICS = ["Precision", "Recall", "F1"];
-
+const SUPPORTED_LANGUAGES = ["en", "fr"];
 const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof AzimuthConfig)[] = [
   "behavioral_testing",
   "similarity",
@@ -548,8 +551,25 @@ const Settings: React.FC = () => {
     </FormGroup>
   );
 
+  const displayAnalysesCustomizationGeneralSection = () => (
+    <FormControl variant="standard" sx={{ m: 1, width: 120 }}>
+      <InputLabel id="supported-language-input-label">
+        supported_language
+      </InputLabel>
+      <Select value={resultingConfig.language} label="supported_language">
+        {SUPPORTED_LANGUAGES.map((language) => (
+          <MenuItem key={language} value={language}>
+            {language}
+          </MenuItem>
+        ))}
+      </Select>
+    </FormControl>
+  );
+
   const getAnalysesCustomizationSection = () => (
     <>
+      {displaySectionTitle("General")}
+      {displayAnalysesCustomizationGeneralSection()}
       {displaySectionTitle("Dataset Warnings")}
       {getAnalysesCustomization("dataset_warnings")}
       {displaySectionTitle("Syntax")}
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -34,7 +34,11 @@ import {
   getDefaultConfigEndpoint,
   updateConfigEndpoint,
 } from "services/api";
-import { AzimuthConfig, PipelineDefinition } from "types/api";
+import {
+  AzimuthConfig,
+  PipelineDefinition,
+  SupportedLanguage,
+} from "types/api";
 import { PickByValue } from "types/models";
 import { UNKNOWN_ERROR } from "utils/const";
 
@@ -197,11 +201,15 @@ const NumberField: React.FC<
 
 const Settings: React.FC = () => {
   const { jobId } = useParams<{ jobId: string }>();
+  const [language, setLanguage] = React.useState<
+    SupportedLanguage | undefined
+  >();
+
   const {
     data: defaultConfig,
     isLoading,
     error,
-  } = getDefaultConfigEndpoint.useQuery({ jobId });
+  } = getDefaultConfigEndpoint.useQuery({ jobId, language });
   const { data: config } = getConfigEndpoint.useQuery({ jobId });
   const [updateConfig, { isLoading: isUpdatingConfig }] =
     updateConfigEndpoint.useMutation();
@@ -556,7 +564,13 @@ const Settings: React.FC = () => {
       <InputLabel id="supported-language-input-label">
         supported_language
       </InputLabel>
-      <Select value={resultingConfig.language} label="supported_language">
+      <Select
+        value={language ?? resultingConfig.language}
+        label="supported_language"
+        onChange={(event) =>
+          setLanguage(event.target.value as SupportedLanguage)
+        }
+      >
         {SUPPORTED_LANGUAGES.map((language) => (
           <MenuItem key={language} value={language}>
             {language}
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -204,7 +204,6 @@ const Settings: React.FC = () => {
   const [language, setLanguage] = React.useState<
     SupportedLanguage | undefined
   >();
-
   const {
     data: defaultConfig,
     isLoading,
@@ -217,6 +216,17 @@ const Settings: React.FC = () => {
   const [partialConfig, setPartialConfig] = React.useState<
     Partial<AzimuthConfig>
   >({});
+
+  React.useEffect(() => {
+    if (defaultConfig && language !== undefined) {
+      setPartialConfig({
+        ...partialConfig,
+        syntax: defaultConfig.syntax,
+        similarity: defaultConfig.similarity,
+        behavioral_testing: defaultConfig.behavioral_testing,
+      });
+    }
+  }, [defaultConfig]);
 
   // If config was undefined, PipelineCheck would not even render the page.
   if (config === undefined) return null;
@@ -560,24 +570,33 @@ const Settings: React.FC = () => {
   );
 
   const displayAnalysesCustomizationGeneralSection = () => (
-    <FormControl variant="standard" sx={{ m: 1, width: 120 }}>
-      <InputLabel id="supported-language-input-label">
-        supported_language
-      </InputLabel>
-      <Select
-        value={language ?? resultingConfig.language}
-        label="supported_language"
-        onChange={(event) =>
-          setLanguage(event.target.value as SupportedLanguage)
-        }
-      >
-        {SUPPORTED_LANGUAGES.map((language) => (
-          <MenuItem key={language} value={language}>
-            {language}
-          </MenuItem>
-        ))}
-      </Select>
-    </FormControl>
+    <Box display="flex" justifyContent="flex-start" gap={5} alignItems="center">
+      <FormControl variant="standard" sx={{ m: 1, width: 120 }}>
+        <InputLabel id="supported-language-input-label">
+          supported_language
+        </InputLabel>
+        <Select
+          value={language ?? resultingConfig.language}
+          label="supported_language"
+          onChange={(event) =>
+            setLanguage(event.target.value as SupportedLanguage)
+          }
+        >
+          {SUPPORTED_LANGUAGES.map((language) => (
+            <MenuItem key={language} value={language}>
+              {language}
+            </MenuItem>
+          ))}
+        </Select>
+      </FormControl>
+      <Box display="flex" flexDirection="row" gap={1}>
+        <Warning color="warning" />
+        <Typography variant="subtitle2">
+          Changing the language would impact the syntax, similarity and
+          behavioral_testing sections
+        </Typography>
+      </Box>
+    </Box>
   );
 
   const getAnalysesCustomizationSection = () => (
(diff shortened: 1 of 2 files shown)


nandhinibsn:
image

webapp/src/pages/Settings.tsx
```diff
@@ -548,8 +569,40 @@ const Settings: React.FC = () => {
     </FormGroup>
   );
 
+  const displayAnalysesCustomizationGeneralSection = () => (
+    <Box display="flex" justifyContent="flex-start" gap={5} alignItems="center">
+      <FormControl variant="standard" sx={{ m: 1, width: 120 }}>
+        <InputLabel id="supported-language-input-label">
+          supported_language
```
JosephMarinier, line 576: As you can see 3 lines below, the field is named `language` and not `supported_language`. The list of the languages we support can be called `SUPPORTED_LANGUAGES`, that's perfect, but not the one language selected from the list. Does that make sense?
```suggestion
          language
```

webapp/src/pages/Settings.tsx
```diff
@@ -548,8 +569,40 @@ const Settings: React.FC = () => {
...
+        </InputLabel>
+        <Select
```
JosephMarinier, line 578: For the `InputLabel`'s `id` to be useful, you must add it as the `labelId` of the `Select`.
```suggestion
        <InputLabel id="language-input-label">
          language
        </InputLabel>
        <Select
          labelId"language-input-label"
```

webapp/src/pages/Settings.tsx
```diff
@@ -548,8 +569,40 @@ const Settings: React.FC = () => {
...
+          {SUPPORTED_LANGUAGES.map((language) => (
+            <MenuItem key={language} value={language}>
+              {language}
+            </MenuItem>
+          ))}
+        </Select>
+      </FormControl>
+      <Box display="flex" flexDirection="row" gap={1}>
+        <Warning color="warning" />
+        <Typography variant="subtitle2">
```
JosephMarinier, line 594: That's not a subtitle.
```suggestion
        <Typography variant="body2">
```

webapp/src/pages/Settings.tsx
nandhinibsn: Okay. Make sense. I will apply the suggestion

webapp/src/pages/Settings.tsx
JosephMarinier: Please review my suggestion carefully. The `InputLabel`'s `id` must be the same string as the `Select`'s `labelId`. That's what links them together for the browsers' accessibility features.

nandhinibsn: Okay. My Apologise. I did not understand the values of both IDs correctly should be the same. I will remove the supported wherever it's inappropriate

webapp/src/pages/Settings.tsx
```diff
@@ -63,7 +70,7 @@ type SubConfigKeys = keyof PickByValue<AzimuthConfig, object | null>;
 
 const CUSTOM_METRICS: string[] = ["Accuracy", "Precision", "Recall", "F1"];
 const ADDITIONAL_KWARGS_CUSTOM_METRICS = ["Precision", "Recall", "F1"];
-
+const SUPPORTED_LANGUAGES = ["en", "fr"];
```
JosephMarinier, line 73: I think it would be wise to enforce de type here, so we notice if the supported languages change.
```suggestion
const SUPPORTED_LANGUAGES: SupportedLanguage[] = ["en", "fr"];
```

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,17 @@ const Settings: React.FC = () => {
     Partial<AzimuthConfig>
   >({});
 
+  React.useEffect(() => {
+    if (defaultConfig && language !== undefined) {
+      setPartialConfig({
+        ...partialConfig,
+        syntax: defaultConfig.syntax,
+        similarity: defaultConfig.similarity,
+        behavioral_testing: defaultConfig.behavioral_testing,
```
JosephMarinier, line 226: I think we should only overwrite the fields that are affected by language. For example, `config.syntax.short_sentence_max_word` doesn't change based on language, so we should not overwrite it. If you search the repo for the comment `# Language-based`, you will find the 6 fields that depend on language.

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,17 @@ const Settings: React.FC = () => {
...
+      });
+    }
+  }, [defaultConfig]);
```
JosephMarinier, line 229: I guess you get an error in the console since this dependency array is missing `language`?
```suggestion
  }, [defaultConfig, language]);
```

webapp/src/pages/Settings.tsx
```diff
@@ -548,8 +569,38 @@ const Settings: React.FC = () => {
     </FormGroup>
   );
 
+  const displayAnalysesCustomizationGeneralSection = () => (
+    <Box display="flex" justifyContent="flex-start" gap={5} alignItems="center">
+      <FormControl variant="standard" sx={{ m: 1, width: 120 }}>
```
JosephMarinier, line 574: A couple of things here:
1. The `justifyContent="flex-start"` doesn't seem to have any effect, can we remove it?
1. Like the other sections, this one should be wrapped inside a `<FormGroup>` to give it the standard margin. You can then remove the `m: 1` on `FormControl`.
1. Let's give the input the same width as the number fields, for consistency. This is done via a class name, currently called `number`, but that we could rename to something like `fixedWidthInput`. You can then remove the `width: 120` on `FormControl`.

```diff
       variant="standard"
       size="small"
       type="number"
-      className="number"
+      className="fixedWidthInput"
```
```diff
-          [`& .number .${inputClasses.root}`]: { width: "12ch" },
+          [`& .fixedWidthInput .${inputClasses.root}`]: { width: "12ch" },
```

```suggestion
    <FormGroup>
      <Box display="flex" gap={5} alignItems="center">
        <FormControl variant="standard" className="fixedWidthInput">
```

webapp/src/pages/Settings.tsx
JosephMarinier: It should also set `language` itself.

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,17 @@ const Settings: React.FC = () => {
     Partial<AzimuthConfig>
   >({});
 
+  React.useEffect(() => {
+    if (defaultConfig && language !== undefined) {
```
JosephMarinier, line 221: To check if we got a new `defaultConfig`, we can check if its `language` is different than the current one. For that, we will need to move initializing `resultingConfig` right before that:
```suggestion
  const resultingConfig = Object.assign({}, config, partialConfig);

  React.useEffect(() => {
    if (defaultConfig && defaultConfig.language !== resultingConfig.language) {
```

webapp/src/pages/Settings.tsx
JosephMarinier: That being said, I think the best solution is rather to not depend on `language` and do this: https://github.com/ServiceNow/azimuth/pull/428/files#r1103166073

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, this one needed a discussion. I will include only those impacted fields here.

webapp/src/pages/Settings.tsx
JosephMarinier: Was there a problem moving `resultingConfig` before the `React.useEffect()`? Instead of doing all that `partialConfig.<...> ?? config.<...>` logic?

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,40 @@ const Settings: React.FC = () => {
...
+    if (config && defaultConfig && defaultConfig.language !== config.language) {
+      setPartialConfig({
+        language: defaultConfig.language,
+        syntax: {
+          ...(partialConfig.syntax ?? config.syntax),
+          spacy_model: defaultConfig.syntax.spacy_model,
+          subj_tags: defaultConfig.syntax.subj_tags,
+          obj_tags: defaultConfig.syntax.obj_tags,
+        },
+        similarity: {
```
JosephMarinier, line 230: Changing language should not change the checkbox for enabling/disabling similarity. I think you have to do something like this to preserve the `null`:
```suggestion
        similarity: partialConfig.similarity && {
```

nandhinibsn: Ohh right! Good catch!

webapp/src/pages/Settings.tsx
nandhinibsn: Right!. I can explain the issue of using"resultingConfig."

If I add "resultingConfig" to the use effect dependency array, I get the warning "Warning: Maximum update depth exceeded. This can happen when a component calls setState inside use effect, but use effect either doesn't have a dependency array or one of the dependencies changes on every render." in the console. 

If I don't add it to the dependency array, the runtime error is gone, but I get compile time warning.

JosephMarinier: If "one of the dependencies changes on every render", I guess that's because your condition for calling `setPartialConfig` had an issue.

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,38 @@ const Settings: React.FC = () => {
...
+      setPartialConfig({
+        language: defaultConfig.language,
+        syntax: {
+          ...resultingConfig.syntax,
+          spacy_model: defaultConfig.syntax.spacy_model,
+          subj_tags: defaultConfig.syntax.subj_tags,
+          obj_tags: defaultConfig.syntax.obj_tags,
+        },
+        similarity: resultingConfig.similarity && {
+          ...(resultingConfig.similarity ?? defaultConfig.similarity!),
```
JosephMarinier, line 233: Since you added `resultingConfig.similarity &&`, `resultingConfig.similarity` will always be defined, here.
```suggestion
          ...resultingConfig.similarity,
```

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,38 @@ const Settings: React.FC = () => {
...
+          faiss_encoder: defaultConfig.similarity!.faiss_encoder,
+        },
+        behavioral_testing: resultingConfig.behavioral_testing && {
+          ...(resultingConfig.behavioral_testing ??
+            defaultConfig.behavioral_testing!),
```
JosephMarinier, line 238: Same:
```suggestion
          ...resultingConfig.behavioral_testing,
```

webapp/src/pages/Settings.tsx
```diff
@@ -194,11 +201,14 @@ const NumberField: React.FC<
...
   const { jobId } = useParams<{ jobId: string }>();
+  const [language, setLanguage] = React.useState<
+    SupportedLanguage | undefined
+  >();
   const {
     data: defaultConfig,
     isLoading,
     error,
-  } = getDefaultConfigEndpoint.useQuery({ jobId });
+  } = getDefaultConfigEndpoint.useQuery({ jobId, language });
```
JosephMarinier, line 211: With this setup, the default config is initially fetched with `language: undefined`, which defaults to English in the back end. Let's imagine a scenario where
- the `config.language` is French;
- the `config.behavioral_testing` is `null`;
- the user doesn't touch the language selector, so the `defaultConfig` stays English;
- the user enables `behavioral_testing`; then
- the `partialConfig` will be populated with the English default value, even though `config.language` is French.

I think the solution is to fetch `getDefaultConfigEndpoint` after `getConfigEndpoint` (so moving it down), using `language: language ?? resultingConfig.language` so it's never `undefined`.
```suggestion
  } = getDefaultConfigEndpoint.useQuery({
    jobId,
    language: language ?? resultingConfig.language,
  });
```

webapp/src/pages/Settings.tsx
```diff
@@ -207,6 +217,38 @@ const Settings: React.FC = () => {
...
+        similarity: resultingConfig.similarity && {
+          ...(resultingConfig.similarity ?? defaultConfig.similarity!),
+          faiss_encoder: defaultConfig.similarity!.faiss_encoder,
+        },
+        behavioral_testing: resultingConfig.behavioral_testing && {
+          ...(resultingConfig.behavioral_testing ??
+            defaultConfig.behavioral_testing!),
+          neutral_token: {
+            ...(resultingConfig.behavioral_testing?.neutral_token ??
+              defaultConfig.behavioral_testing!.neutral_token),
```
JosephMarinier, line 241: Same:
```suggestion
            ...resultingConfig.behavioral_testing.neutral_token,
```

nandhinibsn: right!.. done

webapp/src/pages/Settings.tsx
nandhinibsn: Yes!. True.. Done.

webapp/src/pages/Settings.tsx
```diff
@@ -220,7 +264,6 @@ const Settings: React.FC = () => {
       </Box>
     );
   }
-  const resultingConfig = Object.assign({}, config, partialConfig);
 
   const displayToggleSectionTitle = (
```
JosephMarinier, line 268: I notice a small problem here. Steps to reproduce:
1. Uncheck Behavioral Testing.
2. Change language.
3. Notice the default values for Behavioral Testing changed as expected.
4. Check Behavioral Testing.
5. Notice all Behavioral Testing fields are reverted to their original values (from `config`), ignoring the new selected language.

I think the solution is to remove `config[field] ?? ` here:
```diff
           onChange={(...[, checked]) =>
             setPartialConfig({
               ...partialConfig,
-              [field]: checked ? config[field] ?? defaultConfig[field] : null,
+              [field]: checked ? defaultConfig[field] : null,
             })
           }
```

webapp/src/pages/Settings.tsx
JosephMarinier: You forgot to remove the unnecessary `?` as suggested. I see 4 unnecessary `!`s and `?`s in total. Let's clean that up.

webapp/src/pages/Settings.tsx
nandhinibsn: Yes. It makes sense. We don't require config if we enable them freshly. In this case, it's similarity and behavioral testing.

webapp/src/pages/Settings.tsx
JosephMarinier: Right!

JosephMarinier (approved):
Cool!

webapp/src/pages/Settings.tsx
```diff
@@ -53,8 +60,8 @@ const FIELDS: Record<
   max_delta_representation: PERCENTAGE,
   max_delta_mean_words: { ...FLOAT, units: "words" },
   max_delta_std_words: { ...FLOAT, units: "words" },
-  short_utterance_max_word: { ...INT, units: "words" },
-  long_utterance_min_word: { ...INT, units: "words" },
+  short_sentence_max_word: { ...INT, units: "words" },
+  long_sentence_min_word: { ...INT, units: "words" },
```
JosephMarinier, line 64: I think you did a mistake while resolving conflicts with `main`.
```suggestion
  short_utterance_max_word: { ...INT, units: "words" },
  long_utterance_min_word: { ...INT, units: "words" },
```

nandhinibsn: I had no conflict showing for it. But, Yeah, I noticed the API response field is different from the UI, so I manually changed it.

JosephMarinier: Restart the back end to get the latest changes.

JosephMarinier (approved):

nandhinibsn:
Thank you for the helpful suggestions and taking your time in reviewing it @JosephMarinier  :)

PR #429, opened 2023-02-09
Adapt route to add proposed actions

gabegma:
Part of #297 

## Description:
- New route to import proposed actions

## Remaining
- [x] Adapt call to update tags in the front end.

## Checklist:

  our users.
* [ ] **DEV CHANGES.**


azimuth/routers/v1/utterances.py
```diff
@@ -264,6 +265,47 @@ def get_utterances(
...
+    dataset_split_manager.add_tags(request_data_complete)
+
+    task_manager.clear_worker_cache()
+    updated_tags = dataset_split_manager.get_tags(row_indices)
+
+    return [
+        UtterancePatch(
+            persistent_id=persistent_id,
+            data_action=next(
+                iter(tag for tag, value in tags.items() if tag in ALL_DATA_ACTIONS and value),
```
JosephMarinier, line 301: I just checked and we don't need to create an iterator; the generator should suffice. I'm not sure why we have `next(iter(...)` everywhere in the repo. Because we supply the default value too, we need some parentheses around the generator, though, otherwise we get a syntax error. Maybe that's what confused us into using `iter()` before. 🤷‍♂️ 
```suggestion
               (tag for tag, value in tags.items() if tag in ALL_DATA_ACTIONS and value),
```

JosephMarinier: Ah! The `iter()` is necessary when we use a list or some `dict.values()`, which are not iterators (whereas generators are). I cleaned the ones I could in https://github.com/ServiceNow/azimuth/pull/429/commits/1379630459d185a444558543c5513320c418d4cb

tests/test_routers/test_utterances.py
```diff
@@ -147,3 +148,18 @@ def test_perturbed_utterances(app: FastAPI, monkeypatch):
...
+    request = [{"persistent_id": 0, "data_action": "remove"}]
+    resp = client.post("/dataset_splits/eval/utterances", json=request)
+    assert resp.status_code == HTTP_200_OK, resp.text
+    assert resp.json() == [{"persistentId": 0, "dataAction": "remove"}]
+
+    # Reset tag to NO_ACTION
+    request = [{"persistent_id": 0, "data_action": "NO_ACTION"}]
+    resp = client.post("/dataset_splits/eval/utterances", json=request)
+    assert resp.status_code == HTTP_200_OK, resp.text
+    assert resp.json() == [{"persistentId": 0, "dataAction": "NO_ACTION"}]
```
JosephMarinier, line 165: Just to be super explicit, at a glance, I'd reuse `request` for asserting the response.
```suggestion
    assert resp.json() == request

    # Reset tag to NO_ACTION
    request = [{"persistent_id": 0, "data_action": "NO_ACTION"}]
    resp = client.post("/dataset_splits/eval/utterances", json=request)
    assert resp.status_code == HTTP_200_OK, resp.text
    assert resp.json() == request
```

azimuth/routers/v1/utterances.py
```diff
@@ -264,6 +265,47 @@ def get_utterances(
...
 
+@router.post(
+    "",
+    summary="Patch utterances",
+    description="Patch utterances, such as updating proposed actions.",
+    tags=TAGS,
+    response_model=List[UtterancePatch],
+)
+def patch_utterances(
+    request_data: List[UtterancePatch] = Body(...),
```
JosephMarinier, line 276: If you're interested, you can rename this variable to whatever you want. I'm saying in case you thought `request_data` was a "reserved" name for that. For example, it could be
```suggestion
    utterances: List[UtterancePatch] = Body(...),
```

azimuth/routers/v1/utterances.py
```diff
@@ -264,6 +265,47 @@ def get_utterances(
...
+    dataset_split_manager: DatasetSplitManager = Depends(get_dataset_split_manager),
+    task_manager: TaskManager = Depends(get_task_manager),
+) -> List[UtterancePatch]:
+    persistent_ids = [utt.persistent_id for utt in request_data]
+    try:
+        row_indices = dataset_split_manager.get_row_indices_from_persistent_id(persistent_ids)
+    except ValueError as e:
+        raise HTTPException(HTTP_404_NOT_FOUND, detail=f"Persistent id not found: {e}.")
+
+    request_data_complete = {}
```
JosephMarinier, line 286: Ah, I see what you mean by `_complete`, but it was more confusing to me than helpful when I first read the name. Like is it adding the "complete" list of utterances? Or adding all utterances' fields? I think I would have caught it if it was named `tags`, in the sense that we are converting to the format of our `tags`. What do you think?
```suggestion
    tags = {}
```

azimuth/routers/v1/utterances.py
```diff
@@ -264,6 +265,47 @@ def get_utterances(
...
+    description="Patch utterances, such as updating proposed actions.",
+    tags=TAGS,
+    response_model=List[UtterancePatch],
+)
+def patch_utterances(
+    request_data: List[UtterancePatch] = Body(...),
+    dataset_split_manager: DatasetSplitManager = Depends(get_dataset_split_manager),
+    task_manager: TaskManager = Depends(get_task_manager),
+) -> List[UtterancePatch]:
+    persistent_ids = [utt.persistent_id for utt in request_data]
```
JosephMarinier, line 280: Like, how beautiful would that be? 😍 
```suggestion
    persistent_ids = [utterance.persistent_id for utterance in utterances]
```

gabegma: Haha sold. I did think `request_data` was maybe standard for a body. Thank you!!

azimuth/routers/v1/utterances.py
gabegma: I like it, thanks!

gabegma: Actually, I think I prefer `data_actions`

gabegma:
Should the front-end be reviewed before merging @JosephMarinier?

tests/test_routers/test_utterances.py
gabegma: Right, this doesn't work as easily because of camel case vs snake case. You can check my last commit and see if that makes sense or if there is an easier way.

JosephMarinier:
I think the front end is fine. It was very straightforward.

JosephMarinier (approved):
You can give camelCase to the API. That's what the front end does. Two changes...

tests/test_routers/test_utterances.py
JosephMarinier: Ah! I didn't notice. Here is what you can do: https://github.com/ServiceNow/azimuth/pull/429#pullrequestreview-1296293067

tests/test_routers/test_utterances.py
```diff
@@ -147,3 +150,18 @@ def test_perturbed_utterances(app: FastAPI, monkeypatch):
...
+    request = [{"persistent_id": 0, "data_action": "remove"}]
+    resp = client.post("/dataset_splits/eval/utterances", json=request)
+    assert resp.status_code == HTTP_200_OK, resp.text
+    assert resp.json()[0] == {to_camel_case(k): v for k, v in request[0].items()}
+
+    # Reset tag to NO_ACTION
+    request = [{"persistent_id": 0, "data_action": "NO_ACTION"}]
+    resp = client.post("/dataset_splits/eval/utterances", json=request)
+    assert resp.status_code == HTTP_200_OK, resp.text
+    assert resp.json()[0] == {to_camel_case(k): v for k, v in request[0].items()}
```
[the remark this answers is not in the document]
gabegma: Ah! Thanks!

PR #435, opened 2023-02-10
Compute syntax smart tags for utterances with multiple sentences.

### LINKED ISSUE #434: Convert short_sentence and long_sentence to short_utterance and long_utterance.

gabegma:
Resolve #434 

## Description:
*  This will make it easier for datasets with long utterances (with multiple sentences) to have meaningful syntax smart tags.
* This will enable us to display the long/short utterances on the token count plot.
* I took this opportunity to refactor the syntax tagging tests, so they test all results (they also weren't working because of using `all` instead of `any`.)

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/dataset_analysis/syntax_tagging.py
+++ b/azimuth/modules/dataset_analysis/syntax_tagging.py
@@ -58,29 +58,27 @@ class SyntaxTaggingModule(DatasetResultModule[SyntaxConfig]):
             # Remove punctuation for word count and smart tags
             tokens = [token.text for token in doc if not token.is_punct]
 
+            if len(tokens) >= syntax_options.long_utterance_min_word:
+                tag[SmartTag.long] = True
+            if len(tokens) <= syntax_options.short_utterance_max_word:
+                tag[SmartTag.short] = True
+
+            sub_toks = [tok for tok in doc if (tok.dep_ in syntax_options.subj_tags)]
+            obj_toks = [tok for tok in doc if (tok.dep_ in syntax_options.obj_tags)]
+            vrb_toks = [tok for tok in doc if (tok.pos_ in self.verb_tags)]
+            if not sub_toks:
+                tag[SmartTag.no_subj] = True
+            if not obj_toks:
+                tag[SmartTag.no_obj] = True
+            if not vrb_toks:
+                tag[SmartTag.no_verb] = True
+
             # Some issues occur with other languages such as french if using doc.sents directly.
             # Hence, we use an English sentencizer that seems to work better for similar languages.
             doc_sentencizer_en = self.spacy_sentencizer_en(clean_utterance(utterance))
             sentence_count = len(list(doc_sentencizer_en.sents))
-
             if sentence_count > 1:
-                # if an utterance has more than one sentence, no other tags are added.
                 tag[SmartTag.multi_sent] = True
-            else:
-                if len(tokens) >= syntax_options.long_sentence_min_word:
-                    tag[SmartTag.long] = True
-                if len(tokens) <= syntax_options.short_sentence_max_word:
-                    tag[SmartTag.short] = True
-
-                sub_toks = [tok for tok in doc if (tok.dep_ in syntax_options.subj_tags)]
-                obj_toks = [tok for tok in doc if (tok.dep_ in syntax_options.obj_tags)]
-                vrb_toks = [tok for tok in doc if (tok.pos_ in self.verb_tags)]
-                if not sub_toks:
-                    tag[SmartTag.no_subj] = True
-                if not obj_toks:
-                    tag[SmartTag.no_obj] = True
-                if not vrb_toks:
-                    tag[SmartTag.no_verb] = True
 
             adds = {DatasetColumn.word_count: len(tokens)}
             records.append(TaggingResponse(tags=tag, adds=adds))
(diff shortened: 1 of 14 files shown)


azimuth/modules/dataset_analysis/syntax_tagging.py
```diff
@@ -58,29 +58,27 @@ def compute(self, batch: Dataset) -> List[TaggingResponse]:  # type: ignore
...
+                tag[SmartTag.no_subj] = True
+            if not obj_toks:
+                tag[SmartTag.no_obj] = True
+            if not vrb_toks:
+                tag[SmartTag.no_verb] = True
+
             # Some issues occur with other languages such as french if using doc.sents directly.
             # Hence, we use an English sentencizer that seems to work better for similar languages.
             doc_sentencizer_en = self.spacy_sentencizer_en(clean_utterance(utterance))
             sentence_count = len(list(doc_sentencizer_en.sents))
```
JosephMarinier, line 79: Not new, nor really important, but I thought you'd find it interesting: I just found out that the fastest way to count the number of items in an iterator:
```suggestion
            sentence_count = sum(1 for _ in doc_sentencizer_en.sents))
```
It avoids instantiating and populating a list, just to get its length.

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
```diff
@@ -21,50 +23,46 @@ def test_syntax_tagging(tiny_text_config):
         ],
         "label": [0, 1, 0, 1],
     }
-    json_output = mod.compute(batch)
-    assert len(json_output) == 4
-    assert json_output[0].tags[SmartTag.short] and json_output[0].tags[SmartTag.no_subj]
-    assert not all(
```
JosephMarinier, line 27: Good catch on the `not all(`s! 👍

azimuth/modules/dataset_analysis/syntax_tagging.py
gabegma: Interesting! I do think it's harder to read though, I would keep the `len` given the list here won't be huge.

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
```diff
@@ -2,16 +2,18 @@
...
 
-def test_syntax_tagging(tiny_text_config):
-    mod = SyntaxTaggingModule(
-        DatasetSplitName.eval,
-        tiny_text_config,
-    )
+def verify_syntax_results(json_output, smart_tags_per_idx, token_count_per_idx):
+    for idx, res in enumerate(json_output):
+        assert all(v for k, v in res.tags.items() if k in smart_tags_per_idx[idx])
+        assert not any(v for k, v in res.tags.items() if k not in smart_tags_per_idx[idx])
```
JosephMarinier, line 12: Would that be equivalent? The error would be a lot clearer, with the actual values and not just `False`.
```suggestion
        assert list(res.tags.values()) == [k in smart_tags_per_idx[idx] for k in res.tags.keys()]
```

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
```diff
@@ -2,16 +2,18 @@
...
+        assert res.adds[DatasetColumn.word_count] == token_count_per_idx[idx]
```
JosephMarinier, line 13: In fact, instead of `enumerate()` and using `idx`, you could `zip()`.
```suggestion
    for res, smart_tags, token_count in zip(json_output, smart_tags_per_idx, token_count_per_idx):
        assert list(res.tags.values()) == [k in smart_tags for k in res.tags.keys()]
        assert res.adds[DatasetColumn.word_count] == token_count
```
It could even be `zip_longest()` to be more robust if the response is shorter than the expected lists.

JosephMarinier (approved):
Cool! And good catch! 👍 I just have small ideas to improve the tests.

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
gabegma: I thought about the zip, but I didn't see any reason why it would be preferable. Could you please tell me more about why you prefer it?
For `zip_longest()`, I think that it should break if a list is shorter.

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
gabegma: Looks equivalent to me, sounds good! Good idea for the error message.

tests/test_modules/test_dataset_analysis/test_syntax_tagging.py
JosephMarinier: Exactly. With `zip_longest()`, if any one list is too short, it will be "padded" with `None`, ensuring the `assert` still happens. Right now, if `json_output` is an empty list, nothing will get tested. That is why I would suggest the method with `zip_longest`.

gabegma: But we do assert that `json_output` has the right length just before this function so that it is never empty, right? I can make the change, I'm just a bit puzzled still. Reading this code, I wouldn't understand why we use 'zip_longest' given that the 3 lists should be equal for the test/function to make sense.

JosephMarinier: Correct. So it's more for developer safety, in the event the test changes later. Right now, `verify_syntax_results()` is a reusable function, but nothing forces the developer to make sure the 3 lengths are equal before calling it. If someone adds a new test and reuses `verify_syntax_results()`, one could think it will check all that based on `verify_syntax_results()`'s name and forget to check the lengths before calling. But I agree it's not necessary. It should never pose a problem as long as we don't change the test.

gabegma: Ok I get it now, thanks!!

PR #437, opened 2023-02-13
Add modal to config page

### LINKED ISSUE #258: Settings icon should be able to close the settings page
**Is your feature request related to a problem? Please describe.**
When we click on the settings icon on the nav. We open a new settings page. And to go back to the previous page I cannot click on the same settings button as of today. I need to click on the dashboard link on the header to go back. But this takes me to the home page. 

**Describe the solution you'd like**
We need to be able to open the settings and then close it using the same settings icon. This will reduce the number of operations the user needs to do to navigate back and also we need to make sure that we go to the same previous page rather than the homepage.

**Describe alternatives you've considered**
None at the moment

**Additional context**
NIL

nandhinibsn:
Resolve #258 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/App.tsx
+++ b/webapp/src/App.tsx
@@ -84,11 +84,6 @@ export default class App extends React.Component<Props> {
                                 <Dashboard />
                               </BasicLayout>
                             </Route>
-                            <Route path="/:jobId/settings" exact>
-                              <BasicLayout maxWidth="md">
-                                <Settings />
-                              </BasicLayout>
-                            </Route>
                             <Route
                               path="/:jobId/dataset_splits/:datasetSplitName/class_overlap"
                               exact
--- a/webapp/src/App.tsx
+++ b/webapp/src/App.tsx
@@ -30,7 +30,6 @@ import PerturbationTestingSummary from "pages/PerturbationTestingSummary";
 import Threshold from "pages/Threshold";
 import { DatasetSplitName } from "types/api";
 import Exploration from "pages/Exploration";
-import Settings from "pages/Settings";
 import NotFound from "pages/NotFound";
 import ClassOverlap from "pages/ClassOverlap";
 import SmartTags from "pages/SmartTags";
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -1,5 +1,12 @@
-import { Settings } from "@mui/icons-material";
-import { Box, Breadcrumbs, IconButton, Link, Typography } from "@mui/material";
+import { Settings as SettingsIcon } from "@mui/icons-material";
+import {
+  Box,
+  Breadcrumbs,
+  IconButton,
+  Link,
+  Modal,
+  Typography,
+} from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import useQueryState from "hooks/useQueryState";
 import React from "react";
@@ -14,6 +21,7 @@ import { DatasetSplitName } from "types/api";
 import { constructSearchString } from "utils/helpers";
 import HelpMenu from "components/HelpMenu";
 import PipelineSelect from "components/PipelineSelect";
+import Settings from "pages/Settings";
 
 const useStyles = makeStyles((theme) => ({
   jobHeader: {
@@ -26,6 +34,16 @@ const useStyles = makeStyles((theme) => ({
   label: {
     fontWeight: "bold",
     marginRight: theme.spacing(1),
+  },
+  modal: {
+    position: "absolute",
+    width: "70%",
+    height: "80%",
+    padding: "10px",
+    backgroundColor: theme.palette.background.paper,
+    top: "50%",
+    left: "50%",
+    transform: "translate(-50%, -50%)",
   },
 }));
 
@@ -79,7 +97,7 @@ const PageHeader = () => {
       })}`
     );
   };
-
+  const [openConfigModal, setOpenConfigModal] = React.useState(false);
   const dashboardPathname = `/${jobId}`;
 
   const isDashboard = location.pathname === dashboardPathname;
@@ -174,10 +192,9 @@ const PageHeader = () => {
               </>
             )}
             <IconButton
-              component={RouterLink}
               size="small"
               color="primary"
-              to={`/${jobId}/settings${searchString}`}
+              onClick={() => setOpenConfigModal(!openConfigModal)}
               sx={{
                 padding: 0,
                 "&:hover > svg": {
@@ -186,10 +203,22 @@ const PageHeader = () => {
                 },
               }}
             >
-              <Settings />
+              <SettingsIcon />
             </IconButton>
             <HelpMenu />
           </Box>
+          <Modal
+            open={openConfigModal}
+            onClose={(_, reason) =>
+              reason === "backdropClick" || reason === "escapeKeyDown"
+                ? setOpenConfigModal(true)
+                : setOpenConfigModal(false)
+            }
+          >
+            <Box className={classes.modal}>
+              <Settings setOpen={setOpenConfigModal} />
+            </Box>
+          </Modal>
         </div>
       )}
     </div>
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -2,9 +2,9 @@ import { Settings as SettingsIcon } from "@mui/icons-material";
 import {
   Box,
   Breadcrumbs,
+  Dialog,
   IconButton,
   Link,
-  Modal,
   Typography,
 } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
@@ -34,16 +34,6 @@ const useStyles = makeStyles((theme) => ({
   label: {
     fontWeight: "bold",
     marginRight: theme.spacing(1),
-  },
-  modal: {
-    position: "absolute",
-    width: "70%",
-    height: "80%",
-    padding: "10px",
-    backgroundColor: theme.palette.background.paper,
-    top: "50%",
-    left: "50%",
-    transform: "translate(-50%, -50%)",
   },
 }));
 
@@ -194,7 +184,7 @@ const PageHeader = () => {
             <IconButton
               size="small"
               color="primary"
-              onClick={() => setOpenConfigModal(!openConfigModal)}
+              onClick={() => setOpenConfigModal(true)}
               sx={{
                 padding: 0,
                 "&:hover > svg": {
@@ -207,18 +197,30 @@ const PageHeader = () => {
             </IconButton>
             <HelpMenu />
           </Box>
-          <Modal
+          <Dialog
+            aria-labelledby="config-dialog-title"
+            maxWidth="md"
+            scroll="paper"
             open={openConfigModal}
             onClose={(_, reason) =>
               reason === "backdropClick" || reason === "escapeKeyDown"
                 ? setOpenConfigModal(true)
                 : setOpenConfigModal(false)
             }
+            sx={{
+              "& .MuiDialogContent-root": {
+                padding: (theme) => theme.spacing(1),
+              },
+              "& .MuiDialogActions-root": {
+                display: "flex",
+                justifyContent: "space-between",
+                paddingX: (theme) => theme.spacing(2),
+                gap: 1,
+              },
+            }}
           >
-            <Box className={classes.modal}>
-              <Settings setOpen={setOpenConfigModal} />
-            </Box>
-          </Modal>
+            <Settings setOpen={setOpenConfigModal} />
+          </Dialog>
         </div>
       )}
     </div>
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -201,12 +201,9 @@ const PageHeader = () => {
             aria-labelledby="config-dialog-title"
             maxWidth="md"
             scroll="paper"
+            disableEscapeKeyDown
             open={openConfigModal}
-            onClose={(_, reason) =>
-              reason === "backdropClick" || reason === "escapeKeyDown"
-                ? setOpenConfigModal(true)
-                : setOpenConfigModal(false)
-            }
+            onClose={(_, reason) => setOpenConfigModal(false)}
             sx={{
               "& .MuiDialogContent-root": {
                 padding: (theme) => theme.spacing(1),
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -202,19 +202,8 @@ const PageHeader = () => {
             maxWidth="md"
             scroll="paper"
             open={openConfigModal}
-            sx={{
-              "& .MuiDialogContent-root": {
-                padding: (theme) => theme.spacing(1),
-              },
-              "& .MuiDialogActions-root": {
-                display: "flex",
-                justifyContent: "space-between",
-                paddingX: (theme) => theme.spacing(2),
-                gap: 1,
-              },
-            }}
           >
-            <Settings setOpen={setOpenConfigModal} />
+            <Settings onClose={setOpenConfigModal} />
           </Dialog>
         </div>
       )}
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -202,7 +202,7 @@ const PageHeader = () => {
             maxWidth="md"
             open={openConfigModal}
           >
-            <Settings onClose={setOpenConfigModal} />
+            <Settings onClose={() => setOpenConfigModal(false)} />
           </Dialog>
         </div>
       )}
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -200,6 +200,7 @@ const PageHeader = () => {
           <Dialog
             aria-labelledby="config-dialog-title"
             maxWidth="md"
+            fullWidth
             open={openConfigModal}
           >
             <Settings onClose={() => setOpenConfigModal(false)} />
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -200,7 +200,6 @@ const PageHeader = () => {
           <Dialog
             aria-labelledby="config-dialog-title"
             maxWidth="md"
-            scroll="paper"
             open={openConfigModal}
           >
             <Settings onClose={setOpenConfigModal} />
--- a/webapp/src/components/PageHeader.tsx
+++ b/webapp/src/components/PageHeader.tsx
@@ -201,9 +201,7 @@ const PageHeader = () => {
             aria-labelledby="config-dialog-title"
             maxWidth="md"
             scroll="paper"
-            disableEscapeKeyDown
             open={openConfigModal}
-            onClose={(_, reason) => setOpenConfigModal(false)}
             sx={{
               "& .MuiDialogContent-root": {
                 padding: (theme) => theme.spacing(1),
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -206,10 +206,10 @@ const NumberField: React.FC<
 };
 
 type props = {
-  setOpen: (close: boolean) => void;
+  onClose: () => void;
 };
 
-const Settings: React.FC<props> = ({ setOpen }) => {
+const Settings: React.FC<props> = ({ onClose }) => {
   const { jobId } = useParams<{ jobId: string }>();
   const [language, setLanguage] = React.useState<
     SupportedLanguage | undefined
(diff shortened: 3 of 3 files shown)


webapp/src/components/PageHeader.tsx
```diff
@@ -186,10 +186,11 @@ const PageHeader = () => {
                 },
               }}
             >
-              <Settings />
+              <SettingsIcon />
             </IconButton>
             <HelpMenu />
           </Box>
+          <Settings open={openConfig} onCloseDialog={setOpenConfig} />
```
JosephMarinier, line 193: I liked the flexibility that your initial approach of defining the modal here (as opposed to inside `Settings`) provided. It was enabling us to use `Settings` both in a modal as well as on its own page (which might still be useful in the future as a landing page if the config is not yet defined).
What made yo change your mind?

webapp/src/components/DialogLayout.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React from "react";
+
+type Props = DialogProps & {
+  open: boolean;
+  onClose: (close: boolean) => void;
+  titleSection: React.ReactElement;
+  content?: React.ReactElement | undefined;
+  action?: React.ReactElement | undefined;
+};
+const DialogLayout: React.FC<Props> = ({
```
JosephMarinier, line 17: What is the value of `DialogLayout`? To me it looks very similar to `Dialog`, but with less flexibility.

webapp/src/pages/Settings.tsx
```diff
@@ -637,90 +651,188 @@ const Settings: React.FC = () => {
...
+            tasks.
+          </Typography>
+          <Tooltip title="close" placement="bottom">
+            <IconButton
+              size="small"
+              color="primary"
+              onClick={() =>
+                Object.keys(partialConfig).length > 0 || isUpdatingConfig
+                  ? showModal(true)
+                  : onCloseDialog(false)
```
JosephMarinier, line 681: Instead of a second modal, I think the simplest and most standard way to prompt for confirmation like that is to use `window.confirm()`, with something like this:
```suggestion
                (Object.keys(partialConfig).length === 0 ||
                  window.confirm(DISCARD_CONFIRM_MESSAGE)) &&
                onCloseDialog(false)
```

webapp/src/components/PageHeader.tsx
nandhinibsn: The purpose is to keep the logic and modal inside settings to reduce communications between parent and child components other than open and closing the dialog. Additionally, it is easier to read the design of config UI in the settings.tsx.

webapp/src/components/DialogLayout.tsx
nandhinibsn: The only purpose is to reduce the number of lines of code.

webapp/src/pages/Settings.tsx
nandhinibsn: True. "window. confirm" has a basic design. But we can consider using it. Thanks

webapp/src/components/PageHeader.tsx
```diff
@@ -174,10 +175,9 @@ const PageHeader = () => {
               </>
             )}
             <IconButton
-              component={RouterLink}
               size="small"
               color="primary"
-              to={`/${jobId}/settings${searchString}`}
+              onClick={() => setOpenConfig(!openConfig)}
```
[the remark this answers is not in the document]
nandhinibsn: It should behave like a toggle like earlier code. But, something is blocking that logic and I am debugging it. I will be able to fix it asap.

webapp/src/components/DialogLayout.tsx
nandhinibsn: May I know why do you think it's not flexible? we can customize the "DialogLayout" component as it accepts DialogProps too.

JosephMarinier: I meant less flexible as it needs exactly one `DialogTitle`, one `DialogContent` and one `DialogActions`. I notice you therefore have to pass it some empty fragments `<></>`, which is confusing.

webapp/src/pages/Settings.tsx
```diff
@@ -637,55 +648,83 @@ const Settings: React.FC = () => {
...
+          </Typography>
+          <Tooltip title="close" placement="bottom">
+            <IconButton
+              size="small"
+              color="primary"
+              sx={{ padding: 0 }}
+              onClick={() =>
+                Object.keys(partialConfig).length > 0 || isUpdatingConfig
+                  ? isUpdatingConfig
+                    ? window.confirm(CONFIG_UPDATE_MESSAGE)
```
JosephMarinier, line 667: The `confirm` doesn't seem to be appropriate here, since you ignore whether the user clicks on OK or Cancel. How about an `alert` that just has an `OK` button?
```suggestion
                    ? window.alert(CONFIG_UPDATE_MESSAGE)
```

webapp/src/App.tsx
```diff
@@ -84,11 +84,6 @@ export default class App extends React.Component<Props> {
                                 <Dashboard />
                               </BasicLayout>
                             </Route>
-                            <Route path="/:jobId/settings" exact>
-                              <BasicLayout maxWidth="md">
-                                <Settings />
```
JosephMarinier, line 89: As you can see from the warning in the terminal, you forgot to remove the `import Settings`.

webapp/src/pages/Settings.tsx
```diff
@@ -637,55 +648,83 @@ const Settings: React.FC = () => {
...
+                  ? isUpdatingConfig
+                    ? window.alert(CONFIG_UPDATE_MESSAGE)
+                    : Object.keys(partialConfig).length > 0 &&
+                      window.confirm(
+                        "Are you sure you want to discard all your changes?"
+                      )
+                    ? setOpen(false)
+                    : setOpen(true)
+                  : setOpen(false)
+              }
```
JosephMarinier, line 675: I find this very hard to read, with `Object.keys(partialConfig).length > 0` and `isUpdatingConfig` being checked twice each, and `setOpen()` being called 3 times. What do you think of this instead?
```suggestion
              onClick={() => {
                if (isUpdatingConfig) {
                  window.alert(CONFIG_UPDATE_MESSAGE);
                } else if (
                  Object.keys(partialConfig).length === 0 ||
                  !window.confirm(
                    "Are you sure you want to discard all your changes?"
                  )
                ) {
                  setOpen(false);
                }
              }}
```

webapp/src/App.tsx
nandhinibsn: Ohh... sorry. I removed unused imports and no more warning on the console and terminal

webapp/src/pages/Settings.tsx
nandhinibsn: So, this means "!window.confirm(
                    "Are you sure you want to discard all your changes?"
                  )" if the user cancel (which is false), then close the modal, right?

JosephMarinier: Oh sorry, the `!` was a mistake. I edited my suggestion.

nandhinibsn: Right. Thank you for clarifying it. I had confusion about using !.

webapp/src/components/PageHeader.tsx
```diff
@@ -186,10 +193,34 @@ const PageHeader = () => {
...
+          <Dialog
+            aria-labelledby="config-dialog-title"
+            maxWidth="md"
+            scroll="paper"
+            open={openConfigModal}
+            onClose={(_, reason) =>
+              reason === "backdropClick" || reason === "escapeKeyDown"
+                ? setOpenConfigModal(true)
+                : setOpenConfigModal(false)
+            }
```
JosephMarinier, line 209: Following up on [this](https://github.com/ServiceNow/azimuth/pull/437#discussion_r1124014825).
You also inverted the `true` and `false`, so the dialog box doesn't close now.
I insist on simplifying this to
```suggestion
            onClose={() => setOpenConfigModal(false)}
```

nandhinibsn: Okay, we need to stay on the modal for both reasons (backdrop click and esckeydown) when there are ongoing changes or updates. Users may accidentally do backdrop click or esc.

webapp/src/pages/Settings.tsx
```diff
@@ -637,55 +648,83 @@ const Settings: React.FC = () => {
...
+                  Object.keys(partialConfig).length === 0 ||
+                  window.confirm(
+                    "Are you sure you want to discard all your changes?"
+                  )
+                ) {
+                  setOpen(false);
+                }
+              }}
+            >
+              <XIcon />
```
JosephMarinier, line 677: I think the `Close` icon from MUI would be more semantically accurate (given it is the same icon).
```suggestion
              <Close />
```

nandhinibsn: Okay, I used XICon as we have used in other places of Azimuth.

webapp/src/components/PageHeader.tsx
JosephMarinier: If that's what you want, you should simply not define `onClose`.
```suggestion
```

webapp/src/pages/Settings.tsx
JosephMarinier: The `XIcon` is only used for the outcome icon (yellow or red), along with `CheckIcon` (green). It is never used as a button to close something. Do you see the distinction?

webapp/src/components/PageHeader.tsx
nandhinibsn: I can use "disableEscapeKeyDown" to stop calling on close if the ESC key is pressed. But, I need to define inside onClose to control the backdrop click. Would you happen to have any other suggestions to handle this? Do you feel this check is optional?

webapp/src/pages/Settings.tsx
nandhinibsn: Okay, I see now. Thanks for bringing up the difference in usage.

webapp/src/components/PageHeader.tsx
nandhinibsn: Thanks. I did not define onClose and it handled both reasons with ease.

webapp/src/pages/Settings.tsx
```diff
@@ -199,7 +205,11 @@ const NumberField: React.FC<
   );
 };
 
-const Settings: React.FC = () => {
+type props = {
+  setOpen: (close: boolean) => void;
```
JosephMarinier, line 209: Detail, but I notice `setOpen` is only always called with `false`. That makes sense since the buttons in the config can only close de config. Inspired by the similar function on `Dialog`, I would call it `onClose`. What do you think?
```suggestion
  onClose: () => void;
```

webapp/src/pages/Settings.tsx
```diff
@@ -637,55 +647,83 @@ const Settings: React.FC = () => {
...
+          </Typography>
+          <Tooltip title="close" placement="bottom">
+            <IconButton
+              size="small"
+              color="primary"
+              sx={{ padding: 0 }}
+              onClick={() => {
+                if (isUpdatingConfig) {
+                  window.alert(CONFIG_UPDATE_MESSAGE);
+                } else if (
```
JosephMarinier, line 666: Alternatively, we could disable the button while the config is updating. What do you think?
```suggestion
              disabled={isUpdatingConfig}
              onClick={() => {
                if (
```

webapp/src/pages/Settings.tsx
```diff
@@ -637,55 +647,83 @@ const Settings: React.FC = () => {
...
-          link="reference/configuration/analyses/"
+    <>
+      <DialogTitle id="config-dialog-title">
+        <Box display="flex" justifyContent="space-between">
+          <Typography variant="subtitle1">
+            View and edit certain fields from your config file. Once your
+            changes are saved, expect some delays for recomputing the affected
+            tasks.
+          </Typography>
+          <Tooltip title="close" placement="bottom">
```
JosephMarinier, line 658: I find this tooltip doesn't bring much information. The X icon is pretty self-explanatory. Do you think we should remove it?

webapp/src/pages/Settings.tsx
```diff
@@ -637,55 +647,83 @@ const Settings: React.FC = () => {
...
+            overflow: "auto",
+            [`& .${formControlLabelClasses.labelPlacementStart}`]: {
+              justifyContent: "flex-end",
+              marginLeft: 0,
+            },
+            [`& .${formGroupClasses.root}`]: { marginX: 2, marginBottom: 2 },
+            [`& .fixedWidthInput .${inputClasses.root}`]: { width: "12ch" },
+            [`& .${inputClasses.input}`]: { fontSize: 14, padding: 0 },
+            [`& .${inputLabelClasses.root}`]: { fontWeight: "bold" },
+          }}
```
JosephMarinier, line 695: A couple of things:
- The `flex: 1` has no effect since its parent `DialogContent` is a `display: block`.
- The `overflow: "auto"` is already handled by the parent `DialogContent`.
- The `DialogContent` already has some padding, which I find enough.

So, I think we can remove that `Box` and move the `sx` prop to the `DialogContent`. Does that make sense?
```suggestion
      <DialogContent
        dividers
        sx={{
          [`& .${formControlLabelClasses.labelPlacementStart}`]: {
            justifyContent: "flex-end",
            marginLeft: 0,
          },
          [`& .${formGroupClasses.root}`]: { marginX: 2, marginBottom: 2 },
          [`& .fixedWidthInput .${inputClasses.root}`]: { width: "12ch" },
          [`& .${inputClasses.input}`]: { fontSize: 14, padding: 0 },
          [`& .${inputLabelClasses.root}`]: { fontWeight: "bold" },
        }}
```

webapp/src/components/PageHeader.tsx
```diff
@@ -186,10 +193,29 @@ const PageHeader = () => {
...
           </Box>
+          <Dialog
+            aria-labelledby="config-dialog-title"
+            maxWidth="md"
+            scroll="paper"
+            open={openConfigModal}
+            sx={{
+              "& .MuiDialogContent-root": {
+                padding: (theme) => theme.spacing(1),
+              },
```
JosephMarinier, line 208: In combination with my previous comment, if you remove this, the padding on the sides defaults to 3, which is exactly the sum of the two paddings you had specified. Perfect!
```suggestion
```

webapp/src/components/PageHeader.tsx
```diff
@@ -186,10 +193,29 @@ const PageHeader = () => {
...
+              "& .MuiDialogActions-root": {
+                display: "flex",
+                justifyContent: "space-between",
+                paddingX: (theme) => theme.spacing(2),
+                gap: 1,
+              },
```
JosephMarinier, line 214: I think this should be moved to an `sx` prop on the `DialogActions` component. Also, `display: "flex"` and `gap: 1` don't have any effect, so they can be cleaned up. I also like the default padding, so up to you if you want this overwrite or not.

webapp/src/pages/Settings.tsx
nandhinibsn: Right!.. Thanks :)

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, I agree too

webapp/src/pages/Settings.tsx
nandhinibsn: Sure. good one. How about we disable the discard button too?

JosephMarinier: Sounds good to me!

webapp/src/components/PageHeader.tsx
```diff
@@ -186,10 +193,18 @@ const PageHeader = () => {
...
             </IconButton>
             <HelpMenu />
           </Box>
+          <Dialog
+            aria-labelledby="config-dialog-title"
+            maxWidth="md"
+            scroll="paper"
+            open={openConfigModal}
+          >
+            <Settings onClose={setOpenConfigModal} />
```
JosephMarinier, line 206: Don't you have an error, here?

webapp/src/pages/Settings.tsx
```diff
@@ -679,13 +709,18 @@ const Settings: React.FC = () => {
...
           {getAnalysesCustomizationSection()}
         </AccordionLayout>
-      </Paper>
-      <Box display="flex" justifyContent="space-between" paddingY={2}>
+      </DialogContent>
+      <DialogActions
+        sx={{
+          justifyContent: "space-between",
+        }}
+      >
```
JosephMarinier, line 717: Detail, but that works on one line, to save a little bit of space.
```suggestion
      <DialogActions sx={{ justifyContent: "space-between" }}>
```

webapp/src/components/PageHeader.tsx
```diff
@@ -186,10 +193,18 @@ const PageHeader = () => {
...
             >
-              <Settings />
+              <SettingsIcon />
             </IconButton>
             <HelpMenu />
           </Box>
+          <Dialog
+            aria-labelledby="config-dialog-title"
+            maxWidth="md"
+            scroll="paper"
```
JosephMarinier, line 203: That's the default value, so I would remove it
```suggestion
```

webapp/src/pages/Settings.tsx
```diff
@@ -199,7 +205,11 @@ const NumberField: React.FC<
   );
 };
 
-const Settings: React.FC = () => {
+type props = {
```
JosephMarinier, line 208: Types are supposed to be capitalized.

webapp/src/components/PageHeader.tsx
JosephMarinier: I notice you moved `Dialog` out of `Settings`, but not `DialogTitle`, `DialogContent` and `DialogActions`, which makes `Settings` tightly coupled with the `Dialog` anyway. Unfortunately, this in-between approach seems like the worst of both worlds. I see that because of `DialogTitle`, `DialogContent` and `DialogActions`, it would be too hard to decouple `Settings` and `Dialog`. I say we forget that idea for now. I suggest moving `Dialog` back inside of `Settings`, with an `open` prop, as you did. It was good. 👍 I'm sorry for the back-and-forth I caused, and thank you for trying it out.

webapp/src/components/PageHeader.tsx
nandhinibsn: Yes, I do, as we pass the boolean value to it. I am working on it to fix that.

webapp/src/pages/Settings.tsx
```diff
@@ -679,13 +709,18 @@ const Settings: React.FC = () => {
...
+        sx={{
+          justifyContent: "space-between",
+        }}
+      >
         <Button
           variant="contained"
           onClick={() => {
             setPartialConfig({});
             setLanguage(undefined);
+            onClose();
```
JosephMarinier, line 723: The `Discard` button should not close the config, right? It's only to reset the changes to the current config.

webapp/src/components/PageHeader.tsx
JosephMarinier: Oh! But the loading and error states... Those have to be rendered inside the `Dialog`. Ok, let's leave everything as it is now.

webapp/src/pages/Settings.tsx
```diff
@@ -637,13 +647,37 @@ const Settings: React.FC = () => {
...
+        <Box display="flex" justifyContent="space-between">
+          <Typography variant="subtitle1">
+            View and edit certain fields from your config file. Once your
+            changes are saved, expect some delays for recomputing the affected
+            tasks.
+          </Typography>
+          <IconButton
+            size="small"
+            color="primary"
+            sx={{ padding: 0 }}
```
JosephMarinier, line 661: Because of the `padding: 0`, the "touch ripple" effect on the button is not round anymore. I would remove the `sx={{ padding: 0 }}` to leave the default look of the button and add `alignItems="center"` on the flex `Box` to center the title on the button. Does that make sense?
```suggestion
        <Box display="flex" justifyContent="space-between" alignItems="center">
          <Typography variant="subtitle1">
            View and edit certain fields from your config file. Once your
            changes are saved, expect some delays for recomputing the affected
            tasks.
          </Typography>
          <IconButton
            size="small"
            color="primary"
```

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, True. I suggest disabling the discard while it's updating. So, the user cannot exit the modal, and no changes are available to discard.

webapp/src/pages/Settings.tsx
nandhinibsn: Okay, Understood. It's a nice trick. Thanks.

webapp/src/components/PageHeader.tsx
JosephMarinier: You just added back the unnecessary boolean argument that I asked you to remove [here](https://github.com/ServiceNow/azimuth/pull/437#discussion_r1131849905)? The right solution is rather as follows, just like you did for the `IconButton`'s `onClick`:
```suggestion
            <Settings onClose={() => setOpenConfigModal(false)} />
```

nandhinibsn: Yes, that error message due to that change was blocking me to apply other changes while I figure out the solution. So, I reverted back and working on it now.

nandhinibsn: Applied the changes to remove the boolean property in my recent commit. Thank you for the suggestion.

webapp/src/pages/Settings.tsx
JosephMarinier: Yes, that is good.
To be clear, I am not talking about "while it's updating" here.

nandhinibsn: So, do you mean to say, "we don't disable the discard button" when an update is in progress? 

If it's updating, the user can exit the modal by clicking the "discard" button. So, I mentioned it is essential to disable the button. Let me know what you think of it.

webapp/src/pages/Settings.tsx
```diff
@@ -716,11 +759,11 @@ const Settings: React.FC = () => {
             disabled={isUpdatingConfig}
             onClick={() => updateConfig({ jobId, body: partialConfig })}
           >
-            Apply
+            Apply and Close
```
JosephMarinier, line 762: We usually capitalize only the first word of our buttons. For example, `See more`, `View details` and `Export behavioral testing summary`. This is not a title.

webapp/src/pages/Settings.tsx
JosephMarinier: No. I'm suggesting this:
```suggestion
```

webapp/src/pages/Settings.tsx
```diff
@@ -716,11 +759,11 @@ const Settings: React.FC = () => {
             disabled={isUpdatingConfig}
             onClick={() => updateConfig({ jobId, body: partialConfig })}
```
JosephMarinier, line 760: Instead of `isUpdateSuccessful`, which is a little bit complicated with the `useEffect`, you can directly do this:
```suggestion
            onClick={() =>
              updateConfig({ jobId, body: partialConfig })
                .unwrap()
                .then(onClose)
            }
```

nandhinibsn: okay, cool. Added the suggestion.

webapp/src/pages/Settings.tsx
JosephMarinier: This was the last pending item, so I committed it in [Revert closing on discard](https://github.com/ServiceNow/azimuth/pull/437/commits/20a9519da30fc30193efff1328f51347b00835c4).

webapp/src/components/PageHeader.tsx
JosephMarinier: I took the time to do that change in [Put loading and error states inside the Dialog box](https://github.com/ServiceNow/azimuth/pull/437/commits/947424ea51434c33467535e8491c6e4fe7e25ec7). The diff in `Settings.tsx` seems overwhelming, but it's only moving code around.

JosephMarinier: Nevermind, I'll make a separate PR. I still have an issue to fix with my implementation.

JosephMarinier (approved):
👍

nandhinibsn:
Thank you for everything :)

PR #440, opened 2023-02-15
Add truncation=True to HFPipelines

### LINKED ISSUE #439: Tensor size mismatch with HF pipeline
Hi,

I'm eager to try Azimuth as the tool looks great, however I've been running into some issues setting it up. I'm trying to run Azimuth with a simple HF binary classification pipeline. My data contains very long texts and causes errors when I try to run Azimuth with it, I get
```
backend_1  | azimuth.utils.logs.MultipleExceptions: We found the following exceptions:  
backend_1  | 	The expanded size of the tensor (972) must match the existing size (514) at non-singleton dimension 1.  Target sizes: [10, 972].  Tensor sizes: [1, 514]
```

I started with the setup indicated in the "Run on your use case" of the getting started guide which did not work for me and then tried to switch to a custom load_pipeline function, but I didn't find a way to fix padding and truncation settings at pipeline definition time. This can be done when calling the inference pipeline but not at definition time as far as I know.

So now I didn't find a way to define a custom caller to the inference pipeline, but I feel like I'm missing something where I could have fed  `truncate=True` instead of jumping through the hoops of redefining everything?

Any pointers would be appreciated.

Best,
Vil

Dref360:
Resolve #439 

## Description:

For long utterances, we need to add truncation=True.

For saliency, we will only show the processed tokens, the utterance will be cut.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/model_contracts/hf_text_classification.py
+++ b/azimuth/modules/model_contracts/hf_text_classification.py
@@ -74,7 +74,12 @@ class HFTextClassificationModule(TextClassificationModule):
                 predictions = np.stack(
                     [
                         self.extract_probs_from_output(
-                            model(utterances, num_workers=0, batch_size=self.config.batch_size)
+                            model(
+                                utterances,
+                                num_workers=0,
+                                batch_size=self.config.batch_size,
+                                truncation=True,
+                            )
                         )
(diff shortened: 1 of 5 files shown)


gabegma (approved):
Awesome, thank you!!

PR #420, opened 2023-02-02
Test for performance analysis comparison table

nandhinibsn:
Resolve #

## Checklist:

You should check all boxes before the PR is ready. If a box does not apply, check it to acknowledge it.

* [x] **ISSUE NUMBER.** You linked the issue number (Ex: Resolve #XXX).
* [x] **PRE-COMMIT.** You ran pre-commit on all commits, or else, you
  ran `pre-commit run --all-files` at the end.
* [x] **USER CHANGES.** The changes are added to CHANGELOG.md and the documentation, if they impact
  our users.
* [x] **DEV CHANGES.**
    * Update the documentation if this PR changes how to develop/launch on the app.
    * Update the `README` files and our wiki for any big design decisions, if relevant.
    * Add unit tests, docstrings, typing and comments for complex sections.

PR #441, opened 2023-02-15
UI test script for class overlap table

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/ClassOverlapTable.test.tsx
+++ b/webapp/src/components/ClassOverlapTable.test.tsx
@@ -74,8 +74,8 @@ describe("ClassOverlapTable", () => {
   it("should not display the Footer component if the number of rows is less than initial number", async () => {
     renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
     await waitFor(() => {
-      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row");
-      expect(screen.queryByText(/See more/i)).toBeNull();
+      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row"); // This test case ensures API responses are fetched to verify the "see more" displayed on the page.
     });
+    expect(screen.queryByText(/See more/i)).toBeNull();
   });
 });
--- a/webapp/src/components/ClassOverlapTable.test.tsx
+++ b/webapp/src/components/ClassOverlapTable.test.tsx
@@ -1,4 +1,4 @@
-import { screen, within } from "@testing-library/react";
+import { screen, waitFor, within } from "@testing-library/react";
 import { renderWithRouterAndRedux } from "mocks/utils";
 import ClassOverlapTable from "./ClassOverlapTable";
 import { AvailableDatasetSplits } from "types/api";
@@ -71,8 +71,11 @@ describe("ClassOverlapTable", () => {
     ).toBeNull();
   });
 
-  it("should not display the Footer component if the number of rows is less than initial number", () => {
+  it("should not display the Footer component if the number of rows is less than initial number", async () => {
     renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
-    expect(screen.queryByText(/See more/)).toBeNull();
+    await waitFor(() => {
+      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row");
+      expect(screen.queryByText(/See more/i)).toBeNull();
+    });
   });
 });
--- a/webapp/src/components/ClassOverlapTable.test.tsx
+++ b/webapp/src/components/ClassOverlapTable.test.tsx
@@ -74,7 +74,8 @@ describe("ClassOverlapTable", () => {
   it("should not display the Footer component if the number of rows is less than initial number", async () => {
     renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
     await waitFor(() => {
-      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row"); // This test case ensures API responses are fetched to verify the "see more" displayed on the page.
+      // Wait for the API data to be loaded before verifying that there is no "See more" button.
+      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row");
     });
     expect(screen.queryByText(/See more/i)).toBeNull();
   });
--- a/webapp/src/components/ClassOverlapTable.test.tsx
+++ b/webapp/src/components/ClassOverlapTable.test.tsx
@@ -70,7 +70,7 @@ describe("ClassOverlapTable", () => {
     ).toBeNull();
   });
 
-  it("should not display the Footer component if the number of rows is lesser than initial number", async () => {
+  it("should not display the Footer component if the number of rows is less than initial number", async () => {
     renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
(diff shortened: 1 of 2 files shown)


webapp/src/components/ClassOverlapTable.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const columnHeaderDescription = [
+      "The class label of the samples being analyzed.",
+      "The class that the source class may look like. For pipeline confusion, this is the prediction.",
+      "Class overlap measures the extent to which source class samples are semantically similar to target class samples, in the training data. The score comes from both the proportion of samples and their degree of similarity.",
+      "Pipeline confusion indicates whether source class samples in the evaluation set are predicted to be in the target class.",
+      "Percent of source class samples that semantically overlap the target class (all in the training set).",
+    ];
+    const actualColumnHeaders = screen.getAllByRole("columnheader");
+    expectedColumnHeaders.forEach((name, index) => {
+      expect(actualColumnHeaders[index].textContent).toEqual(name);
```
JosephMarinier, line 48: Detail: should we use this?
```suggestion
      expect(actualColumnHeaders[index]).toHaveTextContent(name);
```

webapp/src/components/ClassOverlapTable.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+        columnHeaderDescription[index]
+      );
+    });
+  });
+
+  it("should display the default sort as desc for colum header 'Semantic Overlap Score'", () => {
+    renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
+    const header = screen.getByRole("columnheader", {
+      name: /Class overlap/,
+    });
```
JosephMarinier, line 59: It is not obvious that the `name: /Class overlap/` corresponds to the `Semantic Overlap Score` column header, since it is the tooltip. Should we add something like `expect(header).toHaveTextContent("Semantic Overlap Score");`? Also, the `getByRole()` fits on one line:
```suggestion
    const header = screen.getByRole("columnheader", { name: /Class overlap/ });
    expect(header).toHaveTextContent("Semantic Overlap Score");
```

webapp/src/components/ClassOverlapTable.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    renderClassOverlapTable(
+      { train: true, eval: true },
+      { pipelineIndex: undefined }
+    );
+    expect(
+      screen.queryByRole("columnheader", { name: /Pipeline confusion/ })
+    ).toBeNull();
+  });
+
+  it("should not display the Footer component if the number of rows is lesser than initial number", async () => {
```
JosephMarinier, line 73:
```suggestion
  it("should not display the Footer component if the number of rows is less than initial number", async () => {
```

webapp/src/components/ClassOverlapTable.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
+    await waitFor(() => expect(screen.queryByText(/See more/)).toBeNull());
+  });
```
JosephMarinier, line 76: The `waitFor()` doesn't have any effect here, as `screen.queryByText(/See more/)` will be null from the start. So, no need for the `async`, and also there is a typo:
```suggestion
  it("should not display the Footer component if the number of rows is less than initial number", () => {
    renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
    expect(screen.queryByText(/See more/)).toBeNull();
  });
```

webapp/src/components/ClassOverlapTable.test.tsx
nandhinibsn: textContent does the exact match where [toHaveTextContent](https://github.com/testing-library/jest-dom#tohavetextcontent) needs regex to perform that. I am fine doing both. I just modified all the [places](https://github.com/ServiceNow/azimuth/pull/420/commits/676e106b55737fc6656283eb4096220e71b06dc3) that have "toHaveTextContent" to textContent.

JosephMarinier: Ah-ha! So it's equivalent to `.toContain()` instead of `.toEqual()`. That's a very good point! I think you're right to favor exact match, then. You can ignore my suggestion, then.

webapp/src/components/ClassOverlapTable.test.tsx
nandhinibsn: Yeah, true; it needs to be more readable about which column we are targeting to test. Here, the purpose is to check the aria-sort of the column. The Name attribute has stored the tooltip, So I handled the test case using the aria label.

webapp/src/components/ClassOverlapTable.test.tsx
nandhinibsn: Okay, I suggest using async, as we display the footer component based on the length of the row. Hence, we should wait for the API response to be fetched. The purpose of this test case is to check the display of the "see more" section after the API response. Ideally, it should be null because my mock API response has only 3 three records.

JosephMarinier: But here, you do NOT "wait for the API response to be fetched". I guess you should. But right now, the test `toBeNull()` will pass on the very first loop of `waitFor()`, and it won't wait for the API response to be fetched.

webapp/src/components/ClassOverlapTable.test.tsx
JosephMarinier: Yes, that is fine. I'm just suggesting adding this simple assertion to make sure. Also, I find it makes the test more readable. Otherwise, when I first read the test, I thought it was a mistake.

webapp/src/components/ClassOverlapTable.test.tsx
JosephMarinier: Will you try to wait for the API response to be fetched? Like adding an `expect()` for something that shows that the API results are loaded, and then checking for the `See more` button?

nandhinibsn: Right. Good approach 👍 . I added a test scenario to check "MuiDataGrid-row" is loaded to ensure API has successfully fetched the records then followed by testing the footer component.

webapp/src/components/ClassOverlapTable.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+      screen.queryByRole("columnheader", { name: /Pipeline confusion/ })
+    ).toBeNull();
+  });
+
+  it("should not display the Footer component if the number of rows is less than initial number", async () => {
+    renderClassOverlapTable({ train: true, eval: true }, { pipelineIndex: 0 });
+    await waitFor(() => {
+      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row");
+      expect(screen.queryByText(/See more/i)).toBeNull();
+    });
```
JosephMarinier, line 79: I'd add a small comment on why we do this, since `expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row")` is not super intuitive. Also, you can move the `expect(screen.queryByText(/See more/i)).toBeNull()` after the `waitFor()` so it's executed only once.
```suggestion
    // Wait for API data to be loaded
    await waitFor(() => {
      expect(screen.getAllByRole("row")[1]).toHaveClass("MuiDataGrid-row");
    });
    expect(screen.queryByText(/See more/i)).toBeNull();
```

JosephMarinier (approved):

PR #444, opened 2023-02-16
Change the outcome per threshold bar chart to area chart

### LINKED ISSUE #430: Change the outcome per threshold bar plot to area plot and add the current threshold value as a vertical line

JosephMarinier:
Resolve #430

## Description:

* [Clean up front-end imports](https://github.com/ServiceNow/azimuth/commit/4a83d8b3b86d9eb2f1c62bb491aec20aadbb32ee)
* [Clean up `get_dataset_split()` by leveraging `super()`](https://github.com/ServiceNow/azimuth/commit/25dc03343925920dd9640be9e5371a928d2492f2)
* [Compute outcome count at 100% threshold](https://github.com/ServiceNow/azimuth/commit/9d970dd448075b14fe50ac408358d6fd9669d690)
* [Return error if post-processing is not available](https://github.com/ServiceNow/azimuth/commit/93642469e70f5eef6088a4a9b1d8c09342be7bb4)
* [Draw area chart](https://github.com/ServiceNow/azimuth/commit/eb0364ca93a18f7b1e2d0c4a3e75e90abfcb238d) 

  - Add `confidenceThreshold` and `utteranceCount` to API response
  - Change the outcome per threshold bar chart to area chart, making the x axis continuous
* [Add a dashed line for current threshold](https://github.com/ServiceNow/azimuth/commit/4284c010972fb4d19294cd0c82b3421ea17f4b36)
* [Log changes](https://github.com/ServiceNow/azimuth/commit/5477bbfc9eca3d14096b9c9e9f35e384bd6b3346)
* [Adapt test to error if post-processing is not available](https://github.com/ServiceNow/azimuth/pull/444/commits/036c54c8bbf0b505fd020e6adf19344f0c1f3d56)

image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/indexable_module.py
+++ b/azimuth/modules/base_classes/indexable_module.py
@@ -48,7 +48,7 @@ class IndexableModule(Module[ConfigScope], ABC):
             The loaded dataset_split.
 
         """
-        dataset_split = self.get_full_dataset_split(name)
+        dataset_split = super().get_dataset_split(name)
         indices = self.mod_options.indices
         if indices:
             dataset_split = dataset_split.select(indices)
--- a/azimuth/modules/model_performance/outcome_count.py
+++ b/azimuth/modules/model_performance/outcome_count.py
@@ -213,4 +213,10 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
                     outcome_count=Counter(postprocessed_outcomes),
                 )
             )
-        return [OutcomeCountPerThresholdResponse(outcome_count_all_thresholds=result)]
+        return [
+            OutcomeCountPerThresholdResponse(
+                outcome_count_per_threshold=result,
+                confidence_threshold=self.get_threshold(),
+                utterance_count=len(self.get_dataset_split()),
+            )
+        ]
--- a/azimuth/modules/model_performance/outcome_count.py
+++ b/azimuth/modules/model_performance/outcome_count.py
@@ -186,8 +186,8 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
 
     def compute_on_dataset_split(self) -> List[OutcomeCountPerThresholdResponse]:  # type: ignore
         if not postprocessing_editable(self.config, self.mod_options.pipeline_index):
-            # This will raise an error if a user gets to this page.
-            return []
+            raise ValueError("Postprocessing is not editable")
+
         x_ticks_count = self.mod_options.x_ticks_count
         ths = np.linspace(0, 1, x_ticks_count)
         result = []
--- a/azimuth/modules/model_performance/outcome_count.py
+++ b/azimuth/modules/model_performance/outcome_count.py
@@ -182,14 +182,14 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
     """Compute the outcome count per threshold."""
 
     required_mod_options = {"pipeline_index"}
-    optional_mod_options = {"nb_bins"}
+    optional_mod_options = {"x_ticks_count"}
 
     def compute_on_dataset_split(self) -> List[OutcomeCountPerThresholdResponse]:  # type: ignore
         if not postprocessing_editable(self.config, self.mod_options.pipeline_index):
             # This will give an empty response to the UI, if a user gets to this page.
             return [OutcomeCountPerThresholdResponse(outcome_count_all_thresholds=[])]
-        nb_bins = self.mod_options.nb_bins
-        ths = np.linspace(0, 1, nb_bins, endpoint=False)
+        x_ticks_count = self.mod_options.x_ticks_count
+        ths = np.linspace(0, 1, x_ticks_count)
         result = []
         for th in tqdm(
             ths,
--- a/azimuth/modules/model_performance/outcome_count.py
+++ b/azimuth/modules/model_performance/outcome_count.py
@@ -186,8 +186,8 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
 
     def compute_on_dataset_split(self) -> List[OutcomeCountPerThresholdResponse]:  # type: ignore
         if not postprocessing_editable(self.config, self.mod_options.pipeline_index):
-            # This will give an empty response to the UI, if a user gets to this page.
-            return [OutcomeCountPerThresholdResponse(outcome_count_all_thresholds=[])]
+            # This will raise an error if a user gets to this page.
+            return []
         x_ticks_count = self.mod_options.x_ticks_count
         ths = np.linspace(0, 1, x_ticks_count)
         result = []
--- a/azimuth/routers/v1/model_performance/outcome_count.py
+++ b/azimuth/routers/v1/model_performance/outcome_count.py
@@ -4,7 +4,8 @@
 
 from typing import List
 
-from fastapi import APIRouter, Depends, Query
+from fastapi import APIRouter, Depends, HTTPException, Query
+from starlette.status import HTTP_400_BAD_REQUEST
 
 from azimuth.app import get_dataset_split_manager, get_task_manager
 from azimuth.dataset_split_manager import DatasetSplitManager
@@ -57,8 +58,7 @@ def get_outcome_count_per_threshold(
     )
 
     if len(task_result) == 0:
-        # Non-Editable postprocessing
-        return []
+        raise HTTPException(HTTP_400_BAD_REQUEST, detail="Postprocessing is not editable")
 
     return task_result[0].outcome_count_all_thresholds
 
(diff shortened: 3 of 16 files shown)


webapp/src/components/ThresholdPlot.tsx
```diff
@@ -38,6 +39,7 @@ const useStyles = makeStyles((theme) => ({
     gridRow: "start / end",
     placeSelf: "center",
     "p&": {
+      // TODO what?
```
christyler3030, line 42: I don't understand this comment.

JosephMarinier: Ah-ha! Well me neither. 😅 I meant to investigate what `"p&"` was supposed to do, and I forgot.

webapp/src/components/ThresholdPlot.tsx
```diff
@@ -48,31 +50,75 @@ const useStyles = makeStyles((theme) => ({
...
+    component="path"
+    d={`M${x},${y} Z`}
+    strokeLinecap="round"
+    strokeWidth={6}
+    vectorEffect="non-scaling-stroke"
+    sx={(theme) => ({ stroke: theme.palette.common.black })}
+  />
+);
+
+const Curve: React.FC<{
```
christyler3030, line 100: Nitpick: I think this might not be the best name for this component.  It's more of a Region, or Area, or something isn't it?

webapp/src/components/ThresholdPlot.tsx
JosephMarinier: Ah-ha! It applies to `.plotArea` only if it is a `p` element. It translates to `p.makeStyles-plotArea-38`. It could be clearer. The white background is also problematic when the plot is not on a paper, like it is on its standalone page. So, I'll rework the loading and error states.
image

webapp/src/components/ThresholdPlot.tsx
JosephMarinier: Correct. I meant it in the sense that this is rendering one "curve" (or one "series" of points as Google Sheets puts it, but I thought "series" was confusing as I mean one series, singular). Yes, it is rendered as an area, plus a line, plus points, but I considered that to be an implementation detail, a bit like the component `ThresholdPlot` isn't like `ThresholdAreaPlot` or `ThresholdBarPlot`. Does that make sense? I don't have a strong preference. In this perspective, it could be `Scatter`? Or do you have other ideas?

christyler3030: No I don't really have any other strong preferences here. No need to overthink this.

christyler3030: I guess the fact that the `Curve` know about outcomes means this is like an OutcomeRegion or something. But this is probably fine as is. We are not building a plotting library here.

christyler3030 (approved):

webapp/src/components/ThresholdPlot.tsx
JosephMarinier: If it was just a line (as opposed to an area), it could still know the outcome to color the line (instead of the area). I don't think that's a problem. I was about to suggest other names, but you're right that the implications are very limited here, so let's not overthink it. I'll leave it as is.

azimuth/modules/base_classes/indexable_module.py
```diff
@@ -48,7 +48,7 @@ def get_dataset_split(self, name: Optional[DatasetSplitName] = None) -> Dataset:
             The loaded dataset_split.
 
         """
-        dataset_split = self.get_full_dataset_split(name)
+        dataset_split = super().get_dataset_split(name)
```
JosephMarinier, line 51: Instead of having `get_full_dataset_split()` and `get_dataset_split()` that are redundant in the base `Module`, let's just have `get_dataset_split()` and use `super().get_dataset_split()` in `IndexableModule`.

azimuth/modules/model_performance/outcome_count.py
```diff
@@ -182,14 +182,14 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
...
     required_mod_options = {"pipeline_index"}
-    optional_mod_options = {"nb_bins"}
+    optional_mod_options = {"x_ticks_count"}
 
     def compute_on_dataset_split(self) -> List[OutcomeCountPerThresholdResponse]:  # type: ignore
         if not postprocessing_editable(self.config, self.mod_options.pipeline_index):
-            # This will give an empty response to the UI, if a user gets to this page.
-            return [OutcomeCountPerThresholdResponse(outcome_count_all_thresholds=[])]
-        nb_bins = self.mod_options.nb_bins
-        ths = np.linspace(0, 1, nb_bins, endpoint=False)
```
JosephMarinier, line 192: Key difference here is that `endpoint=False` is gone.

azimuth/modules/model_performance/outcome_count.py
```diff
@@ -182,14 +182,14 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
     """Compute the outcome count per threshold."""
 
     required_mod_options = {"pipeline_index"}
-    optional_mod_options = {"nb_bins"}
+    optional_mod_options = {"x_ticks_count"}
 
     def compute_on_dataset_split(self) -> List[OutcomeCountPerThresholdResponse]:  # type: ignore
         if not postprocessing_editable(self.config, self.mod_options.pipeline_index):
-            # This will give an empty response to the UI, if a user gets to this page.
-            return [OutcomeCountPerThresholdResponse(outcome_count_all_thresholds=[])]
```
JosephMarinier, line 190: We were checking `len(task_result) == 0`, which was never happening because of that.

azimuth/routers/v1/model_performance/outcome_count.py
```diff
@@ -57,10 +57,9 @@ def get_outcome_count_per_threshold(
     )
 
     if len(task_result) == 0:
-        # Non-Editable postprocessing
-        return []
+        raise HTTPException(HTTP_400_BAD_REQUEST, detail="Postprocessing is not editable")
```
JosephMarinier, line 60: I don't have a specific reason to change that behavior, I simply noticed it might make more sense to return an error than silently return an empty plot. What do you think?

azimuth/modules/base_classes/indexable_module.py
gabegma: Ok, I'm ok with it! I created the 2 methods initially because I thought they were not exactly doing the same thing (one was filtering, and the other not), and I thought that some modules might require the full dataset split sometimes. But given that it's not used anywhere in our repo, I'm fine with your change.

azimuth/modules/model_performance/outcome_count.py
```diff
@@ -182,14 +182,14 @@ class OutcomeCountPerThresholdModule(AggregationModule[ModelContractConfig]):
...
-        nb_bins = self.mod_options.nb_bins
-        ths = np.linspace(0, 1, nb_bins, endpoint=False)
+            # This will raise an error if a user gets to this page.
+            return []
```
gabegma, line 190: Should we raise here, instead of returning an empty list?

azimuth/modules/model_performance/outcome_count.py
gabegma: Great idea.

azimuth/routers/v1/model_performance/outcome_count.py
gabegma: The front end will never call this plot if post-processing is unavailable right? I think I would raise in the module though, not in the route?

gabegma (approved):
LGTM for the back end!! Good job! Some minor comments for the `raise`, but I'm fine with the current implementation too if you think that it's preferred.

azimuth/modules/model_performance/outcome_count.py
gabegma: `nb_bins` was just plain wrong, good job for changing that.

azimuth/routers/v1/model_performance/outcome_count.py
JosephMarinier: Right. The user doesn't get a way to navigate to the page, but nothing prevents him to type the URL (or refresh the page with a different config). So it's not impossible, but it's not in the usual workflow.

The module used to return an empty response
https://github.com/ServiceNow/azimuth/blob/5b9f9693d0900bee8b90fbc7c3aa5f336ec17f39/azimuth/modules/model_performance/outcome_count.py#L187-L190
And I changed it to return no response (see my comment https://github.com/ServiceNow/azimuth/pull/444/files#r1109929478 about the change)
https://github.com/ServiceNow/azimuth/blob/1015ad301d07bdea3cbf7630e33624b31a95f6de/azimuth/modules/model_performance/outcome_count.py#L187-L190
but it doesn't seem to `raise`. Am I missing something?

Do you think the module should `raise`? And here we `except` and re-`raise` an `HTTPException`?

JosephMarinier: Here it is! What do you think? d5648c6b8eaf43f71461296a03a6e3a6b3c68e33

gabegma: Yes, perfect :) Thanks!!

PR #450, opened 2023-02-17
Clean up pipeline vs model and previous PR

gabegma:
## Description:
- Cleanup following the previous PRs.
- For model vs pipeline, I tried to remain as coherent as possible. I refer model as the callable function in PipelineDefinition, even though this model can be a HF pipeline, a model, a custom pipeline, or a file prediction reader. 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/module.py
+++ b/azimuth/modules/base_classes/module.py
@@ -155,7 +155,8 @@ class Module(DaskModule[ConfigScope]):
         """Load the model according to the config and module options.
 
         This will invoke the Python method supplied in the config.
-        See our API Doc for more details.
+        See our API Doc for more details. It can return a HF Pipeline, a Callable (which can be just
+        a model or a custom pipeline) and a file prediction reader.
 
         Returns:
             Loaded model according to user spec.
@@ -163,7 +164,7 @@ class Module(DaskModule[ConfigScope]):
         Raises:
             ValueError if no valid pipeline exists.
         """
-        _ = self.get_current_pipeline()  # Validate current pipeline exists
+        _ = self.get_current_pipeline_definition()  # Validate current pipeline exists
         return self.artifact_manager.get_model(self.config, self.mod_options.pipeline_index)
 
     def _get_table_key(self) -> Optional[PredictionTableKey]:
@@ -176,7 +177,7 @@ class Module(DaskModule[ConfigScope]):
             self.config, ModelContractConfig
         ):
             return None
-        current_pipeline = self.get_current_pipeline()
+        current_pipeline = self.get_current_pipeline_definition()
         use_bma = self.mod_options.use_bma
         table_key = PredictionTableKey(
             temperature=current_pipeline.temperature,
@@ -188,18 +189,18 @@ class Module(DaskModule[ConfigScope]):
 
     def get_threshold(self) -> Optional[float]:
         # The default is None so we have to handle it this way.
-        current_pipeline = self.get_current_pipeline()
+        current_pipeline = self.get_current_pipeline_definition()
 
         thresh = self.mod_options.threshold
         if thresh is None:
             thresh = current_pipeline.threshold
         return thresh
 
-    def get_current_pipeline(self) -> PipelineDefinition:
-        """Get current pipeline for this spec.
+    def get_current_pipeline_definition(self) -> PipelineDefinition:
+        """Get the current pipeline definition from the config.
 
         Returns:
-            Config of the current pipeline.
+            Definition of the current pipeline from the config.
 
         Raises:
             ValueError if no valid pipeline exists.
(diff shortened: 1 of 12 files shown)


tests/test_modules/test_model_contracts/test_hf_text_classification.py
```diff
@@ -28,10 +28,7 @@ def test_create_sentiment_pipeline(simple_text_config):
     # By default, use L2 gradient calculation
     assert task.gradient_calculation == GradientCalculation.L2
 
-    sa = task.get_model()
-
-    # As of Jan 6, 2021, fast tokenizers are not supported in pipelines
```
gabegma, line 33: This is no longer true, but the tokenizer still has `is_fast=False` in this case. I removed this line in the test since it wasn't testing something useful.

azimuth/modules/model_contracts/hf_text_classification.py
```diff
@@ -85,20 +85,20 @@ def predict(self, batch: Dataset) -> List[PredictionResponse]:
                     ],
                     axis=-1,
                 )
-                epistemic: List[float] = BALD().get_uncertainties(predictions)
-                pipeline_out = predictions.mean(-1)
+                epistemic: List[float] = BALD().get_uncertainties(multiple_predictions)
+                predictions = multiple_predictions.mean(-1)
```
JosephMarinier, line 89: So, `get_postprocessed_output()` is calling its argument `pipeline_output`, but it's more the "model output", right? And it adds our postprocessing. Except the "model" can have some postprocessing in it already. Am I following?
Anyway, I have concerns `multiple_predictions`. Since `predictions` is already plural, I find `multiple_predictions` doesn't really convey it has one more dimension. I think we generally refer to the `output` (singular) of the mode/pipeline, and `outputs` (plural) would convey it did multiple predictions. What do you think?
```suggestion
                output = outputs.mean(-1)
```
And `output` is just to avoid saying `model_` or `pipeline_`, but where we know, we can specify the stage of the output.

azimuth/modules/base_classes/module.py
```diff
@@ -200,18 +201,18 @@ def _get_table_key(self) -> Optional[PredictionTableKey]:
...
+        current_pipeline = self.get_current_pipeline_definition()
 
         thresh = self.mod_options.threshold
         if thresh is None:
             thresh = current_pipeline.threshold
         return thresh
 
-    def get_current_pipeline(self) -> PipelineDefinition:
-        """Get current pipeline for this spec.
+    def get_current_pipeline_definition(self) -> PipelineDefinition:
```
JosephMarinier, line 211: Getters usually get the "current" thing, unsell specified otherwise. Is there a reason for specifying "current" here?
```suggestion
    def get_pipeline_definition(self) -> PipelineDefinition:
```
Also, the type really is `PipelineDefinition`, so I wouldn't hesitate to call the variables `pipeline_definition` too, for example:
```diff
-        current_pipeline = self.get_current_pipeline_definition()
+        pipeline_definition = self.get_pipeline_definition()
```

azimuth/modules/base_classes/module.py
```diff
@@ -223,7 +224,7 @@ def get_current_pipeline(self) -> PipelineDefinition:
                 " as it does not use ModelContractScope."
             )
         if self.config.pipelines is None:
-            raise ValueError("A model was not provided in the config.")
+            raise ValueError("A pipeline was not provided in the config.")
```
JosephMarinier, line 227: While we're at it, I'd suggest something like
```suggestion
            raise ValueError("No pipelines were provided in the config.")
```
or even
```suggestion
            raise ValueError("No pipelines configured.")
```

azimuth/modules/base_classes/module.py
gabegma: I agree for `current`, good call. As for the `get`, I would be tempted to stay consistent with the other methods, such as `get_threshold`.  And also given that it is a method and not a property. Thoughts?

JosephMarinier: I agree. I did not suggest removing `get_`.

azimuth/modules/model_contracts/hf_text_classification.py
JosephMarinier: Right, I saw the 3 model contracts. Especially in the other 2, the "scope" of the variable is so short that I wouldn't mind the vague name `output`. And it the context of a method called `predict`, I think it makes sense. But yeah, more clarity never hurts.

You `n` suggestion works for me too. Or do you think `iterations_predictions` would work?
```suggestion
                predictions = predictions_iterations.mean(-1)
```

Alternatively, could we consider the model/pipeline output the "prediction"? Singular. And when we get multiple predictions with `iterations`, then that is `predictions`?
```suggestion
                prediction = predictions.mean(-1)
```

azimuth/modules/base_classes/module.py
gabegma: Oh sorry, I read the comment too quickly it seems!

JosephMarinier: I figured. No worries!

azimuth/modules/model_contracts/hf_text_classification.py
gabegma: I don't like prediction singular because we send a batch of input to the function and receive a prediction for every input, so predictions. I'll go with `n_predictions` I think.

JosephMarinier (approved):
👍

PR #452, opened 2023-02-17
Refactor get_utterances

### LINKED ISSUE #451: Refactor get_utterances so it runs faster

gabegma:
Resolve #451

## Description:
I'm sure we could do more, but this is still a significant gain.
Before: 
Screen Shot 2023-02-17 at 15 48 29

After:
Screen Shot 2023-02-17 at 15 47 17

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/routers/v1/utterances.py
+++ b/azimuth/routers/v1/utterances.py
@@ -111,8 +111,11 @@ def get_utterances(
     else:
         threshold, table_key = None, None
 
+    ds = dataset_split_manager.get_dataset_split(table_key)
+    if indices is not None:
+        ds = ds.select(indices)
     ds = filter_dataset_split(
-        dataset_split_manager.get_dataset_split(table_key),
+        ds,
         named_filters.to_dataset_filters(dataset_split_manager.get_class_names()),
         config,
         without_postprocessing,
@@ -120,13 +123,6 @@ def get_utterances(
     ds = dataset_split_manager.get_dataset_split_with_class_names(table_key=table_key).select(
         ds[DatasetColumn.row_idx]
     )
-
-    if indices is not None:
-        if len(set(indices) - set(ds[DatasetColumn.row_idx])) > 0:
-            raise HTTPException(
-                HTTP_404_NOT_FOUND, detail=f"Indices {indices} not found after filtering"
-            )
-        ds = ds.filter(lambda i: i in indices, input_columns=DatasetColumn.row_idx)
 
     # We create _top_conf and _top_prediction because we can't sort on columns made of lists.
     # They start with an underscore to emphasize that they are not saved and that therefore they
--- a/azimuth/routers/v1/utterances.py
+++ b/azimuth/routers/v1/utterances.py
@@ -64,6 +64,7 @@ from azimuth.utils.routers import (
     require_available_model,
     require_pipeline_index,
 )
+from azimuth.utils.validation import assert_not_none
 
 router = APIRouter()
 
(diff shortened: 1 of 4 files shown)


azimuth/routers/v1/utterances.py
```diff
@@ -235,29 +189,36 @@ def get_utterances(
...
+                preprocessing_steps=data[DatasetColumn.pipeline_steps]["preprocessing_steps"],
+                postprocessing_steps=data[DatasetColumn.pipeline_steps]["postprocessing_steps"],
+            )
+            if predictions_available(config) and pipeline_index is not None
+            else None,
             model_saliency=model_saliency,
             # Smart tags families
             **{
-                family.value: [t for t in tags_in_family if tag[t]]
+                family.value: [t for t in tags_in_family if data[t]]
```
JosephMarinier, line 215: 😲 
👍

azimuth/routers/v1/utterances.py
```diff
@@ -100,44 +101,30 @@ def get_utterances(
...
+        ds[DatasetColumn.row_idx]
     )
 
     if indices is not None:
         if len(set(indices) - set(ds[DatasetColumn.row_idx])) > 0:
             raise HTTPException(
-                HTTP_404_NOT_FOUND, detail=f"Indices not found after filtering {indices}"
+                HTTP_404_NOT_FOUND, detail=f"Indices {indices} not found after filtering"
             )
         ds = ds.filter(lambda i: i in indices, input_columns=DatasetColumn.row_idx)
```
JosephMarinier, line 129: Is it possible to select the desired `indices` using `ds.select()`? If we do it before `filter_dataset_split()`?

JosephMarinier (approved):
That's insane! 🤯 Good job! 👍

azimuth/routers/v1/utterances.py
```diff
@@ -173,54 +160,21 @@ def get_utterances(
...
         return GetUtterancesResponse(
             utterances=[], utterance_count=utterance_count, confidence_threshold=threshold
         )
 
-    indices_subset = ds[DatasetColumn.row_idx]
-    tags = dataset_split_manager.get_tags(indices_subset, table_key=table_key)
-
-    if predictions_available(config) and pipeline_index is not None:
-        # For memory efficiency
-        ds = ds.with_format(
```
JosephMarinier, line 187: 🤮

azimuth/routers/v1/utterances.py
gabegma: I think it would be possible, but if we do it before `filter_dataset_split()`, we won't be able to `raise` here if the filters and the indices were conflicting, right? But IMO, we could also block this route from working if both indices and filters are sent. I don't see a use case for it!

gabegma:
FYI @JosephMarinier - I did push 2 new commits. It seems we gain a few extra `ms`. I was intrigued by the use of `input_columns` in the filter function.

JosephMarinier:
Oh yeah, cool! 👍

azimuth/routers/v1/utterances.py
JosephMarinier: Yeah, I don't see a use case for indices and filters at the same time, and even less so for when this "Indices not found after filtering" would be useful.
I think we could totally block using indices and filters at the same time, but I'm also fine with keeping it and NOT erroring out if "Indices not found after filtering".
Am I missing something?

gabegma: I was going exactly in that direction just now - I was thinking of just returning the utterances that both satisfy the filters and the indices condition. Are we on the same page?

JosephMarinier: Yup! 👍 If a user makes a weird request, we'll answer with a weird response. Seems expected.

gabegma:
@JosephMarinier I added a test too because I thought we weren't testing properly the route.

JosephMarinier (approved):
Nice! Perfect! 👍

PR #460, opened 2023-02-21
Support ignoring missing persistent-ids in patch utterances route

gabegma:
Part of #297 

## Description:
- I put `False` as a default. I think the method should have False as a default, but we could put the default at `True` for the route. 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/routers/v1/utterances.py
+++ b/azimuth/routers/v1/utterances.py
@@ -276,15 +276,22 @@ def patch_utterances(
     utterances: List[UtterancePatch] = Body(...),
     dataset_split_manager: DatasetSplitManager = Depends(get_dataset_split_manager),
     task_manager: TaskManager = Depends(get_task_manager),
+    ignore_not_found: bool = False,
 ) -> List[UtterancePatch]:
     persistent_ids = [utterance.persistent_id for utterance in utterances]
     try:
-        row_indices = dataset_split_manager.get_row_indices_from_persistent_id(persistent_ids)
+        row_indices = dataset_split_manager.get_row_indices_from_persistent_id(
+            persistent_ids, ignore_not_found
+        )
     except ValueError as e:
         raise HTTPException(HTTP_404_NOT_FOUND, detail=f"Persistent id not found: {e}.")
 
+    row_indices_found = [idx for idx in row_indices if idx is not None]
+    utterances_found = [
+        utterance for idx, utterance in zip(row_indices, utterances) if idx is not None
+    ]
     data_actions = {}
-    for row_idx, utterance in zip(row_indices, utterances):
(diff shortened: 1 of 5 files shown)


azimuth/routers/v1/utterances.py
```diff
@@ -276,24 +276,32 @@ def patch_utterances(
...
+        row_indices = dataset_split_manager.get_row_indices_from_persistent_id(
+            persistent_ids, ignore_not_found
+        )
     except ValueError as e:
         raise HTTPException(HTTP_404_NOT_FOUND, detail=f"Persistent id not found: {e}.")
 
+    row_indices_found = [idx for idx in row_indices if idx is not None]
+    utterances_found = [
+        utterance for idx, utterance in zip(row_indices, utterances) if idx is not None
+    ]
```
gabegma, line 292: I initially did it in one loop, but `mypy` would complain. LMK if you have a better idea. 
```
row_indices_found, utterances_found = zip(
    *[(idx, utterance) for idx, utterance in zip(row_indices, utterances) if idx is not None]
)
```

JosephMarinier: Instead of checking for missing persistent ids inside `get_row_indices_from_persistent_id()`, returning `None`, and then having to filter out the `None`s, I think we can filter out the `utterances` with missing persistent ids straight at the beginning. What do you think of this #461?

azimuth/routers/v1/utterances.py
```diff
@@ -276,24 +276,32 @@ def patch_utterances(
     utterances: List[UtterancePatch] = Body(...),
     dataset_split_manager: DatasetSplitManager = Depends(get_dataset_split_manager),
     task_manager: TaskManager = Depends(get_task_manager),
+    ignore_not_found: bool = False,
```
JosephMarinier, line 279: > I put `False` as a default. I think the method should have False as a default, but we could put the default at `True` for the route. 

I think `False` is perfect for the route too. Even though `True` might be more useful, I wouldn't want someone to use the API, thinking all his data has been imported (otherwise it should have errored out), and then delete his export CSV file, to then realize that he didn't patch to the correct Azimuth instance and that he lost his proposed actions. When having to specify `?ignore_not_found=true`, at least he is conscious of the behavior and might verify the result.

azimuth/routers/v1/utterances.py
gabegma: Right, that's simpler. It's true that the method doesn't need to support it. Thanks.

azimuth/routers/v1/utterances.py
gabegma: My thought too.

JosephMarinier:
Will be handled in #461

PR #464, opened 2023-02-23
Add fix to browser history after navigating to exploration page with no pipeline

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Dashboard.tsx
+++ b/webapp/src/pages/Dashboard.tsx
@@ -71,7 +71,9 @@ const Dashboard = () => {
           color="secondary"
           variant="contained"
           component={Link}
-          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/prediction_overview${searchString}`}
+          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/${
+            !isPipelineSelected(pipeline) ? `utterances` : `prediction_overview`
+          }${searchString}`}
           sx={{ gap: 1 }}
         >
           <Telescope fontSize="large" />
--- a/webapp/src/pages/Dashboard.tsx
+++ b/webapp/src/pages/Dashboard.tsx
@@ -71,7 +71,9 @@ const Dashboard = () => {
           color="secondary"
           variant="contained"
           component={Link}
-          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/prediction_overview${searchString}`}
+          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/${
+            isPipelineSelected(pipeline) ? "prediction_overview" : "utterances"
+          }${searchString}`}
           sx={{ gap: 1 }}
         >
           <Telescope fontSize="large" />
--- a/webapp/src/pages/Dashboard.tsx
+++ b/webapp/src/pages/Dashboard.tsx
@@ -71,9 +71,7 @@ const Dashboard = () => {
           color="secondary"
           variant="contained"
           component={Link}
-          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/${
-            !isPipelineSelected(pipeline) ? `utterances` : `prediction_overview`
-          }${searchString}`}
+          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/prediction_overview${searchString}`}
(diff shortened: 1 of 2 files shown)


nandhinibsn:
There was a console error which occurred when we navigate to the exploration page with no pipeline.
Screen Shot 2023-02-23 at 5 17 01 PM

I simplified the solution to navigating on the Dashboard page instead of the Exploration page. I believe it's more of a simplified way. Let me know your thoughts @JosephMarinier

webapp/src/pages/Dashboard.tsx
```diff
@@ -71,7 +71,9 @@ const Dashboard = () => {
           color="secondary"
           variant="contained"
           component={Link}
-          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/prediction_overview${searchString}`}
+          to={`/${jobId}/dataset_splits/${firstAvailableDatasetSplit}/${
+            !isPipelineSelected(pipeline) ? `utterances` : `prediction_overview`
```
JosephMarinier, line 75: I find the condition is easier to read if you remove the `!` and flip. Also, regular strings would be goof enough here.
```suggestion
            isPipelineSelected(pipeline) ? "prediction_overview" : "utterances"
```

nandhinibsn:
Hey @JosephMarinier, the solution added to Dashboard has been reverted as it resolved the issue with tracking history. Still, it did not switch tabs from prediction overview to utterance table if no pipeline was selected from the exploration page.

Additionally, I understood why the console error was caused by reading this [blog](https://www.appsloveworld.com/reactjs/100/18/cannot-update-during-an-existing-state-transition-such-as-within-render-rend). 

To solve the bug, I added useEffect to handle the pipeline change from the exploration page to solve the issue with tracking history while the console error is not occurring anymore. Let me know what you think

JosephMarinier:
Excellent! 👍 
As you said, the `React.useEffect` is perfect for if you select _No pipeline_ when already on the Exploration Space. However, when navigating from the Dashboard to the Exploration Space, it now works (👍) but it adds an intermediary state in the history, on the invalid Prediction Overview tab. So, I think the two solutions could be combined. You could also add your initial solution to the `Link to=` on the Dashboard, to polish the navigation. It also has the advantage of showing the correct destination when hovering the button.

webapp/src/pages/Exploration.tsx
```diff
@@ -67,9 +67,11 @@ const Exploration = () => {
...
   };
 
-  if (!isPipelineSelected(pipeline) && mainView !== "utterances") {
-    setMainView("utterances");
-  }
+  React.useEffect(() => {
+    if (!isPipelineSelected(pipeline) && mainView !== "utterances") {
+      setMainView("utterances");
+    }
+  }, [pipeline]);
```
JosephMarinier, line 74: From the terminal
> React Hook React.useEffect has missing dependencies: 'mainView' and 'setMainView'.

webapp/src/pages/Dashboard.tsx
nandhinibsn: Sure. done

webapp/src/pages/Exploration.tsx
nandhinibsn: I added all the necessary dependencies.

nandhinibsn:
> Excellent! 👍 As you said, the `React.useEffect` is perfect for if you select _No pipeline_ when already on the Exploration Space. However, when navigating from the Dashboard to the Exploration Space, it now works (👍) but it adds an intermediary state in the history, on the invalid Prediction Overview tab. So, I think the two solutions could be combined. You could also add your initial solution to the `Link to=` on the Dashboard, to polish the navigation. It also has the advantage of showing the correct destination when hovering the button.

Okay, Make sense. I will add the condition back to the Dashboard page.

webapp/src/pages/Exploration.tsx
JosephMarinier: You added only one of the two dependencies.

nandhinibsn: Okay, I was not aware adding a function to dependencies is acceptable. I added the dependencies. Please check my recent commit.

JosephMarinier (approved):
🚢 🇮🇹

PR #465, opened 2023-02-24
Import proposed actions to utterance table

### LINKED ISSUE #297: Ability to import back proposed actions

nandhinibsn:
Resolve #297

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -409,7 +409,6 @@ const UtterancesTable: React.FC<Props> = ({
     fileReader.onload = (event) => {
       if (event.target) {
         const data = event.target.result as string;
-        console.log(data);
         const headers: string[] = data.slice(0, data.indexOf("\n")).split(",");
         const parsedCSVData: UtterancePatch[] = data
           .slice(data.indexOf("\n") + 1)
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -163,6 +163,7 @@ const UtterancesTable: React.FC<Props> = ({
       })) ?? [],
     [utterancesResponse]
   );
+
   const [selectedPersistentIds, setSelectedPersistentIds] = React.useState<
     number[]
   >([]);
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -1,6 +1,7 @@
 import { ArrowDropDown, GetApp, SvgIconComponent } from "@mui/icons-material";
 import FilterAltOutlinedIcon from "@mui/icons-material/FilterAltOutlined";
 import MultilineChartIcon from "@mui/icons-material/MultilineChart";
+import UploadIcon from "@mui/icons-material/Upload";
 import { Box, Button, Menu, MenuItem } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import {
@@ -73,6 +74,7 @@ const useStyles = makeStyles((theme) => ({
   gridHeaderActions: {
     display: "flex",
     flexDirection: "row",
+    justifyContent: "space-between",
   },
   searchContainer: {
     marginLeft: theme.spacing(2),
@@ -406,49 +408,63 @@ const UtterancesTable: React.FC<Props> = ({
           text="Explore utterances and propose actions. Click on a row to inspect the utterance details."
           link="user-guide/exploration-space/utterance-table/"
         />
-        <Button
-          id="export-button"
-          aria-controls="export-menu"
-          aria-haspopup="true"
-          className={classes.exportButton}
-          onClick={(event) => setAnchorEl(event.currentTarget)}
-          startIcon={<GetApp />}
-          endIcon={<ArrowDropDown />}
-        >
-          Export
-        </Button>
-        <Menu
-          id="export-menu"
-          anchorEl={anchorEl}
-          keepMounted
-          open={Boolean(anchorEl)}
-          onClose={() => setAnchorEl(null)}
-        >
-          <MenuItem
-            onClick={() => {
-              downloadDatasetSplit({
-                jobId,
-                datasetSplitName,
-                ...filters,
-                ...pipeline,
-              });
-              setAnchorEl(null);
-            }}
+        <Box display="flex" alignItems="center" gap={2}>
+          <Button
+            className={classes.exportButton}
+            component="label"
+            startIcon={<UploadIcon />}
           >
-            Export utterances
-          </MenuItem>
-          <MenuItem
-            onClick={() => {
-              downloadUtteranceProposedActions({
-                jobId,
-                datasetSplitName,
-              });
-              setAnchorEl(null);
-            }}
+            Import
+            <input
+              hidden
+              accept=".csv, application/vnd.openxmlformats-officedocument.spreadsheetml.sheet, application/vnd.ms-excel"
+              type="file"
+            />
+          </Button>
+          <Button
+            id="export-button"
+            aria-controls="export-menu"
+            aria-haspopup="true"
+            className={classes.exportButton}
+            onClick={(event) => setAnchorEl(event.currentTarget)}
+            startIcon={<GetApp />}
+            endIcon={<ArrowDropDown />}
           >
-            Export proposed actions
-          </MenuItem>
-        </Menu>
+            Export
+          </Button>
+          <Menu
+            id="export-menu"
+            anchorEl={anchorEl}
+            keepMounted
+            open={Boolean(anchorEl)}
+            onClose={() => setAnchorEl(null)}
+          >
+            <MenuItem
+              onClick={() => {
+                downloadDatasetSplit({
+                  jobId,
+                  datasetSplitName,
+                  ...filters,
+                  ...pipeline,
+                });
+                setAnchorEl(null);
+              }}
+            >
+              Export utterances
+            </MenuItem>
+            <MenuItem
+              onClick={() => {
+                downloadUtteranceProposedActions({
+                  jobId,
+                  datasetSplitName,
+                });
+                setAnchorEl(null);
+              }}
+            >
+              Export proposed actions
+            </MenuItem>
+          </Menu>
+        </Box>
       </div>
       <Table
         pagination
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -2,7 +2,7 @@ import { ArrowDropDown, GetApp, SvgIconComponent } from "@mui/icons-material";
 import FilterAltOutlinedIcon from "@mui/icons-material/FilterAltOutlined";
 import MultilineChartIcon from "@mui/icons-material/MultilineChart";
 import UploadIcon from "@mui/icons-material/Upload";
-import { Box, Button, Menu, MenuItem, Typography } from "@mui/material";
+import { Box, Button, Menu, MenuItem } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import {
   GridCellParams,
@@ -412,23 +412,22 @@ const UtterancesTable: React.FC<Props> = ({
         }
         const headers: string[] = header.split(",");
         if (
-          ![config.columns.persistent_id, "proposed_action"].every((h) =>
-            headers.includes(h)
+          ![config.columns.persistent_id, "proposed_action"].every(
+            (h, index) => headers.includes(h) || headers.indexOf(h) === index
           )
         ) {
           raiseErrorToast(
-            `The CSV file did not have the ${config.columns.persistent_id} and proposed_action column headers to update the proposed action.`
+            `The CSV file either did not have the ${config.columns.persistent_id} and proposed_action column headers, or the columns are not in the correct order to update the proposed action.`
           );
           return null;
         }
-
         const utterancePatch = rows.map((row) => {
           const [persistentId, dataAction] = row.split(",");
           return { persistentId, dataAction } as UtterancePatch;
         });
         updateDataAction({
           utterancePatch,
-          utteranceQuery: getUtterancesQueryState,
+          ...getUtterancesQueryState,
         });
       }
     };
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -2,7 +2,7 @@ import { ArrowDropDown, GetApp, SvgIconComponent } from "@mui/icons-material";
 import FilterAltOutlinedIcon from "@mui/icons-material/FilterAltOutlined";
 import MultilineChartIcon from "@mui/icons-material/MultilineChart";
 import UploadIcon from "@mui/icons-material/Upload";
-import { Box, Button, Menu, MenuItem } from "@mui/material";
+import { Box, Button, Menu, MenuItem, Modal, Typography } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import {
   GridCellParams,
@@ -85,6 +85,16 @@ const useStyles = makeStyles((theme) => ({
     flexDirection: "row",
     justifyContent: "space-between",
   },
+  modal: {
+    position: "absolute",
+    width: "20",
+    height: "15%",
+    padding: "10px",
+    backgroundColor: theme.palette.background.paper,
+    top: "50%",
+    left: "50%",
+    transform: "translate(-50%, -50%)",
+  },
   searchContainer: {
     marginLeft: theme.spacing(2),
     alignItems: "center",
@@ -166,7 +176,8 @@ const UtterancesTable: React.FC<Props> = ({
       })) ?? [],
     [utterancesResponse]
   );
-
+  const [modal, showModal] = React.useState<boolean>(false);
+  const [modalMessage, showModalMessage] = React.useState<string>("");
   const [selectedPersistentIds, setSelectedPersistentIds] = React.useState<
     number[]
   >([]);
@@ -404,52 +415,83 @@ const UtterancesTable: React.FC<Props> = ({
 
   const searchString = constructSearchString(pipeline);
 
+  const removeSpecialCharacters = (text: string) =>
+    text.replace(/[^a-zA-Z0-9_]/g, "");
+
+  const parseCSVData = (data: string) => {
+    const headers: string[] = data.slice(0, data.indexOf("\n")).split(",");
+    const headersWithoutSpecialCharacters: string[] = headers.map((h) =>
+      removeSpecialCharacters(h)
+    ); // This line is to remove all the extra space and special characters if any except underscore, alphabets and numbers.
+    const validHeaders = ["row_idx", "proposed_action"].map((header) =>
+      headersWithoutSpecialCharacters.includes(header)
+    );
+    if (validHeaders.some((valid) => !valid)) {
+      showModalMessage(
+        "The CSV file did not have the row_idx and proposed_action column headers to update the proposed action."
+      );
+      showModal(true);
+    } else {
+      const parsedData: UtterancePatch[] = data
+        .slice(data.indexOf("\n") + 1)
+        .split("\n")
+        .map((row) => {
+          if (row) {
+            const values = row
+              .split(",")
+              .map((r) => removeSpecialCharacters(r));
+            const obj = headersWithoutSpecialCharacters.reduce(
+              (object, header, index) => {
+                return Object.assign(object, {
+                  [API_HEADERS[header as keyof typeof API_HEADERS]]:
+                    removeSpecialCharacters(values[index]),
+                });
+              },
+              {}
+            );
+            return obj as UtterancePatch;
+          } else {
+            showModalMessage("There are no records in the CSV file.");
+            showModal(true);
+            return {} as UtterancePatch;
+          }
+        });
+      Object.keys(parsedData[0]).length > 0 && updateProposedAction(parsedData);
+    }
+  };
+
+  const updateProposedAction = (data: UtterancePatch[]) => {
+    const groupByPersistentId: Record<
+      string,
+      UtterancePatch["persistentId"][]
+    > = data.reduce(
+      (
+        accumulated: Record<string, UtterancePatch["persistentId"][]>,
+        { persistentId, dataAction }: UtterancePatch
+      ) => ({
+        ...accumulated,
+        [dataAction]: accumulated[dataAction]
+          ? accumulated[dataAction].concat(persistentId)
+          : [persistentId],
+      }),
+      {}
+    );
+    Object.entries(groupByPersistentId).map(([dataAction, persistentIds]) => {
+      const newValue = dataAction as DataAction;
+      updateDataAction({
+        persistentIds,
+        newValue,
+        ...getUtterancesQueryState,
+      });
+    });
+  };
+
   const importProposedActions = (file: File) => {
     const fileReader = new FileReader();
-    fileReader.onload = (event) => {
-      if (event.target) {
-        const data = event.target.result as string;
-        const headers: string[] = data.slice(0, data.indexOf("\n")).split(",");
-        const parsedCSVData: UtterancePatch[] = data
-          .slice(data.indexOf("\n") + 1)
-          .split("\n")
-          .map((row) => {
-            const values = row.split(",");
-            const obj = headers.reduce((object, header, index) => {
-              return Object.assign(object, {
-                [API_HEADERS[
-                  header.replace(/[^a-zA-Z_]/g, "") as keyof typeof API_HEADERS
-                ]]: values[index].replace(/[^a-zA-Z0-9_]/g, ""),
-              });
-            }, {});
-            return obj as UtterancePatch;
-          });
-
-        const groupByPersistentId: Record<
-          string,
-          UtterancePatch["persistentId"][]
-        > = parsedCSVData.reduce(
-          (
-            accumulated: Record<string, UtterancePatch["persistentId"][]>,
-            { persistentId, dataAction }: UtterancePatch
-          ) => ({
-            ...accumulated,
-            [dataAction]: accumulated[dataAction]
-              ? accumulated[dataAction].concat(persistentId)
-              : [persistentId],
-          }),
-          {}
-        );
-        Object.entries(groupByPersistentId).map(
-          ([dataAction, persistentIds]) => {
-            const newValue = dataAction as DataAction;
-            updateDataAction({
-              persistentIds,
-              newValue,
-              ...getUtterancesQueryState,
-            });
-          }
-        );
+    fileReader.onload = ({ target }) => {
+      if (target) {
+        const result = target.result as string;
+        parseCSVData(result);
       }
     };
     fileReader.readAsText(file, "UTF8");
@@ -466,6 +508,28 @@ const UtterancesTable: React.FC<Props> = ({
 
   return (
     <Box className={classes.gridContainer}>
+      <Modal open={modal} onClose={() => showModal(false)}>
+        <Box className={classes.modal}>
+          <Box
+            display="flex"
+            flexDirection="column"
+            alignItems="center"
+            gap={1}
+          >
+            <Typography variant="body1">{modalMessage}</Typography>
+            <Button
+              size="small"
+              sx={{ width: 8 }}
+              variant="contained"
+              onClick={() => {
+                showModal(false);
+              }}
+            >
+              Ok
+            </Button>
+          </Box>
+        </Box>
+      </Modal>
       <div className={classes.gridHeaderActions}>
         <Description
           text="Explore utterances and propose actions. Click on a row to inspect the utterance details."
@@ -482,9 +546,9 @@ const UtterancesTable: React.FC<Props> = ({
               hidden
               accept=".csv, application/vnd.openxmlformats-officedocument.spreadsheetml.sheet, application/vnd.ms-excel"
               type="file"
-              onChange={(event) => {
-                if (event.target.files) {
-                  importProposedActions(event.target.files[0]);
+              onChange={({ target: { files } }) => {
+                if (files) {
+                  importProposedActions(files[0]);
                 }
               }}
             />
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -400,34 +400,6 @@ const UtterancesTable: React.FC<Props> = ({
 
   const searchString = constructSearchString(pipeline);
 
-  const updateProposedAction = (data: UtterancePatch[]) => {
-    const groupByPersistentId: Record<
-      string,
-      UtterancePatch["persistentId"][]
-    > = data.reduce(
-      (
-        accumulated: Record<string, UtterancePatch["persistentId"][]>,
-        { persistentId, dataAction }: UtterancePatch
-      ) => ({
-        ...accumulated,
-        [dataAction]: accumulated[dataAction]
-          ? accumulated[dataAction].concat(persistentId)
-          : [persistentId],
-      }),
-      {}
-    );
-    Object.entries(groupByPersistentId).forEach(
-      ([dataAction, persistentIds]) => {
-        const newValue = dataAction as DataAction;
-        updateDataAction({
-          persistentIds,
-          newValue,
-          ...getUtterancesQueryState,
-        });
-      }
-    );
-  };
-
   const importProposedActions = (file: File) => {
     const fileReader = new FileReader();
     fileReader.onload = ({ target }) => {
@@ -454,7 +426,10 @@ const UtterancesTable: React.FC<Props> = ({
           const [persistentId, dataAction] = row.split(",");
           return { persistentId, dataAction } as UtterancePatch;
         });
-        updateProposedAction(utterancePatch);
+        updateDataAction({
+          utterancePatch,
+          utteranceQuery: getUtterancesQueryState,
+        });
       }
     };
     fileReader.readAsText(file);
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -422,6 +422,7 @@ const UtterancesTable: React.FC<Props> = ({
           return { persistentId, dataAction } as UtterancePatch;
         });
         updateDataAction({
+          ignoreNotFound: true,
           body,
           ...getUtterancesQueryState,
         });
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -2,7 +2,7 @@ import { ArrowDropDown, GetApp, SvgIconComponent } from "@mui/icons-material";
 import FilterAltOutlinedIcon from "@mui/icons-material/FilterAltOutlined";
 import MultilineChartIcon from "@mui/icons-material/MultilineChart";
 import UploadIcon from "@mui/icons-material/Upload";
-import { Box, Button, Menu, MenuItem, Modal, Typography } from "@mui/material";
+import { Box, Button, Dialog, Menu, MenuItem, Typography } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import {
   GridCellParams,
@@ -57,10 +57,6 @@ import { getUtteranceIdTooltip } from "utils/getUtteranceIdTooltip";
 import { constructSearchString, isPipelineSelected } from "utils/helpers";
 
 const SMART_TAG_WIDTH = 30;
-const API_HEADERS = {
-  row_idx: "persistentId",
-  proposed_action: "dataAction",
-};
 
 const useStyles = makeStyles((theme) => ({
   hoverableDataCell: {
@@ -86,14 +82,7 @@ const useStyles = makeStyles((theme) => ({
     justifyContent: "space-between",
   },
   modal: {
-    position: "absolute",
-    width: "20",
-    height: "15%",
-    padding: "10px",
-    backgroundColor: theme.palette.background.paper,
-    top: "50%",
-    left: "50%",
-    transform: "translate(-50%, -50%)",
+    padding: theme.spacing(1),
   },
   searchContainer: {
     marginLeft: theme.spacing(2),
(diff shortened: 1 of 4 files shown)


webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
   ];
 
   const searchString = constructSearchString(pipeline);
+
+  const removeSpecialCharacters = (text: string) =>
+    text.replace(/[^a-zA-Z0-9_]/g, "");
```
JosephMarinier, line 419: Can you explain why we have to remove special characters? How about if the persistent ids do contain special characters? I think it will fail.
For your information, this regular expression seems to be equivalent to this shortcut:
```suggestion
    text.replace(/\W/g, "");
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -51,6 +57,10 @@ import { getUtteranceIdTooltip } from "utils/getUtteranceIdTooltip";
 import { constructSearchString, isPipelineSelected } from "utils/helpers";
 
 const SMART_TAG_WIDTH = 30;
+const API_HEADERS = {
+  row_idx: "persistentId",
```
JosephMarinier, line 61: It's not necessarily `row_idx`. The `persistent_id` in the config is the name of a column in the dataset containing an id that would persist throughout dataset versions. It defaults to `"row_idx"`, but the user can specify anything, for example `"incident_number"`.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
...
+      showModalMessage(
+        "The CSV file did not have the row_idx and proposed_action column headers to update the proposed action."
+      );
+      showModal(true);
+    } else {
+      const parsedData: UtterancePatch[] = data
+        .slice(data.indexOf("\n") + 1)
+        .split("\n")
+        .map((row) => {
+          if (row) {
```
JosephMarinier, line 439: ...and thanks to the `.trim()` I am suggesting, there shouldn't be an empty row anymore.
```suggestion
```

JosephMarinier (commented):
I don't think you need any `slice()` for that. Here are some suggestions...

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
...
+                });
+              },
+              {}
+            );
+            return obj as UtterancePatch;
+          } else {
+            showModalMessage("There are no records in the CSV file.");
+            showModal(true);
+            return {} as UtterancePatch;
+          }
```
JosephMarinier, line 457: Thanks to the `.trim()` I am suggesting, there shouldn't be an empty row anymore, so no need for this `if`.
Also, I think parsing the row can be simplified to something like this:
```suggestion
            const [persistentId, dataAction] = row.split(",");
            return { persistentId, dataAction };
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
...
+        });
+      Object.keys(parsedData[0]).length > 0 && updateProposedAction(parsedData);
```
JosephMarinier, line 459: Maybe we should toast an error if there are no rows in the CSV? Instead of ignoring silently. What do you think?

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
...
+
+  const importProposedActions = (file: File) => {
+    const fileReader = new FileReader();
+    fileReader.onload = ({ target }) => {
+      if (target) {
+        const result = target.result as string;
+        parseCSVData(result);
+      }
+    };
+    fileReader.readAsText(file, "UTF8");
```
JosephMarinier, line 497: UTF-8 is the default, so we can clean that up:
```suggestion
    fileReader.readAsText(file);
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -401,54 +508,95 @@ const UtterancesTable: React.FC<Props> = ({
 
   return (
     <Box className={classes.gridContainer}>
+      <Modal open={modal} onClose={() => showModal(false)}>
```
JosephMarinier, line 511: The visual is not really polished.
image
I see in MUI documentation:
> If you are creating a modal dialog, you probably want to use the [Dialog](https://mui.com/material-ui/react-dialog/) component rather than directly using Modal. Modal is a lower-level construct [...]

I think that would solve most problems you were trying to solve with `classes.modal`, and more.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -401,54 +508,95 @@ const UtterancesTable: React.FC<Props> = ({
...
+        <Box display="flex" alignItems="center" gap={2}>
+          <Button
+            className={classes.exportButton}
+            component="label"
+            startIcon={<UploadIcon />}
+          >
+            Import
+            <input
+              hidden
+              accept=".csv, application/vnd.openxmlformats-officedocument.spreadsheetml.sheet, application/vnd.ms-excel"
```
JosephMarinier, line 547: Where did you take that? I don't think we want to accept Excel sheets or anything other than `.csv`. The export comes from Azimuth and we know it has a `.csv` extension.
```suggestion
              accept=".csv"
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
...
+      updateDataAction({
+        persistentIds,
+        newValue,
+        ...getUtterancesQueryState,
+      });
+    });
+  };
+
+  const importProposedActions = (file: File) => {
+    const fileReader = new FileReader();
```
JosephMarinier, line 490: I notice that if you import a CSV, then click the `Import` button a second time but then `Cancel`, we get this error:
```error
UtterancesTable.tsx:498 Uncaught TypeError: Failed to execute 'readAsText' on 'FileReader': parameter 1 is not of type 'Blob'.
    at importProposedActions (UtterancesTable.tsx:498:1)
    at onChange (UtterancesTable.tsx:552:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4070:1)
    at executeDispatch (react-dom.development.js:8243:1)
    at processDispatchQueueItemsInOrder (react-dom.development.js:8275:1)
    at processDispatchQueue (react-dom.development.js:8288:1)
    at dispatchEventsForPlugins (react-dom.development.js:8299:1)
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+    );
+  };
+
+  const importProposedActions = (file: File) => {
+    const fileReader = new FileReader();
+    fileReader.onload = ({ target }) => {
+      if (target) {
+        const result = target.result as string;
+        const [header, ...rows] = result.split("\n");
+        if (rows.join()) {
```
JosephMarinier, line 441: Any reason for joining back the array into a string? Or can you simply check the length of the array?
```suggestion
        if (rows.length) {
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+          const headers: string[] = header.trim().split(",");
+          if (
+            [config.columns.persistent_id, "proposed_action"].every((h) =>
+              headers.includes(h)
```
JosephMarinier, line 445: In the `map()` below, we assume the columns are in the right order, so I think this should check it.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+            )
+          ) {
+            const utterancePatch: UtterancePatch[] = rows.map((row) => {
+              const [persistentId, dataAction] = row.trim().split(",");
```
JosephMarinier, line 449: Do you have an example why we need to `.trim()` here? I would not have expected that.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+              return { persistentId, dataAction } as UtterancePatch;
+            });
+            updateProposedAction(utterancePatch);
+          } else {
+            showDialogMessage(
+              "The CSV file did not have the row_idx and proposed_action column headers to update the proposed action."
```
JosephMarinier, line 455: Again, we are not necessarily expecting `row_idx`.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+            [config.columns.persistent_id, "proposed_action"].every((h) =>
+              headers.includes(h)
+            )
+          ) {
+            const utterancePatch: UtterancePatch[] = rows.map((row) => {
+              const [persistentId, dataAction] = row.trim().split(",");
+              return { persistentId, dataAction } as UtterancePatch;
+            });
+            updateProposedAction(utterancePatch);
+          } else {
```
JosephMarinier, line 453: Instead of nesting all those `if`/`else`, could you please check the different conditions for error in a linear fashion and return early of there is an error? That is much easier to read and takes up a lot less indentation.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -401,54 +476,95 @@ const UtterancesTable: React.FC<Props> = ({
...
-                jobId,
-                datasetSplitName,
-                ...filters,
-                ...pipeline,
-              });
-              setAnchorEl(null);
-            }}
+        <Box display="flex" alignItems="center" gap={2}>
+          <Button
+            className={classes.exportButton}
```
JosephMarinier, line 508: Not that you added `justifyContent: "space-between"` to the parent, this class is no longer useful. You can remove its definition and the two places it's used.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+        const result = target.result as string;
+        const [header, ...rows] = result.split("\n");
+        if (rows.join()) {
+          const headers: string[] = header.trim().split(",");
+          if (
+            [config.columns.persistent_id, "proposed_action"].every((h) =>
+              headers.includes(h)
+            )
+          ) {
+            const utterancePatch: UtterancePatch[] = rows.map((row) => {
```
JosephMarinier, line 448: With the ` as UtterancePatch` bellow, this one is redundant:
```suggestion
            const utterancePatch = rows.map((row) => {
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+            updateProposedAction(utterancePatch);
+          } else {
+            showDialogMessage(
+              "The CSV file did not have the row_idx and proposed_action column headers to update the proposed action."
+            );
+            showDialog(true);
+          }
+        } else {
+          showDialogMessage("There are no records in the CSV file.");
+          showDialog(true);
```
JosephMarinier, line 461: Now that I see where the `Dialog` box is used, it seems to only be showing error messages. Right? Have you considered using `raiseErrorToast` instead? It seems more aligned with the rest of the app.
```suggestion
          raiseErrorToast("There are no records in the CSV file");
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +403,68 @@ const UtterancesTable: React.FC<Props> = ({
...
+      }
+    );
+  };
+
+  const importProposedActions = (file: File) => {
+    const fileReader = new FileReader();
+    fileReader.onload = ({ target }) => {
+      if (target) {
+        const result = target.result as string;
+        const [header, ...rows] = result.split("\n");
```
JosephMarinier, line 440: You forgot the `.trim()` from my suggestion [here](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1118934022). Anyway, I have a more rigorous solution that gives the same result: `.slice(0, -1)`. Otherwise, there is an empty line at the end of `rows`, resulting in an error when parsing that line.
```suggestion
        const [header, ...rows] = result.split("\n").slice(0, -1);
```

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: I guess the reason is the error I mentioned [here](https://github.com/ServiceNow/azimuth/pull/465/files#r1119895819). If you fix it there, you will be able to apply my suggestion here.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
   ];
 
   const searchString = constructSearchString(pipeline);
+
+  const removeSpecialCharacters = (text: string) =>
+    text.replace(/[^a-zA-Z0-9_]/g, "");
+
+  const parseCSVData = (data: string) => {
+    const headers: string[] = data.slice(0, data.indexOf("\n")).split(",");
```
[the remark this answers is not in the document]
nandhinibsn: I can explain why did I ignore "trim() before doing the split. It did remove extra space but it did not remove other escape characters.

For eg: The csv records contain \r characters. So, I added trim individually to headers and rows like below:
const headers: string[] = header.trim().split(",");
const [persistentId, dataAction] = row.trim().split(",");

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: The reason I had to join was that it created an empty string array after the split. So rows.length is calculated as 1, making the if condition true. So I had to join().

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Yes, true. slice would solve the problem here. I will add the suggestion. Thank you for suggesting it :)

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Sure, I can give you an example: After doing split individually each row, there are still escape characters. So, trim() helped here to remove those.

image

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: So, do you suggest we use the "raiseErrorToast" for both scenarios?

JosephMarinier: That's what I am contemplating, yes. What do you think?

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Can you give me an example of an "escape character"? I don't see any.

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Yes, I understood. 👍

webapp/src/components/Analysis/UtterancesTable.tsx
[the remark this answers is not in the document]
JosephMarinier: There doesn't seem to be `\r`s in the CSV files we generate, just `\n`s. For example, I see:
```js
{result: 'row_idx,proposed_action\n4,augment_with_similar\n5,relabel\n'}
```
Maybe there are `\r`s if the Azimuth back end runs on Windows? 🤷‍♂️ To support that, I would suggest dealing with it (optionally) when we split the lines, using a regular expression:
```diff
-        const [header, ...rows] = result.split("\n").slice(0, -1);
+        const [header, ...rows] = result.split(/\r?\n/).slice(0, -1);
```
With that, there shouldn't be the need for `trim()`ing every row.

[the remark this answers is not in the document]
nandhinibsn: Yeah, it works with this regex. thank you :)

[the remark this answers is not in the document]
nandhinibsn: I am unsure why there is a \r, but mine is on mac.

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Sorry. its carriage return character (\r) I meant to denote here.

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Do you mean to say "exportButton" is not needed anymore.?

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Yes, you are right; showing an error message is an excellent idea using toast.

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Yes, this class `classes.exportButton` is no longer useful.

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Have you tested your recent changes?
Don't forget the early returns in case of errors, otherwise the execution continues and it will call `updateProposedAction()` with garbage data.

nandhinibsn: True, It did not trigger the API call as it was empty array but the code was executed. I fixed that in my recent commit. Please check.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +414,89 @@ const UtterancesTable: React.FC<Props> = ({
...
+        ...accumulated,
+        [dataAction]: accumulated[dataAction]
+          ? accumulated[dataAction].concat(persistentId)
+          : [persistentId],
+      }),
+      {}
+    );
+    Object.entries(groupByPersistentId).map(([dataAction, persistentIds]) => {
+      const newValue = dataAction as DataAction;
+      updateDataAction({
```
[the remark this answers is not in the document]
nandhinibsn: I was able to fix the issue I have been facing and changes are pushed. Now, the API should accept UtterancePatch[] .

webapp/src/services/api.ts
```diff
@@ -195,25 +195,21 @@ export const api = createApi({
...
       {
-        persistentIds: UtterancePatch["persistentId"][];
-        newValue: DataAction;
-      } & GetUtterancesQueryState
+        utterancePatch: {
+          persistentId: UtterancePatch["persistentId"];
+          dataAction: DataAction;
+        }[];
+        utteranceQuery: GetUtterancesQueryState;
+      }
```
JosephMarinier, line 203: I think it was perfect with ` & GetUtterancesQueryState`, instead of nesting it, so it's uniform with the other endpoints.
```suggestion
      } & GetUtterancesQueryState
```

webapp/src/services/api.ts
```diff
@@ -195,25 +195,21 @@ export const api = createApi({
     updateDataActions: build.mutation<
       UtterancePatch[],
       {
-        persistentIds: UtterancePatch["persistentId"][];
-        newValue: DataAction;
-      } & GetUtterancesQueryState
+        utterancePatch: {
+          persistentId: UtterancePatch["persistentId"];
+          dataAction: DataAction;
+        }[];
```
JosephMarinier, line 201: That seems redundant, instead:
```suggestion
        utterancePatch: UtterancePatch[];
```
Also, I think that should be called `body` directly:
```suggestion
        body: UtterancePatch[];
```

JosephMarinier (changes requested):
With the two following suggestions, I think you should be able to simply this bit:
```diff
-      queryFn: async ({ utterancePatch, utteranceQuery }) =>
-        responseToData(
+      queryFn: responseToData(
          fetchApi({
            path: "/dataset_splits/{dataset_split_name}/utterances",
            method: "patch",
          }),
          "Something went wrong updating proposed actions"
-        )({ ...utteranceQuery, body: utterancePatch }),
+        ),
```

webapp/src/services/api.ts
```diff
@@ -225,17 +221,17 @@ export const api = createApi({
...
-                utterance.dataAction = newValue;
-              }
-            });
-          })
+          api.util.updateQueryData(
+            "getUtterances",
+            args.utteranceQuery,
+            (draft) => {
+              Object.assign(draft, utterancePatch);
+            }
```
JosephMarinier, line 233: This optimistic update doesn't work. Can you fix it and try it to make sure it works?
The `draft` is of type
```ts
    {
      utterances: Utterance[];
      utteranceCount: number;
      confidenceThreshold: number | null;
    }
```
while `utterancePatch` is of type
```ts
    UtterancePatch[]
```
so `Object.assign(draft, utterancePatch);` produces garbage.
I think the previous code, with the `forEach()`, is closer to what we need.

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: Any progress on this?

nandhinibsn: Hey Joseph, I started working on this today morning. I derive a solution to it and work towards it. I will be able to push it in sometime

webapp/src/services/api.ts
nandhinibsn: Ohh right! Thank you for pointing out to me that. Please check my recent commit and let me know if that looks ok.

webapp/src/services/api.ts
nandhinibsn: I considered the above and simplified my recent commit as much as possible. But, I face some errors I am unsure of when I remove the async with"queryFn" and call the utterancePatch as body directly. But isn't the term "utterancePatch" more readable than calling body? Or I am unsure if there is any specific reason to call it that way.

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Please check my latest commit and let me know what do you think of that approach.

webapp/src/services/api.ts
JosephMarinier: This is the "request body" as seen in the generated types.
```ts
    requestBody: {
      content: {
        "application/json": components["schemas"]["UtterancePatch"][];
      };
    };
```
I called it `body` in our helper `fetchApi` to align with the native `fetch()` API, which expects `body`.
I don't pretend it is a better name, just that it allows for that code simplification.
This mutation is similar to the config mutation that has a `body`: `updateConfig({ jobId, body: partialConfig })}`.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +399,41 @@ const UtterancesTable: React.FC<Props> = ({
...
+        if (rows.length === 0) {
+          raiseErrorToast("There are no records in the CSV file.");
+          return null;
+        }
+        const headers: string[] = header.split(",");
+        if (
+          ![config.columns.persistent_id, "proposed_action"].every(
+            (h, index) => headers.includes(h) || headers.indexOf(h) === index
+          )
+        ) {
```
JosephMarinier, line 418: This makes no sense. I guess you meant `&&` instead of `||`, otherwise this will still pass if the headers are out of order, which is the problem I raised [here](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1119560156). In that case, I think `headers.includes(h)` is useless.
Why not simplify to this?
```suggestion
        if (headers !== `${config.columns.persistent_id},proposed_action`) {
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +399,41 @@ const UtterancesTable: React.FC<Props> = ({
...
+          raiseErrorToast(
+            `The CSV file either did not have the ${config.columns.persistent_id} and proposed_action column headers, or the columns are not in the correct order to update the proposed action.`
```
JosephMarinier, line 420: I think it would more straightforward to state what we expect, instead of what might be wrong. Here is a suggestion:
```suggestion
            `The CSV file must have column headers ${config.columns.persistent_id} and proposed_action, in that order.`
```

webapp/src/services/api.ts
```diff
@@ -225,15 +217,18 @@ export const api = createApi({
...
             draft.utterances.forEach((utterance) => {
-              if (persistentIds.includes(utterance.persistentId)) {
-                utterance.dataAction = newValue;
-              }
+              Object.assign(
+                utterance,
+                utterancePatch.filter(
+                  ({ persistentId }) => persistentId === utterance.persistentId
+                )
+              );
```
JosephMarinier, line 231: Follow up on [this](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1126572158).
That still doesn't work; `utterance` is a single utterance, while `utterancePatch.filter()` returns an array of utterances. The result of `Object.assign()` will be garbage: a combination of the original utterance and the added field `"0": <the whole utterancePatch>`.
I would suggest avoiding using `Object.assign()` as it is not strongly typed and lets you do mistakes like that.
And again, can you fix the optimistic update and **try it** to make sure it works before you commit/push? Ask questions if you need help.

webapp/src/services/api.ts
JosephMarinier: Ah-ha! I think I know what issue you faced. The type `GetUtterancesQueryState` has a `body: never` field, so when renaming this to `body`, in combination with the other suggestion, you will have to change
```diff
-      } & GetUtterancesQueryState
+      } Omit<GetUtterancesQueryState, "body">
```

webapp/src/services/api.ts
```diff
@@ -195,25 +195,17 @@ export const api = createApi({
     updateDataActions: build.mutation<
       UtterancePatch[],
       {
-        persistentIds: UtterancePatch["persistentId"][];
-        newValue: DataAction;
+        utterancePatch: UtterancePatch[];
       } & GetUtterancesQueryState
     >({
-      queryFn: async ({ jobId, datasetSplitName, persistentIds, newValue }) =>
+      queryFn: async ({ utterancePatch, ...utteranceQuery }) =>
```
JosephMarinier, line 201: Most of the fields from `utteranceQuery` are not allowed for the `PATCH`. They are used in the `args` for `api.util.updateQueryData("getUtterances", args, ...` to perform the right optimistic update.
This should be reverted to using only `jobId` and `datasetSplitName`, as well as `utterancePatch`/`body`, and ignore the rest.
```suggestion
      queryFn: async ({ jobId, datasetSplitName, body }) =>
```

webapp/src/services/api.ts
```diff
@@ -195,25 +195,17 @@ export const api = createApi({
...
           "Something went wrong updating proposed actions"
-        )({
-          jobId,
-          datasetSplitName,
-          body: persistentIds.map((persistentId) => ({
-            persistentId,
-            dataAction: newValue,
-          })),
-        }),
+        )({ ...utteranceQuery, body: utterancePatch }),
```
JosephMarinier, line 208: Accordingly:
```suggestion
        )({ jobId, datasetSplitName, body }),
```

webapp/src/services/api.ts
nandhinibsn: Okay, Thank you for explaining to me why we use the term "body". It makes sense to my confusion now. 

Exactly, that was the error message I have been facing since last week. Thank you for the suggestion to solve the issue here.

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Ohh yeah, this approach is so simple. I couldn't think of it this way earlier. thank you :)

webapp/src/services/api.ts
```diff
@@ -224,16 +216,16 @@ export const api = createApi({
...
-        { dispatch, queryFulfilled }
-      ) {
+      async onQueryStarted({ body, ...args }, { dispatch, queryFulfilled }) {
         const patchResult = dispatch(
           api.util.updateQueryData("getUtterances", args, (draft) => {
             draft.utterances.forEach((utterance) => {
-              if (persistentIds.includes(utterance.persistentId)) {
-                utterance.dataAction = newValue;
-              }
+              return {
```
JosephMarinier, line 223: Follow up on [this](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1126572158) and [this](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1126975146).
That still doesn't work.
A `return` inside a `forEach` doesn't do anything. Anyway, as you can see in the documentation of [`updateQueryData()`](https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#updatequerydata:~:text=The%20callback%20receives%20an%20Immer%2Dwrapped%20draft%20of%20the%20current%20state%2C%20and%20may%20modify%20the%20draft%20to%20match%20the%20expected%20results%20after%20the%20mutation%20completes%20successfully.), we have to
> modify the `draft`

not return a new object. That's why we were assigning to `utterance.dataAction = newValue;`.

For the third time, can you try your solution to make sure it works before you commit/push? You should see the value in the dropdown menu change instantly, instead of waiting for the response from the back end. Ask questions if you need help.

webapp/src/services/api.ts
```diff
@@ -224,16 +216,16 @@ export const api = createApi({
...
+                ...utterance,
+                ...body.filter(
+                  ({ persistentId }) => persistentId === utterance.persistentId
+                )[0],
```
JosephMarinier, line 227: This `.filter(...)[0]` gives the same result as (but in a less efficient way than) `.find(...)`.
```suggestion
                ...body.find(
                  ({ persistentId }) => persistentId === utterance.persistentId
                ),
```

webapp/src/services/api.ts
```diff
@@ -224,15 +216,15 @@ export const api = createApi({
...
         const patchResult = dispatch(
           api.util.updateQueryData("getUtterances", args, (draft) => {
             draft.utterances.forEach((utterance) => {
-              if (persistentIds.includes(utterance.persistentId)) {
-                utterance.dataAction = newValue;
+              const patchUtterance = body.find(
+                ({ persistentId }) => persistentId === utterance.persistentId
+              );
+              if (patchUtterance) {
+                utterance.dataAction = patchUtterance.dataAction;
```
JosephMarinier, line 227: The name `patchUtterance` sounds like a verb and hints at a function. The name `utterancePatch`, matching the type of `body` would have been better. However, I think `found = body.find(...)` sounds even better, especially with `if (found)`. Do you like that?
```suggestion
              const found = body.find(
                ({ persistentId }) => persistentId === utterance.persistentId
              );
              if (found) {
                utterance.dataAction = found.dataAction;
```

webapp/src/components/Utterance/UtteranceDataAction.tsx
```diff
@@ -24,9 +24,14 @@ const UtteranceDataAction: React.FC<Props> = ({
...
+    const utterancePatch: UtterancePatch[] = persistentIds.map(
+      (persistentId) => ({
+        persistentId,
+        dataAction: newValue,
+      })
+    );
     updateDataAction({
-      persistentIds,
-      newValue,
+      body: utterancePatch,
```
JosephMarinier, line 34: I find it confusing that `utterancePatch` is in fact an array of `utterancePatch`es, plural. I think in this case I would simply call it body to avoid renaming it here.
```suggestion
      body,
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +399,37 @@ const UtterancesTable: React.FC<Props> = ({
...
+          );
+          return null;
+        }
+
+        const utterancePatch = rows.map((row) => {
+          const [persistentId, dataAction] = row.split(",");
+          return { persistentId, dataAction } as UtterancePatch;
+        });
+        updateDataAction({
+          body: utterancePatch,
```
JosephMarinier, line 425: Same here
```suggestion
        const body = rows.map((row) => {
          const [persistentId, dataAction] = row.split(",");
          return { persistentId, dataAction } as UtterancePatch;
        });
        updateDataAction({
          body,
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +399,37 @@ const UtterancesTable: React.FC<Props> = ({
...
+        const [header, ...rows] = result.split(/\r?\n/).slice(0, -1);
+        if (rows.length === 0) {
+          raiseErrorToast("There are no records in the CSV file.");
+          return null;
+        }
+        if (header !== `${config.columns.persistent_id},proposed_action`) {
+          raiseErrorToast(
+            `The CSV file must have column headers ${config.columns.persistent_id} and proposed_action, in that order.`
+          );
+          return null;
```
JosephMarinier, line 417: `fileReader.onload` is not supposed to return anything, so I would remove those `null`s.
```suggestion
          return;
        }
        if (header !== `${config.columns.persistent_id},proposed_action`) {
          raiseErrorToast(
            `The CSV file must have column headers ${config.columns.persistent_id} and proposed_action, in that order.`
          );
          return;
```

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +399,37 @@ const UtterancesTable: React.FC<Props> = ({
   ];
 
   const searchString = constructSearchString(pipeline);
+
+  const importProposedActions = (file: File) => {
+    const fileReader = new FileReader();
+    fileReader.onload = ({ target }) => {
+      if (target) {
+        const result = target.result as string;
+        const [header, ...rows] = result.split(/\r?\n/).slice(0, -1);
```
JosephMarinier, line 408: I'm sorry for changing my mind again, but the more I think about it, the more I like my original suggestion of trimming the whole `result` before splitting. The current approach of discarding the last row with `.slice(0, -1)` would silently lose a row if there were no newline at the end of the file, which might happen if the user manually edits the file, and would result in losing data silently. I'd use `.trimEnd()` as I don't expect extra newlines at the beginning of the file, only at the end. Does that make sense?
```suggestion
        const [header, ...rows] = result.trimEnd().split(/\r?\n/);
```

webapp/src/services/api.ts
```diff
@@ -224,15 +216,15 @@ export const api = createApi({
...
-        { dispatch, queryFulfilled }
-      ) {
+      async onQueryStarted({ body, ...args }, { dispatch, queryFulfilled }) {
         const patchResult = dispatch(
           api.util.updateQueryData("getUtterances", args, (draft) => {
             draft.utterances.forEach((utterance) => {
-              if (persistentIds.includes(utterance.persistentId)) {
-                utterance.dataAction = newValue;
+              const patchUtterance = body.find(
+                ({ persistentId }) => persistentId === utterance.persistentId
```
JosephMarinier, line 224: Follow up on [this](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1126572158), [this](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1126975146) and [this](https://github.com/ServiceNow/azimuth/pull/465#discussion_r1128000847).
There is still an issue. Have you tried it with a CSV file to make sure it works?
The `utterance.persistentId` can be a `string` or a `number`. However, when reading the CSV file, there is no distinction between a `string` and a `number`, so we assume `persistentId` is a string. The back end supports that, no problem. Here, we just have to check equality with `==` instead of `===` to support that.
```suggestion
                // Support number | string
                // eslint-disable-next-line eqeqeq
                ({ persistentId }) => persistentId == utterance.persistentId
```

nandhinibsn: Yes, it did work for me with the sample import of data. It's good to use equality to avoid the type-checking, but I wonder why the type would be changed in the CSV

JosephMarinier: There is no typing in CSV. There is no way to tell if `...,42,...` is the number `42` or the string `"42"`. Does that make sense?

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Otherwise, we should check if there is any new line at the end to do a slice(0,-1).

webapp/src/services/api.ts
nandhinibsn: Sounds good to me. how about foundPersistenId or foundId?

JosephMarinier: It would be a `foundUtterancePatch`, but I don't think this extra information is necessary, given the scope of this variable is so small.

webapp/src/services/api.ts
nandhinibsn: Yes :)

JosephMarinier: Please add the comments I suggested, otherwise we get a warning in the terminal.

webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -390,6 +399,37 @@ const UtterancesTable: React.FC<Props> = ({
   ];
 
   const searchString = constructSearchString(pipeline);
+
+  const importProposedActions = (file: File) => {
+    const fileReader = new FileReader();
+    fileReader.onload = ({ target }) => {
```
JosephMarinier, line 405: Bug: When importing a CSV file of proposed actions, importing the same file path twice fails silently. The file browser closes normally but the `fileReader.onload` is never called.
Use case: One might have edited the file and wants to re-import it.
Workaround: Refresh the app.
Let's fix that in another PR. I filed a bug: #479

JosephMarinier (approved):
Very cool feature! 👍

webapp/src/components/Analysis/UtterancesTable.tsx
nandhinibsn: Ohh, Let me check that use case. Thanks for finding that out and creating a follow-up issue for it.

nandhinibsn:
> Very cool feature! 👍

I appreciate your patience in reviewing it and your input. :) Thank you.

PR #467, opened 2023-02-25
Refactor FastAPI tags

JosephMarinier:
## Description:

Those tags organize the sections of the Swagger UI.

* Change Admin to Config
* Move tags definitions to the app level
* Also, remove "v1" for every tag as we don't have any other versions of the API. If one day we do, we can add v1 back.

image

## Checklist:

  our users.
* [x] **DEV CHANGES.**


gabegma (approved):
Good idea! For the `v1`, I remember in the past suggesting to remove the `v1` folder (which is different than what you are doing here) and getting some pushback from the team that it was best practice to keep the `v1` references. Do you know if that would apply to tags as well?

JosephMarinier:
I don't mind removing the `v1` folder.
The thing that would have been a good practice (and I guess that's where the confusion came from) would have been to have `/v1` in the API paths (aligning with the current folder structure). This way, if we have to do a big breaking change in say `/v1/example`, we can add a `/v2/example` and let clients transition progressively from v1 (deprecated) to v2.
That being said, our users don't really use our API directly (unlike Toolkit for example) and we control our only client: the front end. So, I don't think this `/v1` thing is necessary.

PR #472, opened 2023-03-02
Load config history

JosephMarinier:
## Description:

* [Make config path optional](https://github.com/ServiceNow/azimuth/pull/472/commits/cf501775419cd1101cf216fc27707783c9a30cbe), defaulting to loading config from env vars only, for example:
```shell
make ARTIFACT_PATH=cache REJECTION_CLASS=NO_INTENT DATASET='{
    "class_name": "loading_resources.load_CLINC150_data",
    "remote": "'$PWD'/azimuth_shr",
    "kwargs": {
        "python_loader": "'$PWD'/azimuth_shr/data/CLINC150.py",
        "full_path": "'$PWD'/azimuth_shr/files/clinc-demo/oos-eval/data/clinc_data_dummy.json"
    }
}' launch.local
```
* [Support launching Azimuth from config history](https://github.com/ServiceNow/azimuth/pull/472/commits/296324bfca33bc363c733d56452ed3b9bdf2da43)
* [Change `artifact_path` default to `"cache"`](https://github.com/ServiceNow/azimuth/pull/472/commits/bf8861aeb2700461ce224f2ac8098f293d1aed53)
* [Make dataset config optional](https://github.com/ServiceNow/azimuth/pull/472/commits/dd94ef2f53cf5e607524381a474f4720028e3c9a), relying on our (already existing) code to validate that the dataset is defined after loading the final config.

image

## Checklist:

  our users.
* [x] **DEV CHANGES.**


azimuth/config.py
```diff
@@ -435,10 +449,19 @@ def load_azimuth_config(config_path: str) -> AzimuthConfig:
...
+    if load_config_history:
+        config_history_path = cfg.get_config_history_path()
+        try:
+            with jsonlines.open(config_history_path, mode="r") as config_history:
+                *_, last_config = config_history
+        except (FileNotFoundError, ValueError):
+            log.info("Empty or invalid config history.")
+        else:
+            log.info(f"Loading latest config from {config_history_path}.")
+            cfg = AzimuthConfig.parse_obj(last_config)
```
gabegma, line 464: Should we try and except here too, and load `cfg` otherwise?

docs/docs/development/launching.md
```diff
@@ -65,6 +65,22 @@ From this point, the back end will launch and compute the start-up task.
 * You can consult the openapi documentation at `localhost:8091/docs`. From there, you can consult the API documentation and try out the different endpoints. This can be useful for debugging.
 * Note that the back end will not reload automatically based on code changes.
 
+After a successful start, Azimuth saves the provided config in its `config_history.jsonl` artifact. If you use the API to edit the config, the edits are saved there. If you restart Azimuth (for example after shutting it down for the night), you can resume where you left off with `LOAD_CONFIG_HISTORY=1`. In order to find the right `config_history.jsonl`, Azimuth needs the same `ARTIFACT_PATH`, so to resume the above example, it needs to be specified:
+```shell
+make ARTIFACT_PATH=cache LOAD_CONFIG_HISTORY=1 launch.local
+```
+In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together, in which case Azimuth will automatically
```
gabegma, line 72:
```suggestion
In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together, in which case Azimuth will automatically derive the `artifact_path` from the `CFG_PATH` to find the config history, if available. Azimuth will then:
```
Is that accurate?

docs/docs/development/launching.md
```diff
@@ -65,6 +65,22 @@ From this point, the back end will launch and compute the start-up task.
 * You can consult the openapi documentation at `localhost:8091/docs`. From there, you can consult the API documentation and try out the different endpoints. This can be useful for debugging.
 * Note that the back end will not reload automatically based on code changes.
 
+After a successful start, Azimuth saves the provided config in its `config_history.jsonl` artifact. If you use the API to edit the config, the edits are saved there. If you restart Azimuth (for example after shutting it down for the night), you can resume where you left off with `LOAD_CONFIG_HISTORY=1`. In order to find the right `config_history.jsonl`, Azimuth needs the same `ARTIFACT_PATH`, so to resume the above example, it needs to be specified:
```
gabegma, line 68: We will need to add these details to the user section too: https://servicenow.github.io/azimuth/main/getting-started/c-run/

docs/docs/development/launching.md
```diff
@@ -65,6 +65,22 @@ From this point, the back end will launch and compute the start-up task.
...
+
+1. load the config from `CFG_PATH` when it first starts (if `config_history.jsonl` is empty); and
+2. load the config from `config_history.jsonl` from then on if it is restarted.
+
+Although confusing, this comes in very handy for a docker container that might be stopped and restarted with the same command. For example:
+```shell
+make CFG_PATH=local_configs/development/clinc_dummy/conf.json \
+     LOAD_CONFIG_HISTORY=1 \
+     launch.local
+```
```
gabegma, line 82: Very clear, thank you!

gabegma (commented):
Thank you for taking the time to make a clear PR. I was hesitant, but I see it more clearly now. I like it. So the default is to ignore the history, right? So no confusion if someone does `make CFG_PATH='bla.json' launch`; they will see 'bla' in Azimuth, no matter the config history.

azimuth/config.py
JosephMarinier: I'm not sure I see in which (normal) use case the latest config from `config_history.jsonl` would be invalid, so I'm not sure what the appropriate response would be, but I'm afraid it would be confusing to silently ignore the error and load another config. I am leaning towards letting the exception kill the process, just like `AzimuthConfig.parse_file(config_path)` does if the config is invalid.

To be clear, the `try`/`except` above is there to catch if the `config_history.jsonl` doesn't exist yet or is empty, which is normal on the first launch. Unfortunately, it also catches (`jsonlines` raises the same type of `ValueError` 😞) if `config_history.jsonl` contains invalid JSON (not even considering the `AzimuthConfig` schema). I don't see a (normal) use case for that to happen, so I think I would prefer to crash the process in that case too, but it's somewhat out of my control. We can revisit this later if we suspect people tend to mess up with their `config_history.jsonl` and run into this situation.

Does that make sense?

docs/docs/development/launching.md
```diff
@@ -65,6 +65,22 @@ From this point, the back end will launch and compute the start-up task.
 * You can consult the openapi documentation at `localhost:8091/docs`. From there, you can consult the API documentation and try out the different endpoints. This can be useful for debugging.
 * Note that the back end will not reload automatically based on code changes.
 
+After a successful start, Azimuth saves the provided config in its `config_history.jsonl` artifact. If you use the API to edit the config, the edits are saved there. If you restart Azimuth (for example after shutting it down for the night), you can resume where you left off with `LOAD_CONFIG_HISTORY=1`. In order to find the right `config_history.jsonl`, Azimuth needs the same `ARTIFACT_PATH`, so to resume the above example, it needs to be specified:
+```shell
+make ARTIFACT_PATH=cache LOAD_CONFIG_HISTORY=1 launch.local
```
JosephMarinier, line 70: In our Python code, we could consider making `"cache"` the default artifact path, useful locally, rather than `"/cache"`, useful in docker, and set it to `"/cache"` in our `Dockerfile`.
```suggestion
make LOAD_CONFIG_HISTORY=1 launch.local
```

docs/docs/development/launching.md
JosephMarinier: Yes! The user might still set an env var `ARTIFACT_PATH`, and not in `CFG_PATH`, but that might be an insignificant detail. Here is a try anyway:
```suggestion
In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together. Azimuth will derive the `artifact_path` from `CFG_PATH` and `ARTIFACT_PATH` to find the right `config_history.jsonl`. Then, Azimuth will automatically:
```

docs/docs/development/launching.md
```diff
@@ -65,6 +65,22 @@ From this point, the back end will launch and compute the start-up task.
...
+```
+In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together, in which case Azimuth will automatically
+
+1. load the config from `CFG_PATH` when it first starts (if `config_history.jsonl` is empty); and
+2. load the config from `config_history.jsonl` from then on if it is restarted.
```
JosephMarinier, line 75: I think I'll clarify that "it" is "Azimuth" here.
```suggestion
2. load the config from `config_history.jsonl` from then on (if Azimuth is restarted).
```

azimuth/config.py
gabegma: Yes, that makes sense. I was thinking of the use case when we introduce breaking changes when Azimuth updates, but even then, it's probably best that user deletes their cache/config history anyway .

docs/docs/development/launching.md
gabegma: Looks good!

docs/docs/development/launching.md
gabegma: Yeah, looks like a good idea

azimuth/config.py
JosephMarinier: Oh right, that's a good one. I'll keep thinking of what do to with breaking changes in the config. For now, let's simply be extra vigilant about not introducing breaking changes in the config, if possible.

docs/docs/development/launching.md
JosephMarinier: I like it. It also makes the documentation more straightforward; a simpler description of the "normal" use case with the default `artifact_path`, and then a warning box about customizing `artifact_path`.

image

azimuth/app.py
```diff
@@ -96,7 +97,7 @@ def start_app(config_path, debug=False) -> FastAPI:
 
     log.info("🔭 Azimuth starting 🔭")
 
-    azimuth_config = load_azimuth_config(config_path)
+    azimuth_config = load_azimuth_config(config_path, load_config_history)
     if azimuth_config.dataset is None:
         raise ValueError("No dataset has been specified in the config.")
```
JosephMarinier, line 102: I made `dataset` `Optional` in the config, but we were already checking it is not `None` here. This was useless before, but now I rely on it. I did that so that `AzimuthConfig()` returns a valid config (without raising an exception), therefore supporting `make LOAD_CONFIG_HISTORY=1 launch.local` without a `CFG_PATH`.

docs/docs/development/launching.md
JosephMarinier: Thank you. In this commit, I [Document LOAD_CONFIG_HISTORY in Getting Started section](https://github.com/ServiceNow/azimuth/pull/472/commits/eba396781a41637805cda9431f987513ecc55c5f).

azimuth/config.py
```diff
@@ -286,8 +286,11 @@ def to_hash(self):
 class CommonFieldsConfig(ProjectConfig, extra=Extra.ignore):
     """Fields that can be modified without affecting caching."""
 
-    # Where to store artifacts. (HDF5 files,  HF datasets, Dask config)
-    artifact_path: str = "/cache"
+    artifact_path: str = Field(
+        "cache",
+        description="Where to store artifacts "
+        "(Azimuth config history, HDF5 files, HF datasets, Dask config).",
```
gabegma, line 292: what is Dask config?

docs/docs/development/launching.md
```diff
@@ -65,22 +65,28 @@ From this point, the back end will launch and compute the start-up task.
...
 
 1. load the config from `CFG_PATH` when it first starts (if `config_history.jsonl` is empty); and
 2. load the config from `config_history.jsonl` from then on if it is restarted.
 
-Although confusing, this comes in very handy for a docker container that might be stopped and restarted with the same command. For example:
+Although confusing, this comes in very handy in some contexts like a docker container that might be stopped and restarted with the same command. For example:
 ```shell
 make CFG_PATH=local_configs/development/clinc_dummy/conf.json \
      LOAD_CONFIG_HISTORY=1 \
      launch.local
```
gabegma, line 81: In a Docker container, we wouldn't use `launch.local` right?

docs/docs/getting-started/c-run.md
```diff
@@ -108,6 +108,20 @@ dataset and model are available in `config/examples` (`CLINC` is also shown belo
 4. The **app will be accessible** at `http://localhost:8080` after a few minutes of waiting. The
    start-up tasks will start.
 
+After a successful start, Azimuth saves the provided config in its `config_history.jsonl` artifact. If you use the API to edit the config, the edits are saved there. If you restart Azimuth (for example after shutting it down for the night), you can resume where you left off with:
+```shell
+make LOAD_CONFIG_HISTORY=1 launch
+```
+In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together, in which case Azimuth will automatically
```
gabegma, line 115: Could you also add a bullet to the environment variable section below? That section (env var) is not great, we should probably reformulate it at some point, but to be consistent, `LOAD_CONFIG_HISTORY=1` should also be there.

azimuth/config.py
JosephMarinier: I don't know.

docs/docs/development/launching.md
JosephMarinier: Correct. In our docker containers, we don't use `make`.
In my commit to [Document LOAD_CONFIG_HISTORY in Getting Started section](https://github.com/ServiceNow/azimuth/pull/472/commits/eba396781a41637805cda9431f987513ecc55c5f), where we use docker, I give those examples with `launch` instead of `launch.local`.

docs/docs/getting-started/c-run.md
JosephMarinier: Done in [Document LOAD_CONFIG_HISTORY in Environment variables](https://github.com/ServiceNow/azimuth/pull/472/commits/32d3849ffe0c573b617e45f66a9cf40cee65dd48)

azimuth/config.py
JosephMarinier: Do you think this information is outdated? I can remove it.

gabegma: Yeah I would remove that last bit.

docs/docs/development/launching.md
gabegma: So perhaps we should remove it here too since this looks like an example for a Docker container? At least this was my interpretation.

docs/docs/getting-started/c-run.md
gabegma: Thank you!!

docs/docs/development/launching.md
JosephMarinier: This is an example that might be useful during local "development". While I am developing this feature, I am running exactly this command. I thought it might be useful to others in the future, but it might be rarer.
The following warning admonition refers to this example
> specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together, like in the example above"

so do you think I should move this example inside the admonition?

azimuth/config.py
JosephMarinier: Done in [Remove outdated doc about Dask config](https://github.com/ServiceNow/azimuth/pull/472/commits/fb726c9a6dc4962bdf00ee1d064a50a55e9f30f8)

docs/docs/development/launching.md
gabegma: Are we talking about the same one? I'm talking about the one that follows "like a docker container that might be stopped and restarted with the same command. For example:"

JosephMarinier: Yes

JosephMarinier: I think I get what is confusing now. This example is not only about docker, it is about this whole section saying that "it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together". I am just mentioning that "this comes in very handy in some contexts like a docker container".
So, I think I should move this mention about docker after the example, to avoid this confusion. Does that make sense?
``````diff
diff --git a/docs/docs/development/launching.md b/docs/docs/development/launching.md
index 8ecee59f..06b7eb0f 100644
--- a/docs/docs/development/launching.md
+++ b/docs/docs/development/launching.md
@@ -74,12 +74,13 @@ In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH`
 1. load the config from `CFG_PATH` when it first starts (if `config_history.jsonl` is empty); and
 2. load the config from `config_history.jsonl` from then on (if Azimuth is restarted).

-Although confusing, this comes in very handy in some contexts like a docker container that might be stopped and restarted with the same command. For example:
+For example:
 ```shell
 make CFG_PATH=local_configs/development/clinc_dummy/conf.json \
      LOAD_CONFIG_HISTORY=1 \
      launch.local
 ```
+Although confusing, this comes in very handy in some contexts like a docker container that might be stopped and restarted with the same command.

 ??? warning "Using `LOAD_CONFIG_HISTORY=1` with a custom `artifact_path`"
     The location of `config_history.jsonl` depends on `artifact_path`, so if you customized it in your config, it needs to be consistent when you restart Azimuth. The easiest solution is to always specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH` together, like in the example above. Alternatively, you can specify the environment variable `ARTIFACT_PATH`:
diff --git a/docs/docs/getting-started/c-run.md b/docs/docs/getting-started/c-run.md
index 40eeaebe..d097e156 100644
--- a/docs/docs/getting-started/c-run.md
+++ b/docs/docs/getting-started/c-run.md
@@ -117,10 +117,11 @@ In fact, it is possible to specify both `LOAD_CONFIG_HISTORY=1` and a `CFG_PATH`
 1. load the config from `CFG_PATH` when it first starts (if `config_history.jsonl` is empty); and
 2. load the config from `config_history.jsonl` from then on (if Azimuth is restarted).

-Although confusing, this enables you to stop and restart the docker container with the same command. For example:
+For example:
 ```shell
 make CFG_PATH=/config/my_project/conf.json LOAD_CONFIG_HISTORY=1 launch
 ```
+Although confusing, this enables you to stop and restart the docker container with the same command.
``````

gabegma: yes, that works!! Thank you!

gabegma (approved):
LGTM!! Awesome new feature, great work!

PR #474, opened 2023-03-02
Validate config min/max values

JosephMarinier:
## Description:

Very simple and repetitive PR about validating the min and max values to the number fields in the config.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -115,7 +115,7 @@ class TemperatureScaling(CustomObject):
     class_name: Literal[
         "azimuth.utils.ml.postprocessing.TemperatureScaling"
     ] = "azimuth.utils.ml.postprocessing.TemperatureScaling"
-    temperature: float = Field(1, env="TEMP")
+    temperature: float = Field(1, ge=0, env="TEMP")
 
     @root_validator()
     def check_temps(cls, values):
@@ -131,7 +131,7 @@ class ThresholdConfig(CustomObject):
     class_name: Literal[
         "azimuth.utils.ml.postprocessing.Thresholding"
     ] = "azimuth.utils.ml.postprocessing.Thresholding"
-    threshold: float = Field(0.5, env="TH")
+    threshold: float = Field(0.5, ge=0, le=1, env="TH")
 
     @root_validator()
     def check_threshold(cls, values):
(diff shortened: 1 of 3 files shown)


azimuth/config.py
```diff
@@ -407,16 +411,15 @@ class AzimuthConfig(
...
-        syntax.spacy_model = syntax.spacy_model or defaults.spacy_model
-        syntax.subj_tags = syntax.subj_tags or defaults.subj_tags
-        syntax.obj_tags = syntax.obj_tags or defaults.obj_tags
-        if values["similarity"]:
-            similarity = values["similarity"]
+        if syntax := values.get("syntax"):
+            syntax.spacy_model = syntax.spacy_model or defaults.spacy_model
+            syntax.subj_tags = syntax.subj_tags or defaults.subj_tags
+            syntax.obj_tags = syntax.obj_tags or defaults.obj_tags
+        if similarity := values.get("similarity"):
```
JosephMarinier, line 422: I changed `values["similarity"]` to `values.get("similarity")` (as well as the others) because the individual fields get validated by Pydantic first, before calling this `@root_validator()`. If Pydantic raises a `ValidationError` while validating `similarity` (like if `no_close_threshold` is outside the min/max range), it caries on with the rest of the validation to get an exhaustive list of validation errors, so we end up here with `similarity` that doesn't exist. In this context, `values["similarity"]` was producing a no-helpful `KeyError` (and not a `ValidationError`) that was "hiding" the validation error.

gabegma: Thanks for the explanation!

azimuth/config.py
```diff
@@ -407,16 +411,15 @@ class AzimuthConfig(
...
+            neutral_token = behavioral_testing.neutral_token
             neutral_token.prefix_list = neutral_token.prefix_list or defaults.prefix_list
             neutral_token.suffix_list = neutral_token.suffix_list or defaults.suffix_list
-        syntax = values["syntax"]
-        syntax.spacy_model = syntax.spacy_model or defaults.spacy_model
-        syntax.subj_tags = syntax.subj_tags or defaults.subj_tags
-        syntax.obj_tags = syntax.obj_tags or defaults.obj_tags
-        if values["similarity"]:
-            similarity = values["similarity"]
+        if syntax := values.get("syntax"):
```
gabegma, line 418: Even though syntax is never optional at the moment?

azimuth/config.py
```diff
@@ -220,16 +220,20 @@ class BehavioralTestingOptions(AzimuthBaseSettings):
...
 
 class SimilarityOptions(AzimuthBaseSettings):
-    faiss_encoder: str = ""  # Language-based dynamic default value
-    # Threshold to use when finding conflicting neighbors.
-    conflicting_neighbors_threshold: float = 0.9
-    # Threshold to determine whether there are close neighbors.
-    no_close_threshold: float = 0.5
+    faiss_encoder: str = Field("", description="Language-based dynamic default value.")
+    conflicting_neighbors_threshold: float = Field(
+        0.9, ge=0, le=1, description="Threshold to use when finding conflicting neighbors."
```
gabegma, line 225: Great idea to convert comments into descriptions!

gabegma: I think we should do it throughout maybe? But it can be in a future PR.

gabegma (approved):
LGTM!

azimuth/config.py
JosephMarinier: Yes, for the reason I gave [here](https://github.com/ServiceNow/azimuth/pull/474#discussion_r1123826242).

gabegma: Ah ok. if I rephrase, do you mean that if there is a problem with one syntax field (like the value being out of bound), the syntax options will be set to None or something like that, which will then break the old line 418, hiding the original error 
with the syntax field? If so, can we think of some comments to add so we don't forget? Or is basically your test testing that we have the appropriate error message?

JosephMarinier: Yes, "hiding the original error" with a `KeyError` on old line 418.
My test ensures that we get the expected `ValidationError`, which is the only type of exception pydantic is supposed to return. A `KeyError` makes the test fail.

gabegma: Awesome, I get it now :)

PR #478, opened 2023-03-08
Support dataset-only smart tag analysis

### LINKED ISSUE #433: Make smart tag aggregation plot work for dataset only

JosephMarinier:
Resolve #433

## Description:

The overall diff is pretty messy, so I recommend looking at the different commits individually.

I made a couple of decisions what we can debate:
- Color the dataset-only bars (with no outcomes) dark blue (`palette.secondary.dark`) like other data visualizations.
- Hide the pipeline-specific columns, including accuracy and some smart tag families.
- Disable the `Prediction` and `Outcome` options from the menu (since hiding the options would result in a weird one-option menu).

Here is an example:
Screen Shot 2023-03-08 at 11 45 51 AM

For reference, here is how it looks with a pipeline:
Screen Shot 2023-03-08 at 11 46 04 AM

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/SmartTagsTable.tsx
+++ b/webapp/src/components/SmartTagsTable.tsx
@@ -38,7 +38,10 @@ import { formatRatioAsPercentageString } from "utils/format";
 import { constructSearchString } from "utils/helpers";
 import DatasetSplitToggler from "./Controls/DatasetSplitToggler";
 
+type MetricPerFilterOption = "label" | "prediction" | "outcome";
+
 type SortBy = "filterValue" | "utteranceCount" | "accuracy" | SmartTagFamily;
+
 type Row = {
   filterValue: string;
   utteranceCount: number;
@@ -64,7 +67,7 @@ const SmartTagsTable: React.FC<{
   const [transpose, setTranspose] = React.useState(false);
 
   const [selectedMetricPerFilterOption, setSelectedMetricPerFilterOption] =
-    React.useState<"label" | "prediction" | "outcome">("label");
+    React.useState<MetricPerFilterOption>("label");
 
   const [ascending, setAscending] = React.useState(false);
   const [sortBy, setSortBy] = React.useState<SortBy>("utteranceCount");
@@ -457,7 +460,7 @@ const SmartTagsTable: React.FC<{
               value={selectedMetricPerFilterOption}
               onChange={(event) =>
                 setSelectedMetricPerFilterOption(
-                  event.target.value as "label" | "prediction"
+                  event.target.value as MetricPerFilterOption
                 )
               }
             >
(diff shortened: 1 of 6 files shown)


gabegma:
Awesome!!! A few comments on your questions:
- I'm ok with the blue, but I think the color should be the same one as in the control panel when there is no pipeline. So either both are blue, or we pick the same dark grey (black?) for this viz too. Side note - is the same dark blue as in the dataset warnings for example?
- I'm not sure about moving the tile before the pipeline quality. I absolutely get the reasoning, but I do feel when you have a pipeline selected, you would care/want to see the overall quality first. Curious what others think.
- The rest is great.

webapp/src/components/SmartTagsTable.tsx
```diff
@@ -339,99 +453,49 @@ const SmartTagsTable: React.FC<{
...
-                      </Tooltip>
-                    ))}
+                        ))
+                      ) : (
+                        <Bar
+                          rowCount={row.utteranceCount}
+                          cellCount={cell.utteranceCount}
+                          filterValue={row.filterValue}
+                          family={family}
+                          bgcolor={(theme) => theme.palette.secondary.dark}
```
JosephMarinier, line 495: @gabegma
> I'm ok with the blue, but I think the color should be the same one as in the control panel when there is no pipeline. So either both are blue, or we pick the same dark grey (black?) for this viz too.

You are reading my mind! So, I initially used the same color as the filter distribution bars here, which is the dark purple from the top navigation bar (`palette.primary.main`), but I thought it looked like a classified military document with some redacted text. 😅 
Screen Shot 2023-03-08 at 3 53 57 PM
compared to
Screen Shot 2023-03-08 at 11 45 51 AM
I would like to change the filter distribution bars to blue, as it occurs to me now that the (almost) black could look like disabled.
Screen Shot 2023-03-08 at 3 58 01 PM
compared to
Screen Shot 2023-03-08 at 3 57 36 PM
which I think better fits our colorful esthetics
Screen Shot 2023-03-08 at 3 57 44 PM

@gabegma
> is the same dark blue as in the dataset warnings for example?

Yes, it's the same as `DataViz2`: `#006DB3`.

Does that make sense?

gabegma: Yes - I'm all in for this change!

webapp/src/pages/Dashboard.tsx
```diff
@@ -138,32 +138,30 @@ const Dashboard = () => {
...
+                jobId={jobId}
+                pipeline={pipeline}
+                availableDatasetSplits={datasetInfo.availableDatasetSplits}
+                datasetSplitName={datasetSplitName}
+                setDatasetSplitName={setDatasetSplitName}
+              />
+            </Box>
+          </PreviewCard>
+        )}
+      />
```
JosephMarinier, line 164: Nothing changed here, except removing the `{isPipelineSelected(pipeline) && (...)}` around the card. I hate how Git deals with changes of indent.

christyler3030 (approved):

PR #480, opened 2023-03-09
Make API errors uniform

JosephMarinier:
## Description:

Don't worry, this is less than 50 lines of code if we ignore the generated types.

We used to have 3 different schemas of errors (as follows).
1. The `HTTPException`s we raise in our code usually have a `detail` string that is used to produce a JSON response. This stays the same. For example:
```json
{
  "detail": "Dataset split not found."
}
```
2. The `ValidationError`s FastAPI raises have a harder-to-read JSON response. For example
```json
{
  "detail": [
    {
      "loc": [
        "query",
        "pipeline_index"
      ],
      "msg": "ensure this value is greater than or equal to 0",
      "type": "value_error.number.not_ge",
      "ctx": {
        "limit_value": 0
      }
    }
  ]
}
```
This PR transforms it to
```json
{
  "detail": "1 validation error for Request\nquery -> pipeline_index\n  ensure this value is greater than or equal to 0 (type=value_error.number.not_ge; limit_value=0)"
}
```
so we can use the `detail` in the front end:
```
1 validation error for Request
query -> pipeline_index
  ensure this value is greater than or equal to 0 (type=value_error.number.not_ge; limit_value=0)
```
3. The other exceptions are handled as unexpected "internal server error" and the response is plain text (not JSON):
```
Internal Server Error
```
This PR transforms it to
```json
{
  "detail": "Internal Server Error"
}
```

## Checklist:

  our users.
* [ ] **DEV CHANGES.**


azimuth/app.py
```diff
@@ -31,6 +42,38 @@
...
+    return JSONResponseIgnoreNan(
+        status_code=HTTP_404_NOT_FOUND  # for errors in paths, e.g., /dataset_splits/potato
+        if "path" in (error["loc"][0] for error in exception.errors())
+        else HTTP_400_BAD_REQUEST,  # for other errors like in query params, e.g., pipeline_index=-1
+        content=HTTPExceptionModel(detail=str(exception)).dict(),
+    )
+
+
+async def handle_internal_error(request: Request, exception: Exception):
+    # Don't expose this unexpected internal error as that could expose a security vulnerability.
```
JosephMarinier, line 70: We could also change that behavior and expose our internal errors. Considering our code is open source, that might not be a real threat.

azimuth/app.py
```diff
@@ -31,6 +42,38 @@
...
+    detail: str
+
+
+async def handle_validation_error(request: Request, exception: RequestValidationError):
+    return JSONResponseIgnoreNan(
+        status_code=HTTP_404_NOT_FOUND  # for errors in paths, e.g., /dataset_splits/potato
+        if "path" in (error["loc"][0] for error in exception.errors())
+        else HTTP_400_BAD_REQUEST,  # for other errors like in query params, e.g., pipeline_index=-1
+        content=HTTPExceptionModel(detail=str(exception)).dict(),
+    )
```
JosephMarinier, line 66: I am not sure the `{ "detail": "..." }` JSON object wrapping the error message is useful. I could instead return a plain text error message.
```suggestion
    return PlainTextResponse(
        status_code=HTTP_404_NOT_FOUND  # for errors in paths, e.g., /dataset_splits/potato
        if "path" in (error["loc"][0] for error in exception.errors())
        else HTTP_400_BAD_REQUEST,  # for other errors like in query params, e.g., pipeline_index=-1
        content=str(exception),
    )
```
I would simply add an exception handler for `HTTPException` that does the same thing.
I could remove the 500 exception handler, as it was already returning plain text.
Let me know if you have an opinion on this.

azimuth/app.py
```diff
@@ -31,6 +43,38 @@
...
+    detail: str
+
+
+async def handle_validation_error(request: Request, exception: RequestValidationError):
+    return JSONResponse(
+        status_code=HTTP_404_NOT_FOUND  # for errors in paths, e.g., /dataset_splits/potato
+        if "path" in (error["loc"][0] for error in exception.errors())
+        else HTTP_400_BAD_REQUEST,  # for other errors like in query params, e.g., pipeline_index=-1
+        content={"detail": str(exception)},
+    )
```
JosephMarinier, line 67: I am not sure the `{ "detail": "..." }` JSON object wrapping the error message is useful. I could instead return a plain text error message.
```suggestion
    return PlainTextResponse(
        status_code=HTTP_404_NOT_FOUND  # for errors in paths, e.g., /dataset_splits/potato
        if "path" in (error["loc"][0] for error in exception.errors())
        else HTTP_400_BAD_REQUEST,  # for other errors like in query params, e.g., pipeline_index=-1
        content=str(exception),
    )
```
I would simply add an exception handler for `HTTPException` that does the same thing. I could then remove the 500 exception handler, as it was already returning plain text.
However, FastAPI's `http_exception_handler()` for `HTTPException` is a little bit more sophisticated as it supports setting response headers, so I thought it was cool to keep it as is.
Let me know if you have an opinion on this.

azimuth/app.py
gabegma: Hmm interesting, maybe we start with that, and see if it happens often enough?

gabegma (commented):
I'll let @christyler3030 make the official review, it's a bit over my head :)

PR #482, opened 2023-03-10
Make config UI string fields editable

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -254,6 +254,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
   const [partialConfig, setPartialConfig] = React.useState<
     Partial<AzimuthConfig>
   >({});
+  const isNullString = (value: string) => value === "null";
   const isEmptyPartialConfig = Object.keys(partialConfig).length === 0;
 
   const handleDiscard = () => {
@@ -493,12 +494,12 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
               ...partialConfig,
               [config]: {
                 ...resultingConfig[config],
-                [field]: newValue,
+                [field]: isNullString(newValue) ? null : newValue,
               },
             })
           : setPartialConfig({
               ...partialConfig,
-              [field]: newValue,
+              [field]: isNullString(newValue) ? null : newValue,
             })
       }
     />
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -71,6 +71,7 @@ const FIELDS: Record<
   long_utterance_min_word: { ...INT, units: "words" },
   temperature: FLOAT,
   threshold: PERCENTAGE,
+  seed: INT,
 };
 
 type SubConfigKeys = keyof PickByValue<AzimuthConfig, object | null>;
@@ -863,8 +864,6 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
             displayArgumentsList(field, value)
           ) : typeof value === "object" ? (
             displayKeywordArguments(field, value)
-          ) : field === "seed" ? (
-            displayReadonlyFields(field, value)
           ) : field === "spacy_model" ? (
             <StringField
               select
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -871,28 +871,26 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
                   <React.Fragment key={index}>
                     <Typography variant="body2">{objField}:</Typography>
                     {Array.isArray(objValue) ? (
-                      <Tooltip title={objValue.join(", ")}>
-                        <StringField
-                          key={objField}
-                          value={objValue.join(", ")}
-                          disabled={
-                            isUpdatingConfig ||
-                            (config && !resultingConfig[config])
-                          }
-                          onChange={(newValue) =>
-                            setPartialConfig({
-                              ...partialConfig,
-                              [config]: {
-                                ...resultingConfig[config],
-                                [field]: {
-                                  ...value,
-                                  [objField]: newValue.split(", "),
-                                },
+                      <StringField
+                        key={objField}
+                        value={objValue.join(", ")}
+                        disabled={
+                          isUpdatingConfig ||
+                          (config && !resultingConfig[config])
+                        }
+                        onChange={(newValue) =>
+                          setPartialConfig({
+                            ...partialConfig,
+                            [config]: {
+                              ...resultingConfig[config],
+                              [field]: {
+                                ...value,
+                                [objField]: newValue.split(", "),
                               },
-                            })
-                          }
-                        />
-                      </Tooltip>
+                            },
+                          })
+                        }
+                      />
                     ) : (
                       <NumberField
                         sx={{ width: "10ch" }}
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -43,6 +43,8 @@ import {
   AzimuthConfig,
   PipelineDefinition,
   SupportedLanguage,
+  SupportedModelContract,
+  SupportedSpacyModels,
 } from "types/api";
 import { PickByValue } from "types/models";
 import { UNKNOWN_ERROR } from "utils/const";
@@ -78,6 +80,15 @@ type SubConfigKeys = keyof PickByValue<AzimuthConfig, object | null>;
 const CUSTOM_METRICS: string[] = ["Accuracy", "Precision", "Recall", "F1"];
 const ADDITIONAL_KWARGS_CUSTOM_METRICS = ["Precision", "Recall", "F1"];
 const SUPPORTED_LANGUAGES: SupportedLanguage[] = ["en", "fr"];
+const SUPPORTED_MODEL_CONTRACT: SupportedModelContract[] = [
+  "hf_text_classification",
+  "file_based_text_classification",
+  "custom_text_classification",
+];
+const SUPPORTED_SPACY_MODELS: SupportedSpacyModels[] = [
+  "en_core_web_sm",
+  "fr_core_news_md",
+];
 const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof AzimuthConfig)[] = [
   "behavioral_testing",
   "similarity",
@@ -482,6 +493,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     <StringField
       label={config === "columns" ? "" : field}
       value={String(value)}
+      disabled={isUpdatingConfig}
       onChange={(newValue) =>
         config
           ? setPartialConfig({
@@ -506,8 +518,10 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     value: string | null
   ) => (
     <StringField
+      key={pipelineIndex}
       label={field}
       value={String(value)}
+      disabled={isUpdatingConfig}
       onChange={(newValue) =>
         field === "class_name" || field === "remote"
           ? setPartialConfig({
@@ -657,10 +671,28 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
       {displaySectionTitle("General")}
       <FormGroup>
         <Columns columns={3}>
-          {displayStringOnlyFields(
-            "model_contract",
-            resultingConfig.model_contract
-          )}
+          <FormControl variant="standard" className="fixedWidthInput">
+            <InputLabel id="model-contract-input-label">
+              model_contract
+            </InputLabel>
+            <Select
+              value={resultingConfig.model_contract}
+              disabled={isUpdatingConfig}
+              labelId="model-contract-input-label"
+              onChange={({ target: { value } }) =>
+                setPartialConfig({
+                  ...partialConfig,
+                  model_contract: value as SupportedModelContract,
+                })
+              }
+            >
+              {SUPPORTED_MODEL_CONTRACT.map((modelContract) => (
+                <MenuItem key={modelContract} value={modelContract}>
+                  {modelContract}
+                </MenuItem>
+              ))}
+            </Select>
+          </FormControl>
           {displayStringOnlyFields(
             "saliency_layer",
             resultingConfig.saliency_layer
@@ -831,6 +863,37 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
             displayKeywordArguments(field, value)
           ) : config === "behavioral_testing" ? (
             displayReadonlyFields(field, value)
+          ) : field === "spacy_model" ? (
+            <FormControl
+              key={field}
+              variant="standard"
+              className="fixedWidthInput"
+            >
+              <InputLabel id="spacy-model-input-label">{field}</InputLabel>
+              <Select
+                value={value}
+                disabled={isUpdatingConfig}
+                labelId="spacy-model-input-label"
+                onChange={({ target: { value } }) => {
+                  setPartialConfig({
+                    ...partialConfig,
+                    syntax: {
+                      ...resultingConfig.syntax,
+                      spacy_model: value as SupportedSpacyModels,
+                    },
+                  });
+                }}
+              >
+                <MenuItem value=" ">
+                  <em>None</em>
+                </MenuItem>
+                {SUPPORTED_SPACY_MODELS.map((spacyModel) => (
+                  <MenuItem key={spacyModel} value={spacyModel}>
+                    {spacyModel}
+                  </MenuItem>
+                ))}
+              </Select>
+            </FormControl>
           ) : (
             displayStringOnlyFields(field, value, config)
           )
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -78,7 +78,7 @@ type SubConfigKeys = keyof PickByValue<AzimuthConfig, object | null>;
 const CUSTOM_METRICS: string[] = ["Accuracy", "Precision", "Recall", "F1"];
 const ADDITIONAL_KWARGS_CUSTOM_METRICS = ["Precision", "Recall", "F1"];
 const SUPPORTED_LANGUAGES: SupportedLanguage[] = ["en", "fr"];
-const SUPPORTED_MODEL_CONTRACT: SupportedModelContract[] = [
+const SUPPORTED_MODEL_CONTRACTS: SupportedModelContract[] = [
   "hf_text_classification",
   "file_based_text_classification",
   "custom_text_classification",
@@ -619,7 +619,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
             "rejection_class",
             resultingConfig.rejection_class
           )}
-          {/* <Box display="flex" flexDirection="column">
+          <Box display="flex" flexDirection="column">
             <Typography variant="caption">columns</Typography>
             <KeyValuePairs>
               <React.Fragment key="text_input">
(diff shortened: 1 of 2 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -159,6 +159,33 @@ const displayReadonlyFields = (label: string, value: string | null) => (
...
+    value: string | null;
+    onChange: (newValue: string) => void;
+  }
+> = ({ label, value, onChange, ...props }) => {
+  return (
+    <TextField
+      key={label}
+      size="small"
+      variant="standard"
+      className="fixedWidthInput"
```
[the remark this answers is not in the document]
JosephMarinier: With that, I think it would be best to widen a bit the gaps between the columns:
```diff
 const Columns: React.FC<{ columns?: number }> = ({ columns = 1, children }) => (
-  <Box display="grid" gap={2} gridTemplateColumns={`repeat(${columns}, 1fr)`}>
+  <Box display="grid" gap={4} gridTemplateColumns={`repeat(${columns}, 1fr)`}>
     {children}
   </Box>
 );
```

webapp/src/pages/Settings.tsx
```diff
@@ -159,6 +159,33 @@ const displayReadonlyFields = (label: string, value: string | null) => (
...
 );
 
+const StringField: React.FC<
+  Omit<TextFieldProps, "onChange"> & {
+    label: string;
+    value: string | null;
+    onChange: (newValue: string) => void;
+  }
+> = ({ label, value, onChange, ...props }) => {
+  return (
```
JosephMarinier, line 169: You can remove the braces and the `return` to save one indent (and two lines).
```suggestion
> = ({ label, value, onChange, ...props }) => (
```

webapp/src/pages/Settings.tsx
```diff
@@ -159,6 +159,33 @@ const displayReadonlyFields = (label: string, value: string | null) => (
   />
 );
 
+const StringField: React.FC<
+  Omit<TextFieldProps, "onChange"> & {
+    label: string;
+    value: string | null;
```
JosephMarinier, line 165: We will need to better handle the `null` scenario. Problem:
- Say `value` is `null`.
- The user types `a`.
- The `value` changes to `"nulla"`.
- The user hits backspace.
- The `value` changes to `"null"`, and not `null`, which is probably not what the user wants.

webapp/src/pages/Settings.tsx
```diff
@@ -159,6 +159,33 @@ const displayReadonlyFields = (label: string, value: string | null) => (
...
+      key={label}
+      size="small"
+      variant="standard"
+      className="fixedWidthInput"
+      label={label}
+      value={String(value)}
+      inputProps={{
+        sx: {
+          textOverflow: "ellipsis",
+          ...(value === null && { fontStyle: "italic" }),
```
JosephMarinier, line 180: Given the last comment, this is not currently used.

webapp/src/pages/Settings.tsx
```diff
@@ -159,6 +159,33 @@ const displayReadonlyFields = (label: string, value: string | null) => (
...
+const StringField: React.FC<
+  Omit<TextFieldProps, "onChange"> & {
+    label: string;
+    value: string | null;
+    onChange: (newValue: string) => void;
+  }
+> = ({ label, value, onChange, ...props }) => {
+  return (
+    <TextField
+      key={label}
```
JosephMarinier, line 171: I see an error in the JavaScript console. Can you debug this?
> Warning: Each child in a list should have a unique "key" prop. [...] at StringField

webapp/src/pages/Settings.tsx
```diff
@@ -583,8 +696,10 @@ const Settings: React.FC = () => {
             displayArgumentsList(field, value)
           ) : typeof value === "object" ? (
             displayKeywordArguments(field, value)
-          ) : (
+          ) : config === "behavioral_testing" ? (
             displayReadonlyFields(field, value)
```
JosephMarinier, line 700: That affects only the `seed` field, which should be a number field instead.
```suggestion
```

JosephMarinier: After that, there is only the postprocessor's `class_name` that uses `displayReadonlyFields()`. Any reason? I think it should be a string field too. We will then be able to remove `displayReadonlyFields()` completely.

webapp/src/pages/Settings.tsx
```diff
@@ -320,6 +347,71 @@ const Settings: React.FC = () => {
     />
   );
 
+  const displayStringOnlyFields = (
+    field: string,
+    value: string | null,
+    config?: SubConfigKeys
+  ) => (
+    <StringField
+      label={config === "columns" ? "" : field}
```
JosephMarinier, line 356: Cool! I think the option of having no `label` should be an argument of the function, as opposed to hardcoding `config === "columns"` here. This is the simplest change I can think of, but it might also be possible to be cleaner:
```suggestion
    config?: SubConfigKeys,
    label = field
  ) => (
    <StringField
      label={label}
```
```diff
               {displayStringOnlyFields(
                 "text_input",
                 resultingConfig.columns.text_input,
-                "columns"
+                "columns",
+                ""
               )}
               <Typography variant="body2">label:</Typography>
               {displayStringOnlyFields(
                 "label",
                 resultingConfig.columns.label,
-                "columns"
+                "columns",
+                ""
               )}
```

webapp/src/pages/Settings.tsx
```diff
@@ -320,6 +347,71 @@ const Settings: React.FC = () => {
...
+              pipelines: [
+                ...resultingConfig.pipelines!.slice(0, pipelineIndex),
+                {
+                  ...pipeline,
+                  [field]: newValue,
+                },
+                ...resultingConfig.pipelines!.slice(pipelineIndex + 1),
+              ],
+            })
+      }
```
JosephMarinier, line 411: That's a lot of duplication. Also, the hardcoded `field === "class_name" || field === "remote"` already causes problems with the postprocessor's `class_name`, right? Can you try making a more robust system for the `onChange` handlers?

webapp/src/pages/Settings.tsx
```diff
@@ -583,8 +696,10 @@ const Settings: React.FC = () => {
             displayArgumentsList(field, value)
           ) : typeof value === "object" ? (
             displayKeywordArguments(field, value)
```
JosephMarinier, line 698: As I mentioned on Zoom, this would ideally not be using the same display as the "keyword arguments". All those objects have a well-defined set of keys, whereas the `kwargs` should support (in a later PR) adding and removing keys.

webapp/src/pages/Settings.tsx
nandhinibsn: I needed some clarity on the type of field (whether it will be a plain text field or a mix of select and text fields) because the class_name might have "azimuth.utils.ml.postprocessing.TemperatureScaling" or "azimuth.utils.ml.postprocessing.Thresholding" or custom value. What would be your suggestion on this?

JosephMarinier: Ah! Good question.
I think we could wrap the `TextField` in an `Autocomplete` to suggest the two values we know, with `freeSolo` to let the user input a custom value too.
```ts
      <Autocomplete
        freeSolo
        options={[
          "azimuth.utils.ml.postprocessing.TemperatureScaling",
          "azimuth.utils.ml.postprocessing.Thresholding",
        ]}
        renderInput={(params) => <TextField {...params} />}
      />
```
To really support custom postprocessors, we would need to support `args` and `kwargs` too, instead of those special `temperature` and `threshold` fields, but we can add that in another PR.

webapp/src/pages/Settings.tsx
JosephMarinier: This is especially important in the case of `rejections_class`, which could totally be `"null"`, as well as `null` (meaning there is not rejection class).
The `remote` fields can also be `null`, for example.

webapp/src/pages/Settings.tsx
JosephMarinier: Like `NumberField`, I recommend setting the `key` on `StringField` when calling it, instead of on `TextField` inside here. That would also let you remove the `label: string;` to leave `label` as it is defined in `TextFieldProps`.

webapp/src/pages/Settings.tsx
```diff
@@ -165,6 +177,56 @@ const displayReadonlyFields = (label: string, value: string | null) => (
     }}
   />
 );
+const SelectField: React.FC<
+  Omit<SelectProps, "onChange"> & {
+    label: string;
+    value: any;
+    onChange: (newValue: any) => void;
+    noneValue?: string;
```
JosephMarinier, line 185: Why a string if we never display it? I guess a better name would be `nullable`.
```suggestion
    nullable?: boolean;
```

JosephMarinier: That said, I don't think `spacy_model` is nullable, so we probably should just remove that.

webapp/src/pages/Settings.tsx
```diff
@@ -165,6 +177,56 @@ const displayReadonlyFields = (label: string, value: string | null) => (
...
+    children: React.ReactElement<typeof MenuItem>[];
+  }
+> = ({ label, value, onChange, children, noneValue, ...props }) => (
+  <FormControl variant="standard" className="fixedWidthInput">
```
JosephMarinier, line 189: The `model_contract` and `spacy_model` values don't fit in the `fixedWidthInput`. Maybe just pass `className="fixedWidthInput"` to the language `SelectField`?
```suggestion
  <FormControl variant="standard">
```

webapp/src/pages/Settings.tsx
JosephMarinier: I resolved this in #494

webapp/src/pages/Settings.tsx
JosephMarinier: I resolved this in #494

webapp/src/pages/Settings.tsx
```diff
@@ -165,6 +177,56 @@ const displayReadonlyFields = (label: string, value: string | null) => (
     }}
   />
 );
+const SelectField: React.FC<
```
JosephMarinier, line 180: I made a small PR to save 25 lines by reusing `StringField` instead of creating a new `SelectField`: #494

webapp/src/pages/Settings.tsx
```diff
@@ -77,6 +80,15 @@ type SubConfigKeys = keyof PickByValue<AzimuthConfig, object | null>;
 const CUSTOM_METRICS: string[] = ["Accuracy", "Precision", "Recall", "F1"];
 const ADDITIONAL_KWARGS_CUSTOM_METRICS = ["Precision", "Recall", "F1"];
 const SUPPORTED_LANGUAGES: SupportedLanguage[] = ["en", "fr"];
+const SUPPORTED_MODEL_CONTRACT: SupportedModelContract[] = [
```
JosephMarinier, line 83: We should add an `S`:
```suggestion
const SUPPORTED_MODEL_CONTRACTS: SupportedModelContract[] = [
```

webapp/src/pages/Settings.tsx
JosephMarinier: I did the remaining cleanup in 999d6cf03f767b47f009e67a9ebf37821a6f0629

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, I added "String | null" considering those scenarios. So, you also suggest keeping both types here, right?

webapp/src/pages/Settings.tsx
nandhinibsn: Thank you :)

webapp/src/pages/Settings.tsx
nandhinibsn: Sure. That's perfect.

webapp/src/pages/Settings.tsx
nandhinibsn: Great. Thanks.

webapp/src/pages/Settings.tsx
nandhinibsn: Thank you :)

webapp/src/pages/Settings.tsx
nandhinibsn: Thank you :)

webapp/src/pages/Settings.tsx
JosephMarinier: What I'm saying is that the user need to be able to input `null` (and not just `"null"`).

JosephMarinier: Is there a problem with this? I see your commit 53f093e79f355483ebc32093d03cd52a4693f3d3 made it possible to input `null`, while making it impossible to input `"null"`. That only moves the problem. Again, the user must be able to input `"null"` as well as `null`.

webapp/src/pages/Settings.tsx
JosephMarinier: Will you tackle that in this PR? Basically treating the analyses sub-objects similarly to the `columns` sub-object?

webapp/src/pages/Settings.tsx
```diff
@@ -403,34 +504,51 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+              </React.Fragment>
+              <React.Fragment key="label">
+                <Typography variant="body2">label:</Typography>
+                {displayStringField(
+                  "label",
+                  resultingConfig.columns.label,
+                  "columns",
+                  ""
+                )}
+              </React.Fragment>
```
JosephMarinier, line 534: No need to add those 2 fragments, as well as the `key="txtInput"`. You can revert those changes.

webapp/src/pages/Settings.tsx
JosephMarinier: Don't forget the initial suggestion here about making `seed` a number field. It should be an `INT`.

webapp/src/pages/Settings.tsx
```diff
@@ -604,8 +751,33 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+              }
+            >
+              {SUPPORTED_SPACY_MODELS.map((spacyModel) => (
+                <MenuItem key={spacyModel} value={spacyModel}>
+                  {spacyModel}
+                </MenuItem>
+              ))}
+            </StringField>
+          ) : (
+            displayStringField(field, value, config)
```
JosephMarinier, line 780: Those fields would need to be `disabled` if `!resultingConfig[config]`.

webapp/src/pages/Settings.tsx
```diff
@@ -335,6 +365,77 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+          pipelines: [
+            ...resultingConfig.pipelines!.slice(0, pipelineIndex),
+            {
+              name: subPipeline ? pipeline.name : newValue,
+              model: {
+                ...pipeline.model,
+                ...(subPipeline === "model" && { [field]: newValue }),
+              },
+              postprocessors:
+                postprocessorIdx && subPipeline === "postprocessors"
```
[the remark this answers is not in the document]
nandhinibsn: Let me check this and get back.

webapp/src/pages/Settings.tsx
```diff
@@ -746,7 +747,65 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+                                  [objField]: newValue.split(", "),
+                                },
+                              },
+                            })
+                          }
+                        />
+                      </Tooltip>
+                    ) : (
+                      <NumberField
+                        sx={{ width: "10ch" }}
```
nandhinibsn, line 786: I had to overwrite the width here because the columns became clumsy with no gap. Do you have any better suggestion to handle the gap ?
image

webapp/src/pages/Settings.tsx
```diff
@@ -746,7 +747,65 @@ const Settings: React.FC<Props> = ({ onClose }) => {
           ) : Array.isArray(value) ? (
             displayArgumentsList(field, value)
           ) : typeof value === "object" ? (
-            displayKeywordArguments(field, value)
+            <Box
+              key={field}
+              display="flex"
+              flexDirection="column"
+              marginRight={2}
+            >
```
JosephMarinier, line 755: Let's not add extra margin here, we need as much place as possible for the text.
```suggestion
            <Box key={field} display="flex" flexDirection="column">
```

webapp/src/pages/Settings.tsx
JosephMarinier: Instead, I would modify `fixedWidthInput` like this:
```diff
-          [`& .fixedWidthInput .${inputClasses.root}`]: { width: "12ch" },
+          [`& .fixedWidthInput .${inputClasses.root}`]: { maxWidth: "12ch" },
```
```suggestion
```

webapp/src/pages/Settings.tsx
```diff
@@ -746,7 +747,65 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+                            [config]: {
+                              ...resultingConfig[config],
+                              [field]: {
+                                ...value,
+                                [objField]: newValue,
+                              },
+                            },
+                          })
+                        }
+                        {...FIELDS[field]}
```
JosephMarinier, line 800: I think you meant
```suggestion
                        {...FIELDS[objField]}
```
since `bevavioral_testing` doesn't exist on `FIELDS`. However, the `threshold` fields will conflict with the existing `FIELDS.threshold`, so I think I would simply do
```suggestion
                        {...INT}
```

webapp/src/pages/Settings.tsx
nandhinibsn: Ohh!.. Right. I added to test the column gap but. I forgot to remove it later. thanks for pointing it out.

webapp/src/pages/Settings.tsx
JosephMarinier: You can now remove `nb_typos_per_utterance` from `FIELDS`.

webapp/src/pages/Settings.tsx
JosephMarinier: Oups! I notice the `language` field now has a problem. 😅 I'll look for a solution tomorrow.

JosephMarinier: I think for the language field we have to set a width since it is in a `flex` container. We can take this opportunity to make it smaller, I think it looks better. I propose:
```diff
         <StringField
           select
           label="language"
-          className="fixedWidthInput"
+          sx={{ width: "6ch" }}
```

webapp/src/pages/Settings.tsx
JosephMarinier: Looks like you accidentally reverted your fix when you rebased and [force-pushed](https://github.com/ServiceNow/azimuth/compare/720b8aab50e70e2f8b3f87457758bcdf779e6219..d16611e32c832dfa566d95b6c9b83be9278f99a0#diff-2ee47bc0fd1761ec5486a838d588adc51eb53c7b901fd5c3a6c1ad24ba6f11b2L749-R956).

webapp/src/pages/Settings.tsx
JosephMarinier: Same

webapp/src/pages/Settings.tsx
JosephMarinier: Same

webapp/src/pages/Settings.tsx
JosephMarinier: Precision, it seems you lost the following two commits. You could cherry-pick them to fix the issue.
- d6e167df236d83f1824b73c1696b6397e64eb630
- 720b8aab50e70e2f8b3f87457758bcdf779e6219

JosephMarinier (approved):
Yaaaass! 👍

PR #491, opened 2023-03-15
Add "created-on" in config history

gabegma:
## Description:
See commit messages.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -4,6 +4,7 @@
 
 import argparse
 import os
+from datetime import datetime, timezone
 from enum import Enum
 from os.path import join as pjoin
 from typing import Any, Dict, List, Literal, Optional, TypeVar, Union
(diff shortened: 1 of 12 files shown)


azimuth/config.py
```diff
@@ -305,7 +305,7 @@ class CommonFieldsConfig(ProjectConfig, extra=Extra.ignore):
     # Disable configuration changes
     read_only_config: bool = Field(False, exclude_from_cache=True)
 
-    def get_artifact_path(self) -> str:
+    def get_project_path(self) -> str:
         """Generate a path for caching.
 
         The path contains the project name, the task and a subset of a hash of the project config.
```
gabegma, line 311:
```suggestion
        The path contains the project name and a subset of a hash of the project config.
```

azimuth/config.py
```diff
@@ -461,7 +461,8 @@ def load_azimuth_config(config_path: Optional[str], load_config_history: bool) -
         config_history_path = cfg.get_config_history_path()
         try:
             with jsonlines.open(config_history_path, mode="r") as config_history:
-                *_, last_config = config_history
+                *_, last_config_info = config_history
+                last_config = last_config_info["config"]
```
JosephMarinier, line 465: If the JSON was malformed, that would raise a `KeyError`. I don't know if we want that. If we want to let pydantic parse the JSON and produce the "clear" errors, we can create a class like `AzimuthConfigHistory`. I gave it a try in https://github.com/ServiceNow/azimuth/pull/493. It also has the advantage of defining how to create the datetime string inside the definition of `AzimuthConfigHistory`, thanks to `default_factory`. I don't know what approach is best. What do you think?

JosephMarinier (approved):
Cool! Thank you! 👍

PR #496, opened 2023-03-17
Fix config page checkbox fields click event

### LINKED ISSUE #488: Make config checkboxes' "hit box" smaller
The "hit box" of the config checkboxes extends all the way to the right of the page. This can cause some misclicks, for example when simply trying to focus out of a field after editing it (clicking in the void).

https://user-images.githubusercontent.com/8386369/224879228-48e0e7f7-e75b-484b-a3cf-74fd8e171312.mov

nandhinibsn:
Resolve #488 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -689,11 +689,8 @@ const Settings: React.FC<Props> = ({ onClose }) => {
       <DialogContent
         dividers
         sx={{
-          [`& .${formControlLabelClasses.root}`]: {
-            width: "fit-content",
-          },
+          [`& .${formControlLabelClasses.root}`]: { alignSelf: "start" },
           [`& .${formControlLabelClasses.labelPlacementStart}`]: {
-            justifyContent: "flex-end",
             marginLeft: 0,
           },
           [`& .${formGroupClasses.root}`]: { marginX: 2, marginBottom: 2 },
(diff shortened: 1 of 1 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -689,6 +689,9 @@ const Settings: React.FC<Props> = ({ onClose }) => {
       <DialogContent
         dividers
         sx={{
+          [`& .${formControlLabelClasses.root}`]: {
+            width: "fit-content",
+          },
           [`& .${formControlLabelClasses.labelPlacementStart}`]: {
             justifyContent: "flex-end",
```
JosephMarinier, line 696: I think it would be best to work with alignment instead of width. It might be fine with `width`, but I'm afraid it might cause unexpected behaviors.
Since we are inside a "flex" container (`FormGroup`) with the default `align-items: stretch`, we could overwrite this for the `formControlLabelClasses.root` with `align-self: start`.
What's cool is that this works for `formControlLabelClasses.labelPlacementStart` too, to which we can remove the `justify-content: flex-end`.
```suggestion
          [`& .${formControlLabelClasses.root}`]: { alignSelf: "start" },
          [`& .${formControlLabelClasses.labelPlacementStart}`]: {
```

JosephMarinier (approved):

PR #498, opened 2023-03-17
Stop saving config when loading config history

gabegma:
## Description:
* This will avoid duplicated lines in config history when loading from history. 

## Checklist:

  our users.
* [ ] **DEV CHANGES.**


azimuth/app.py
```diff
@@ -103,7 +103,7 @@ def start_app(config_path: Optional[str], load_config_history: bool, debug: bool
 
     local_cluster = default_cluster(large=azimuth_config.large_dask_cluster)
 
-    run_startup_tasks(azimuth_config, local_cluster)
+    run_startup_tasks(azimuth_config, local_cluster, load_config_history)
```
JosephMarinier, line 106: If we launch Azimuth with `load_config_history` but the config history is empty, it will load the config from `CFG_PATH` and/or from the other env vars. I think it would be best to then save that config in config history, otherwise, it will remain empty. For that, we would need to rely on the logic inside `load_azimuth_config()` to determine if we save to the config history or not.

Also, I just realized, it would be nice to also avoid duplicates if someone kills and restarts Azimuth with the same config, even without `load_config_history`. So, we could brainstorm on a more general solution. Something like checking if the loaded config is different than the last one in the config history.

What do you think?

gabegma: Yes I agree. I didn't think of the first use case, and I already thought it was a shame we didn't cover the second use case. I thought you preferred that we don't load the last one to check if it differs? I guess we don't have much choice in the end?

JosephMarinier: In case we have issues with that, here is what I had in mind to fix only the first use case:
[Have `load_azimuth_config()` return if it `came_from_config_history`](https://github.com/ServiceNow/azimuth/pull/498/commits/6d7235c4963a93907c8c86188aa1e894e0d76a4c)

gabegma: Attempted in #506

gabegma:
Close in favor of #506

PR #500, opened 2023-03-17
Allow user to change dask cluster from route

gabegma:
## Description:
* The user could not change the dask cluster size in the previous implementation of the route. 

## Checklist:

  our users.
* [ ] **DEV CHANGES.**

### DIFF
--- a/azimuth/routers/config.py
+++ b/azimuth/routers/config.py
@@ -1,7 +1,7 @@
 # Copyright ServiceNow, Inc. 2021 – 2022
 # This source code is licensed under the Apache 2.0 license found in the LICENSE file
 # in the root directory of this source tree.
-from typing import Dict
+from typing import Any, Dict
 
 import structlog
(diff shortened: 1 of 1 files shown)


azimuth/routers/config.py
```diff
@@ -98,3 +100,11 @@ def patch_config(
...
     task_manager.clear_worker_cache()
     return new_config
+
+
+def attribute_changed_in_config(
+    attribute: str, partial_config: Dict[str, Any], config: AzimuthConfig
+) -> bool:
+    return attribute in partial_config and partial_config[attribute] != config.__getattribute__(
+        attribute
+    )
```
JosephMarinier, line 110: `__getattribute__()` is the magic method that gets called when using Python’s built-in function `getattr()`. It's very rare that we have to call the magic methods directly.
```suggestion
    return attribute in partial_config and partial_config[attribute] != getattr(config, attribute)
```

PR #505, opened 2023-03-21
Add common fields configuration section and persistentId under the project section

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -236,7 +236,16 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
   >({});
 
   const isEmptyPartialConfig = Object.keys(partialConfig).length === 0;
-  const isNull = (value: string) => (value === "" ? null : value);
+
+  const handleDiscard = () => {
+    setPartialConfig({});
+    setLanguage(undefined);
+  };
+
+  const handleClose = () => {
+    handleDiscard();
+    onClose();
+  };
 
   const handleDiscard = () => {
     setPartialConfig({});
@@ -942,7 +951,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
                 batch_size: newValue,
               })
             }
-            {...FIELDS["batch_size"]}
+            {...INT}
           />
           <StringField
             select
@@ -1046,7 +1055,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
       </AccordionLayout>
       <AccordionLayout
         name="Common Fields Configuration"
-        description="View and edit generic fields adpated based on the user's machine."
+        description="View and edit generic fields that can be adapted based on the user's machine."
         link="reference/configuration/common/"
       >
         {getCommonFieldsConfigSection()}
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -950,7 +950,6 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
             value={String(resultingConfig.use_cuda)}
             disabled={isUpdatingConfig}
             onChange={(newValue: string | boolean) =>
-              newValue &&
               setPartialConfig({
                 ...partialConfig,
                 use_cuda:
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -175,47 +175,16 @@ const StringField: React.FC<
     {...props}
   />
 );
-const SelectField: React.FC<
-  Omit<SelectProps, "onChange"> & {
-    label: string;
-    value: any;
-    onChange: (newValue: any) => void;
-    noneValue?: string;
-    children: React.ReactElement<typeof MenuItem>[];
-  }
-> = ({ label, value, onChange, children, noneValue, ...props }) => (
-  <FormControl variant="standard" className="fixedWidthInput">
-    <InputLabel id={`${label}-input-label`} shrink={true}>
-      {label}
-    </InputLabel>
-    <Select
-      value={value}
-      labelId={`${label}-input-label`}
-      onChange={({ target: { value } }) => onChange(value)}
-      {...props}
-    >
-      {noneValue && (
-        <MenuItem value="">
-          <em>None</em>
-        </MenuItem>
-      )}
-      {children}
-    </Select>
-  </FormControl>
-);
 
 const StringField: React.FC<
   Omit<TextFieldProps, "onChange"> & {
-    label: string;
     value: string;
     onChange: (newValue: string) => void;
   }
-> = ({ label, value, onChange, ...props }) => (
+> = ({ onChange, ...props }) => (
   <TextField
     size="small"
     variant="standard"
-    label={label}
-    value={String(value)}
     inputProps={{
       sx: {
         textOverflow: "ellipsis",
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -156,6 +156,27 @@ const displayArgumentsList = (name: string, args: any[]) => (
   </Box>
 );
 
+const displayReadonlyField = (label: string, value: string | null) => (
+  <TextField
+    key={label}
+    size="small"
+    variant="standard"
+    label={label}
+    InputLabelProps={{ shrink: true }}
+    value={String(value)}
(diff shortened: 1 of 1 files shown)


JosephMarinier:
Did you mean to make a PR against `nandhini/config_ui_string_editable` instead of `main`?

webapp/src/pages/Settings.tsx
```diff
@@ -531,6 +532,15 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
               </React.Fragment>
+              <React.Fragment key="persistent_id">
+                <Typography variant="body2">persistent_id:</Typography>
+                {displayStringField(
+                  "persistent_id",
+                  resultingConfig.columns.persistent_id,
+                  "columns",
+                  ""
+                )}
+              </React.Fragment>
```
JosephMarinier, line 543: The fragment is not necessary, you can remove it
```suggestion
              <Typography variant="body2">persistent_id:</Typography>
              {displayStringField(
                "persistent_id",
                resultingConfig.columns.persistent_id,
                "columns",
                ""
              )}
```

webapp/src/pages/Settings.tsx
```diff
@@ -783,6 +793,69 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+            className="fixedWidthInput"
+            value={String(resultingConfig.use_cuda)}
+            disabled={isUpdatingConfig}
+            onChange={(newValue: string | boolean) =>
+              newValue &&
+              setPartialConfig({
+                ...partialConfig,
+                use_cuda: newValue === "auto" ? newValue : Boolean(newValue),
+              })
+            }
```
JosephMarinier: There are two mistakes here that make it impossible to set `false`. Here is a suggestion that works:
```suggestion
            onChange={(newValue) => {
              const use_cuda =
                newValue === "true" || (newValue === "auto" && "auto");
              setPartialConfig({ ...partialConfig, use_cuda });
            }}
```

webapp/src/pages/Settings.tsx
```diff
@@ -783,6 +793,69 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+                onChange={(...[, checked]) =>
+                  setPartialConfig({
+                    ...partialConfig,
+                    large_dask_cluster: checked,
+                  })
+                }
+              />
+            }
+            label="large_dask_cluster"
+            labelPlacement="start"
```
JosephMarinier, line 852: I think we should leave the label and the `"end"`, to match the metrics.
```suggestion
```

webapp/src/pages/Settings.tsx
```diff
@@ -783,6 +793,69 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+              <MenuItem key={item} value={item}>
+                {item}
+              </MenuItem>
+            ))}
+          </StringField>
+          <FormControlLabel
+            control={
+              <Checkbox
+                size="small"
+                checked={Boolean(resultingConfig.large_dask_cluster)}
```
JosephMarinier, line 841: `resultingConfig.large_dask_cluster` is already a boolean, no need to convert it
```suggestion
                checked={resultingConfig.large_dask_cluster}
```

webapp/src/pages/Settings.tsx
```diff
@@ -881,6 +954,13 @@ const Settings: React.FC<Props> = ({ onClose }) => {
         >
           {getModelContractConfigSection()}
         </AccordionLayout>
+        <AccordionLayout
+          name="Common Fields Configuration"
+          description="View and edit generic fields adpated based on the user's machine."
```
JosephMarinier, line 959: I guess you meant "adapted" instead of "adpated", but even then the sentence is grammatically incorrect.

webapp/src/pages/Settings.tsx
```diff
@@ -783,6 +793,69 @@ const Settings: React.FC<Props> = ({ onClose }) => {
     </FormGroup>
   );
 
+  const getCommonFieldsConfigSection = () => (
+    <Box marginY={1.5}>
```
JosephMarinier, line 797: I need another option. @gabegma, do you think we should add a "General" sub-title to the new section even though there are no other sub-sections? That way it would persist the same "structure" as the other sections.
image

webapp/src/pages/Settings.tsx
```diff
@@ -783,6 +793,69 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
+      <FormGroup>
+        <Columns columns={4}>
+          {displayReadonlyFields(
+            "artifact_path",
+            resultingConfig.artifact_path
+          )}
+          <NumberField
+            key="batch_size"
```
JosephMarinier, line 805: This `key` is not necessary.
```suggestion
```

webapp/src/pages/Settings.tsx
```diff
@@ -783,6 +793,69 @@ const Settings: React.FC<Props> = ({ onClose }) => {
     </FormGroup>
   );
 
+  const getCommonFieldsConfigSection = () => (
+    <Box marginY={1.5}>
+      <FormGroup>
+        <Columns columns={4}>
+          {displayReadonlyFields(
```
JosephMarinier, line 800: We could take this opportunity to remove the `s`.

webapp/src/pages/Settings.tsx
gabegma: Oh interesting! I actually don't think it looks so odd like that. I'm not sure we need to add "General". What do you think?

JosephMarinier: I good either way. Let's keep it like that.
The only thing I'd do is adjust the spacing, to give the result shown in my previous screenshot.
```suggestion
    <Box marginTop={2}>
```
It doesn't look as sharp (not centered) as it is now:
image

webapp/src/pages/Settings.tsx
nandhinibsn: Sorry, I am not sure I get it. Do you mean to say, "remove the label and let the label placement be the default"?

webapp/src/pages/Settings.tsx
nandhinibsn: My bad! I will use Grammarly to fix the grammar here.

webapp/src/pages/Settings.tsx
JosephMarinier: Why did you leave the `newValue &&`? Won't it always be truthy?

webapp/src/pages/Settings.tsx
```diff
@@ -254,6 +254,7 @@ const Settings: React.FC<Props> = ({ onClose }) => {
     Partial<AzimuthConfig>
   >({});
   const isEmptyPartialConfig = Object.keys(partialConfig).length === 0;
+  const isNull = (value: string) => (value === "" ? null : value);
```
JosephMarinier, line 257: That `(value === "" ? null : value)` can be simplified to `value || null`. I don't think we need a function for that.

webapp/src/pages/Settings.tsx
```diff
@@ -374,20 +375,20 @@ const Settings: React.FC<Props> = ({ onClose }) => {
...
               [config]: {
                 ...resultingConfig[config],
-                [field]: newValue,
+                [field]: isNull(newValue),
               },
             })
           : setPartialConfig({
               ...partialConfig,
-              [field]: newValue,
+              [field]: isNull(newValue),
```
JosephMarinier, line 391: Watch out! Most fields can't be `null`. Only `rejection_class`, `saliency_layer`, and `remote` can be `null`.

webapp/src/pages/Settings.tsx
```diff
@@ -374,20 +375,20 @@ const Settings: React.FC<Props> = ({ onClose }) => {
     <StringField
       key={field}
       label={label}
-      value={String(value)}
+      value={value === null ? "" : String(value)}
```
JosephMarinier, line 378: As I said earlier, that's as simple as
```suggestion
      value={value ?? ""}
```

webapp/src/pages/Settings.tsx
nandhinibsn: Thank you :) I simplified it as suggested.

JosephMarinier (approved):

PR #506, opened 2023-03-21
Save config only when different from last config in history

gabegma:
## Description:
* What do you think of this implementation? Should I add an integration test? I'm not sure how to do it.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/app.py
+++ b/azimuth/app.py
@@ -367,7 +367,8 @@ def run_startup_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
     if _dataset_split_managers.get(DatasetSplitName.eval):
         run_validation(DatasetSplitName.eval, task_manager, azimuth_config)
 
-    save_config(azimuth_config)  # Save only after the validation modules ran successfully
+    if not azimuth_config.is_last_in_config_history():
+        save_config(azimuth_config)  # Save only after the validation modules ran successfully
 
     global _startup_tasks, _ready_flag
     _startup_tasks = startup_tasks(_dataset_split_managers, task_manager)
--- a/azimuth/app.py
+++ b/azimuth/app.py
@@ -33,7 +33,7 @@ from azimuth.types import DatasetSplitName, ModuleOptions
 from azimuth.utils.cluster import default_cluster
 from azimuth.utils.conversion import JSONResponseIgnoreNan
 from azimuth.utils.logs import set_logger_config
-from azimuth.utils.project import load_dataset_split_managers_from_config, save_config
+from azimuth.utils.project import load_dataset_split_managers_from_config
 from azimuth.utils.validation import assert_not_none
 
 _dataset_split_managers: Dict[DatasetSplitName, Optional[DatasetSplitManager]] = {}
@@ -368,7 +368,7 @@ def run_startup_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
         run_validation(DatasetSplitName.eval, task_manager, azimuth_config)
 
     if not azimuth_config.is_last_in_config_history():
-        save_config(azimuth_config)  # Save only after the validation modules ran successfully
+        azimuth_config.save()  # Save only after the validation modules ran successfully
 
     global _startup_tasks, _ready_flag
     _startup_tasks = startup_tasks(_dataset_split_managers, task_manager)
--- a/azimuth/app.py
+++ b/azimuth/app.py
@@ -367,8 +367,7 @@ def run_startup_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
     if _dataset_split_managers.get(DatasetSplitName.eval):
         run_validation(DatasetSplitName.eval, task_manager, azimuth_config)
 
-    if not azimuth_config.is_last_in_config_history():
-        azimuth_config.save()  # Save only after the validation modules ran successfully
+    azimuth_config.save()  # Save only after the validation modules ran successfully
 
     global _startup_tasks, _ready_flag
     _startup_tasks = startup_tasks(_dataset_split_managers, task_manager)
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -450,16 +450,28 @@ class AzimuthConfig(
 
         if load_config_history:
             config_history_path = cfg.get_config_history_path()
-            try:
-                with jsonlines.open(config_history_path, mode="r") as config_history:
-                    *_, last_config = config_history
-            except (FileNotFoundError, ValueError):
-                log.info("Empty or invalid config history.")
-            else:
+            last_config = cls.load_last_config_from_history(config_history_path)
+            if last_config:
                 log.info(f"Loading latest config from {config_history_path}.")
-                return AzimuthConfigHistory.parse_obj(last_config).config
+                return last_config
+            else:
+                log.info("Empty or invalid config history.")
 
         return cls.parse_file(config_path) if config_path else cls()
+
+    @classmethod
+    def load_last_config_from_history(cls, config_history_path: str) -> Optional["AzimuthConfig"]:
+        try:
+            with jsonlines.open(config_history_path, mode="r") as config_history:
+                *_, last_config = config_history
+        except (FileNotFoundError, ValueError):
+            return None
+        else:
+            return AzimuthConfigHistory.parse_obj(last_config).config
+
+    def is_last_in_config_history(self) -> bool:
+        last_config = self.load_last_config_from_history(self.get_config_history_path())
+        return self == last_config
 
     def log_info(self):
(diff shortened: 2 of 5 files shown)


tests/test_config.py
```diff
@@ -260,7 +260,7 @@ def test_update_config(tiny_text_config, monkeypatch, dask_client):
 
 def test_load_from_config_history(tiny_text_config):
     # With no config history, the loaded config is the default, not the tiny_text_config.
-    cfg = load_azimuth_config(config_path=None, load_config_history=True)
+    cfg = load_azimuth_config(config_path=None, load_config_history=False)
```
JosephMarinier, line 263: Why change that? This is `test_load_from_config_history` and I was testing that when the config history is empty, it defaults to the default config.

tests/test_config.py
```diff
@@ -269,3 +269,4 @@ def test_load_from_config_history(tiny_text_config):
     cfg = load_azimuth_config(config_path=None, load_config_history=True)
     assert cfg == tiny_text_config
     del os.environ["ARTIFACT_PATH"]
+    assert cfg.is_last_in_config_history()
```
JosephMarinier, line 272: > Should I add an integration test?

I think that's good enough. It wouldn't be critical if that breaks and we get duplicated configs.

One easy test would be in `test_routers.test_config.test_update_config`, making an empty `client.patch("/config", json={})` and `assert len(loaded_configs) == initial_config_count` so there is no duplicate.

azimuth/app.py
```diff
@@ -367,7 +367,8 @@ def run_startup_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
     if _dataset_split_managers.get(DatasetSplitName.eval):
         run_validation(DatasetSplitName.eval, task_manager, azimuth_config)
 
-    save_config(azimuth_config)  # Save only after the validation modules ran successfully
+    if not azimuth_config.is_last_in_config_history():
```
JosephMarinier, line 370: Have you considered adding the `if` inside `save_config`? In fact, `save_config` could directly do
```py
if azimuth_config == azimuth_config.load_last_config_from_history(self.get_config_history_path()):
    return
```
and we wouldn't need `is_last_in_config_history()` anymore.

azimuth/config.py
```diff
@@ -450,17 +450,29 @@ def load(cls, config_path: Optional[str], load_config_history: bool) -> "Azimuth
 
         if load_config_history:
             config_history_path = cfg.get_config_history_path()
-            try:
-                with jsonlines.open(config_history_path, mode="r") as config_history:
-                    *_, last_config = config_history
-            except (FileNotFoundError, ValueError):
-                log.info("Empty or invalid config history.")
-            else:
+            last_config = cls.load_last_config_from_history(config_history_path)
```
JosephMarinier, line 453: > What do you think of this implementation?

I really like the approach, plus I like this refactor. 👍

tests/test_config.py
gabegma: Right, I like it!

azimuth/app.py
gabegma: I'm ok with it! I was debating if it could be considered doing 2 things too, I'm ok either way.

And I 💯 % agree with switching to a method! I was going to propose it but then I thought I was maybe stretching the scope of this PR. It wouldn't be a class method right? So `azimuth_config.save()`. 

Last question: in your suggestion, you `return `nothing if they are equal. Wouldn't it be cleaner to check if they are different and then continue with the save function? Avoiding an empty return?

JosephMarinier: Right, we could argue either way. To use a fancy word, I'd say `save()` does one idempotent action, meaning if you save multiple times with the same input, you get the same result as saving it only one time.

Right. I meant `AzimuthConfig.save()` as the general signature. Here, `azimuth_config` is a clear name, but it's still a variable name that doesn't "mean" anything. Technically, the method does exist on the class, and you can call it by passing `self`, for example, `AzimuthConfig.save(azimuth_config)`.

I prefer "early `return`s" over nesting in `if`s, to avoid indenting the "body" of the function. Imagine there are multiple conditions, I prefer
```py
def a():
    something = ...
    if something:
        return

    something_else = ...
    if something_else:
        return

    one_last_thing = ...
    if one_last_thing:
        return

    do_the_actual_thing()
```
over
```py
def a():
    something = ...
    if something:
        something_else = ...
        if something_else:
            one_last_thing = ...
            if one_last_thing:
                do_the_actual_thing()
```
The `a` looks more linear, or sequential. The same structure works with `raise ...` instead of `return`, so we can write more uniform code. And finally, if we add or remove a condition, we get only a small diff in Git, instead of the whole "body" of the function because of an indentation change.

There are actually some "clean code" automated tools that score your code on multiple metrics, including the "depth" of indentations, considered bad, so `a` scores best.

That said, at the end of the day, it is a preference thing. What do you think?

gabegma: Oh, I didn't realize we could pass `self` like that, got it!

Yeah, I get that, I actually finished the clean code chapter on that last week. I was not tempted to do it because it was just one indent and a few lines of code, but it can easily get bigger. Plus I didn't realize we could just have empty returns like that. I'll make your suggested change!

azimuth/config.py
```diff
@@ -450,17 +450,25 @@ def load(cls, config_path: Optional[str], load_config_history: bool) -> "Azimuth
...
-            except (FileNotFoundError, ValueError):
-                log.info("Empty or invalid config history.")
-            else:
+            last_config = cls.load_last_from_config_history(config_history_path)
+            if last_config:
                 log.info(f"Loading latest config from {config_history_path}.")
-                return AzimuthConfigHistory.parse_obj(last_config).config
+                return last_config
+            else:
+                log.info("Empty or invalid config history.")
```
JosephMarinier, line 458: Now that this is a `if`, and no longer a `try`, we can remove the `else`:
```suggestion

            log.info("Empty or invalid config history.")
```

gabegma: Right, because we return in the `if`.

tests/test_routers/test_config.py
```diff
@@ -262,6 +262,11 @@ def test_update_config(app: FastAPI, wait_for_startup_after):
     get_config = client.get("/config").json()
     assert not get_config["pipelines"]
 
+    # Empty update
+    res = client.patch("/config", json={})
+    assert res.status_code == 200
+    assert get_config == client.get("/config").json()
```
JosephMarinier, line 268: We already have `res` from the `patch`, with the current config.
```suggestion
    assert get_config == res.json()
```
Also, didn't we want to test that the config history didn't get longer?
```py
    with jsonlines.open(jsonl_file_path, "r") as reader:
        loaded_configs = list(reader)
    assert len(loaded_configs) == initial_config_count
```

JosephMarinier (approved):
Love it! 👍 Thank you!

tests/test_routers/test_config.py
gabegma: We have done the same throughout this test, to test that the config that we `get` after the `patch` is right (even though the patch returns it). 

I think that since we are testing once at the end of the test that the `config_history` length is right, we don't need to test it for just this case?

JosephMarinier: > testing once at the end of the test

won't tell us which action increased the config history by one. This patch could start adding to the config history, and the following action could stop, at the same time. I agree it's unlikely, but tests are there for that.

gabegma: Ok, but at this point we were already testing 3 changes, and making sure at the end that only once did it save a new line - would you rather I change the logic so we test the length after each change? Or just for this one? 

Also, it will already be `initial_config_count + 1`, unless I move this piece of code to the beginning of the code. Would you rather I do that?

gabegma: Ok, let me know what you think of my last commit!

tests/test_routers/test_config.py
```diff
@@ -262,9 +265,14 @@ def test_update_config(app: FastAPI, wait_for_startup_after):
...
+
     with jsonlines.open(jsonl_file_path, "r") as reader:
         loaded_configs = list(reader)
```
JosephMarinier, line 274: I think we are missing updating the `new_config_count`, otherwise the following `assert` doesn't test anything new.
```suggestion
        new_config_count = len(list(reader))
```

gabegma: The `new_config_count` is the variable that I set after the first patch (so equal to `initial_config_count + 1`). So here I'm checking that the length of the `loaded_configs` is still equal to `new_config_count`. Also, I need the `loaded_configs` variable for further down in the test.

Am I missing something?

JosephMarinier: Oh I see. I was confused by the two `with ... as reader` being different, I thought it was a mistake. Sorry. That's good. 👍

PR #518, opened 2023-03-26
Tackle dask issues

gabegma:
## Description:
We are now often seeing `distributed.comm.core.CommClosedError: in <TCP (closed) ConnectionPool.broadcast local=tcp://127.0.0.1:62774 remote=tcp://127.0.0.1:62755>: Stream is closed` errors in the logs, which causes tasks to be lost. They are often trigger when `self.client.run(ArtifactManager.clear_cache)` is called.
* This PR does not fix the issue. I believe that the problem happens as often. However, it enables Dask Dashboard which can help troubleshoot. 
* I believe that the only thing that made the issue appear less often was updating `dask`/`distributed`; however, I can't get the router tests to work with it. 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/artifact_manager.py
+++ b/azimuth/modules/base_classes/artifact_manager.py
@@ -1,6 +1,7 @@
 # Copyright ServiceNow, Inc. 2021 – 2022
 # This source code is licensed under the Apache 2.0 license found in the LICENSE file
 # in the root directory of this source tree.
+from multiprocessing import Lock
 from typing import Callable, Dict, Optional
 
 from datasets import DatasetDict
@@ -38,9 +39,10 @@ class ArtifactManager:
 
     @classmethod
     def get_instance(cls):
-        if cls.instance is None:
-            cls.instance = cls()
-        return cls.instance
+        with Lock():
+            if cls.instance is None:
+                cls.instance = cls()
+            return cls.instance
 
     def get_dataset_split_manager(
         self, config: AzimuthConfig, name: DatasetSplitName
@@ -124,4 +126,5 @@ class ArtifactManager:
 
     @classmethod
     def clear_cache(cls) -> None:
-        cls.instance = None
+        with Lock():
+            cls.instance = None
--- a/azimuth/modules/base_classes/dask_module.py
+++ b/azimuth/modules/base_classes/dask_module.py
@@ -6,9 +6,10 @@ import abc
 import os
 import threading
 import time
+import uuid
 from functools import partial
 from os.path import join as pjoin
-from typing import Any, Callable, Dict, Generic, List, Optional, Tuple, TypeVar, cast
+from typing import Any, Callable, Dict, Generic, List, Optional, TypeVar, cast
 
 import structlog
 from datasets import Dataset
@@ -78,9 +79,8 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
         raise NotImplementedError
 
     @property
-    def task_id(self) -> Tuple[str, int]:
-        indices_hash = hash(tuple(self.get_caching_indices()))
-        return self.name, indices_hash
+    def task_id(self) -> str:
+        return "_".join(map(str, (self.name, hash(tuple(self.get_caching_indices())))))
 
     def _get_config_scope(self, config) -> ConfigScope:
         """Get the current config scope from full/partial config."""
@@ -111,12 +111,13 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
         deps = [d.done_event for d in dependencies] if dependencies is not None else []
         if not all(deps):
             raise ValueError("Can't wait for an unstarted Module.")
-        self.done_event = Event(name="-".join(map(str, self.task_id)), client=client)
+        self.done_event = Event(name=self.task_id, client=client)
         # pure=false to be sure that everything is rerun.
         self.future = client.submit(
             self._compute_on_dataset_split_with_deps,
             pure=False,
             dependencies=deps,
+            key=self.task_id + "_" + str(uuid.uuid4()),  # Unique identifier
         )
         # Tell that this future is used on which indices.
         self.future.indices = self.get_caching_indices()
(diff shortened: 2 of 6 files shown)


azimuth/modules/base_classes/dask_module.py
```diff
@@ -111,12 +111,13 @@ def start_task_on_dataset_split(
...
         if not all(deps):
             raise ValueError("Can't wait for an unstarted Module.")
-        self.done_event = Event(name="-".join(map(str, self.task_id)), client=client)
+        self.done_event = Event(name=self.task_id, client=client)
         # pure=false to be sure that everything is rerun.
         self.future = client.submit(
             self._compute_on_dataset_split_with_deps,
             pure=False,
             dependencies=deps,
+            key=self.task_id + "_" + str(uuid.uuid4()),  # Unique identifier
```
gabegma, line 120: This allows seeing which task is which in the Dask dashboard. I used the same method as Dask to produce a unique id, but I prepended it with the `task_id` instead of the function name.

azimuth/modules/base_classes/dask_module.py
```diff
@@ -111,12 +111,13 @@ def start_task_on_dataset_split(
         deps = [d.done_event for d in dependencies] if dependencies is not None else []
         if not all(deps):
             raise ValueError("Can't wait for an unstarted Module.")
-        self.done_event = Event(name="-".join(map(str, self.task_id)), client=client)
+        self.done_event = Event(name=self.task_id, client=client)
```
gabegma, line 114: For some reason, here it was a "-" and not a "_". Not sure why and if it was causing any issues.

azimuth/modules/base_classes/artifact_manager.py
```diff
@@ -38,9 +39,10 @@ def __init__(self):
 
     @classmethod
     def get_instance(cls):
-        if cls.instance is None:
-            cls.instance = cls()
-        return cls.instance
+        with Lock():
```
gabegma, line 42: I'm not sure I have seen an improvement when adding this, but I guess it can't hurt?

azimuth/modules/base_classes/dask_module.py
```diff
@@ -129,6 +130,10 @@ def start_task_on_dataset_split(
         th.start()
         return self
 
+    def custom_query_task_id(self, custom_query):
```
gabegma, line 133: Does this need to be a property?

gabegma:
@JosephMarinier @Dref360 I was unsuccessful at solving the issue. LMK what you think of this PR, I think we should at least merge the Dask Dashboard support.

azimuth/modules/base_classes/dask_module.py
JosephMarinier: I don't think it can be a property, since it takes an argument (`custom_query`).

azimuth/modules/base_classes/artifact_manager.py
JosephMarinier: Right

azimuth/modules/base_classes/dask_module.py
gabegma: Right - that was dumb, sorry! :D I still don't have a good understanding for properties.

azimuth/task_manager.py
```diff
@@ -197,7 +197,7 @@ def get_custom_task(
                 mod_options=mod_options,
             )
             # Check if this task already exist.
-            key = "_".join(map(str, task.task_id + (hash(str(custom_query)),)))
+            key = task.custom_query_task_id(custom_query)
             if key in self.current_tasks:
```
JosephMarinier, line 201: I know it's not new, but we should be able to simplify the next 4 lines, if you want:
```diff
-            if key in self.current_tasks:
-                task = self.current_tasks[key]
-            else:
-                self.current_tasks[key] = task
+            task = self.current_tasks.setdefault(key, task)
```

azimuth/modules/base_classes/dask_module.py
```diff
@@ -78,9 +79,8 @@ def get_caching_indices(self) -> List[int]:
         raise NotImplementedError
 
     @property
-    def task_id(self) -> Tuple[str, int]:
-        indices_hash = hash(tuple(self.get_caching_indices()))
-        return self.name, indices_hash
+    def task_id(self) -> str:
+        return "_".join(map(str, (self.name, hash(tuple(self.get_caching_indices())))))
```
JosephMarinier, line 83: Any reason for keeping the `map()` here? I think it's rather confusing.
```suggestion
        return f"{self.name}_{hash(tuple(self.get_caching_indices()))}"
```

gabegma: Right, I agree. I had something ish similar but `mypy` was complaining. 
`return self.name+"_"+str(hash(tuple(self.get_caching_indices())))}`.

But the f-string is much cleaner and `mypy` shouldn't complain. Thanks!

azimuth/modules/base_classes/dask_module.py
JosephMarinier: While we're here.
```suggestion
            key=f"{self.task_id}_{uuid.uuid4()}",  # Unique identifier
```

azimuth/modules/base_classes/dask_module.py
JosephMarinier: Right! Since f-strings implicitly calls `str()` on things, so you don't have to do it yourself.

azimuth/modules/base_classes/dask_module.py
gabegma: Not sure how I forgot about f-strings :/ Good catch.

azimuth/modules/base_classes/dask_module.py
gabegma: Oh, right!!

azimuth/modules/base_classes/dask_module.py
JosephMarinier: Haha! No worries!

JosephMarinier (approved):

PR #523, opened 2023-03-28
Clean up config

JosephMarinier:
## Description:

This doesn't affect the app, it simply cleans up the code, making it shorter, while providing more robust typing for fields that can be `null` and fields with literal strings (select menus). Please see the commit descriptions.

I reduced the code by almost 100 lines! Splitting into different files added back 20 lines of imports and overhead (which is fine, those lines are trivial), and the extra typing added a bit more (but added great type security too), hence why the net reduction is 56 lines.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -598,33 +598,6 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     />
   );
 
-  const displayPipelineStringField = (
-    pipelineIndex: number,
-    pipeline: PipelineDefinition,
-    field: string,
-    value: string | null,
-    subPipeline?: keyof PipelineDefinition,
-    postprocessorIdx?: number
-  ) => (
-    <StringField
-      key={field}
-      label={field}
-      value={String(value)}
-      disabled={isUpdatingConfig}
-      onChange={(newValue) => {
-        if (subPipeline === "model") {
-          updateModel(pipelineIndex, { [field]: newValue });
-        } else if (subPipeline === "postprocessors") {
-          updatePostprocessor(pipelineIndex, postprocessorIdx!, {
-            [field]: newValue,
-          });
-        } else {
-          updatePipeline(pipelineIndex, { [field]: newValue });
-        }
-      }}
-    />
-  );
-
   const displayPostprocessorNumberField = (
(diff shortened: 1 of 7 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -914,7 +914,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
               key={field}
               label={field}
               value={value}
-              disabled={!resultingConfig[config] || isUpdatingConfig}
+              disabled={resultingConfig[config] === null || isUpdatingConfig}
```
christyler3030, line 728: Could be extracted for reuse instead of checking every time.

webapp/src/pages/Settings.tsx
```diff
@@ -135,17 +135,25 @@ const FIELD_COMMON_PROPS = {
   InputLabelProps: { shrink: true },
 } as const;
 
-const StringField: React.FC<
-  Omit<TextFieldProps, "onChange"> & FieldProps<string>
-> = ({ onChange, ...props }) => (
+const StringField = <T extends string | null>({
+  value,
```
christyler3030: Out of curiosity: why is this explicit now?

christyler3030 (approved):

webapp/src/pages/Settings.tsx
JosephMarinier: To support `null` as a valid value, I now do `value={value ?? ""}`. Without that, `value={null}` would make the component "uncontrolled", and changing the `TextField` from controlled to uncontrolled (and vice versa) results in errors in the console. The `value={value ?? ""}` treats `null` like an empty string (which is a valid value for a controlled component), which makes the `placeholder: "null"` appear in pale gray.
You might notice that it is therefore impossible to set empty string to a nullable fields. We think this is good enough for now, as none of the nullable fields support the empty string as a valid value.
Does that make sense?

webapp/src/pages/Settings.tsx
JosephMarinier: Right. I hesitated. The pattern is repeated 9 times, but it's with 3 different things on the right, so it's not trivial. I have some ideas to go further in removing duplication, so I'll keep that in mind for future refactors.

PR #524, opened 2023-03-30
Show error messages from backend on API failure

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
+++ b/webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
@@ -139,7 +139,7 @@ describe("PerturbationTestingPreview with Failure response", () => {
       // expected error message
       expect(
         screen.getByText(
-          "Something went wrong fetching behavioral testing summary"
+          /Something went wrong fetching behavioral testing summary/i
         )
       ).toBeVisible();
     });
--- a/webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
+++ b/webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
@@ -50,20 +50,20 @@ describe("PerturbationTestingPreview", () => {
     renderPerturbationTestingPreview({ train: true, eval: true });
     await waitFor(() => {
       // Verify if the actual response data from API is visibly displayed on the toggle
-      expect(screen.getByText(/14.0%/i)).toBeVisible();
+      expect(screen.getByText("14.0%")).toBeVisible();
       expect(screen.getByText("Failure rate - Evaluation Set")).toBeVisible();
-      expect(screen.getByText(/16.0%/i)).toBeVisible();
+      expect(screen.getByText("16.0%")).toBeVisible();
       expect(screen.getByText("Failure rate - Training Set")).toBeVisible();
 
       // styling should be applied for the selected toggle as expected
-      expect(screen.getByText(/14.0%/)).toHaveStyle(
+      expect(screen.getByText("14.0%")).toHaveStyle(
         "font-size: 3.75rem; color: rgb(237, 108, 2)"
       );
 
       // verify with some random unexpected values to not have been displayed.
-      expect(screen.queryByText(/17.0%/i)).toBeNull();
+      expect(screen.queryByText("17.0%")).toBeNull();
       expect(screen.queryByText("some random text")).toBeNull();
-      expect(screen.queryByText(/26.0%/i)).toBeNull();
+      expect(screen.queryByText("26.0%")).toBeNull();
     });
   });
 
@@ -139,7 +139,7 @@ describe("PerturbationTestingPreview with Failure response", () => {
       // expected error message
       expect(
         screen.getByText(
-          /Something went wrong fetching behavioral testing summary/i
+          /Something went wrong fetching behavioral testing summary/
         )
       ).toBeVisible();
     });
--- a/webapp/src/services/api.ts
+++ b/webapp/src/services/api.ts
@@ -1,6 +1,6 @@
 import { QueryReturnValue } from "@reduxjs/toolkit/dist/query/baseQueryTypes";
 import { createApi, fakeBaseQuery } from "@reduxjs/toolkit/query/react";
-import { AzimuthConfig, UtterancePatch } from "types/api";
+import { AzimuthConfig, HTTPExceptionModel, UtterancePatch } from "types/api";
 import {
   fetchApi,
   GetUtterancesQueryState,
@@ -18,8 +18,12 @@ const responseToData =
     try {
       const response = await responsePromise(arg);
       return { data: await response.json() };
-    } catch {
-      return { error: { message } };
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        return { error: { message: message + "\n" + err.message } };
+      } else {
+        return { error: { message } };
+      }
     }
   };
 
--- a/webapp/src/services/api.ts
+++ b/webapp/src/services/api.ts
@@ -1,6 +1,6 @@
 import { QueryReturnValue } from "@reduxjs/toolkit/dist/query/baseQueryTypes";
 import { createApi, fakeBaseQuery } from "@reduxjs/toolkit/query/react";
-import { AzimuthConfig, HTTPExceptionModel, UtterancePatch } from "types/api";
+import { AzimuthConfig, UtterancePatch } from "types/api";
 import {
   fetchApi,
   GetUtterancesQueryState,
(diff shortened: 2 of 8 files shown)


christyler3030 (commented):
I think this makes sense although I am still unsure of the design for this. The toasts generated by this are not easy to read. 

Also we need to decide what to do with the general messages such as the one found here:
https://github.com/ServiceNow/azimuth/pull/524/files#diff-eab2d4424b736fc0a65af85697bd0abf297013425f52e684f71463ddcf22e16dL60

webapp/src/services/api.ts
```diff
@@ -18,8 +18,12 @@ const responseToData =
     try {
       const response = await responsePromise(arg);
       return { data: await response.json() };
-    } catch {
-      return { error: { message } };
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        return { error: { message: message + "\n" + err.message } };
```
JosephMarinier, line 23: The `\n` has no effect unless you add a `white-space: pre-wrap` to the toast (which I think we should).

webapp/src/services/api.ts
```diff
@@ -18,8 +18,12 @@ const responseToData =
     try {
       const response = await responsePromise(arg);
       return { data: await response.json() };
-    } catch {
-      return { error: { message } };
+    } catch (err: unknown) {
+      if (err instanceof Error) {
```
JosephMarinier, line 22: Is `err instanceof Error` ever false? Do you have an example?

webapp/src/utils/api.ts
```diff
@@ -88,7 +89,9 @@ export const fetchApi =
     // fetch() might throw if the user is offline, or some unlikely networking error occurs, such a DNS lookup failure.
     // Let's also throw if the status is not OK, so it's uniform.
     if (!response.ok) {
-      throw Error(`${response.status} : ${response.statusText}`);
+      const { detail: errorMessage } =
+        (await response.json()) as HTTPExceptionModel;
```
JosephMarinier, line 93: The `response` might not be `JSON`. For example, if the back end is unreachable, we get plain text `Error occured while trying to proxy: localhost:3000/status`. That results in an unhelpful error: `Unexpected token 'E', "Error occu"... is not valid JSON`.

webapp/src/services/api.ts
```diff
@@ -18,8 +18,12 @@ const responseToData =
     try {
       const response = await responsePromise(arg);
       return { data: await response.json() };
-    } catch {
-      return { error: { message } };
+    } catch (err: unknown) {
```
JosephMarinier, line 21: This `err` is already `unknown`, no need to specify it.
```suggestion
    } catch (err) {
```

webapp/src/services/api.ts
JosephMarinier: Done in b13a570e80cd486af6cb5d17aec3acf7985e841e

webapp/src/services/api.ts
```diff
@@ -1,6 +1,6 @@
 import { QueryReturnValue } from "@reduxjs/toolkit/dist/query/baseQueryTypes";
 import { createApi, fakeBaseQuery } from "@reduxjs/toolkit/query/react";
-import { AzimuthConfig, UtterancePatch } from "types/api";
+import { AzimuthConfig, HTTPExceptionModel, UtterancePatch } from "types/api";
```
JosephMarinier, line 3: Don't forget to check the warnings in the terminal. This new import is not used:
```suggestion
import { AzimuthConfig, UtterancePatch } from "types/api";
```

webapp/src/services/api.ts
```diff
@@ -18,12 +18,9 @@ const responseToData =
...
       return { data: await response.json() };
-    } catch (err: unknown) {
-      if (err instanceof Error) {
-        return { error: { message: message + "\n" + err.message } };
-      } else {
-        return { error: { message } };
-      }
+    } catch (err) {
+      const error = err as Error;
+      return { error: { message: message + error.message } };
```
JosephMarinier: I think it was better with the `\n`.

webapp/src/utils/api.ts
```diff
@@ -88,7 +89,18 @@ export const fetchApi =
...
     if (!response.ok) {
-      throw Error(`${response.status} : ${response.statusText}`);
+      const httpExceptionModelCodes: number[] = [
+        400, 401, 403, 404, 422, 500, 503,
+      ];
+      if (httpExceptionModelCodes.includes(response.status)) {
+        const { detail: errorMessage } =
+          (await response.json()) as HTTPExceptionModel;
+        throw Error(errorMessage);
+      } else {
```
JosephMarinier, line 99: Instead of assuming which status code comes with JSON or not, have you considered simply `try`ing to call `.json()` and default to the generic message if you `catch` and exception?
```suggestion
      try {
        const { detail: errorMessage } =
          (await response.json()) as HTTPExceptionModel;
        throw Error(errorMessage);
      } catch {
```

webapp/src/services/api.ts
```diff
@@ -18,8 +18,9 @@ const responseToData =
     try {
       const response = await responsePromise(arg);
       return { data: await response.json() };
-    } catch {
-      return { error: { message } };
+    } catch (err) {
+      const error = err as Error;
+      return { error: { message: message + error.message } };
```
JosephMarinier, line 23: Also, the implicit conversion of ar error to a string is the same as accessing the `.message` property, so no need to assert the type and everything here:
```suggestion
    } catch (error) {
      return { error: { message: `${message}\n${error}` } };
```

webapp/src/utils/api.ts
```diff
@@ -1,5 +1,6 @@
 import { paths } from "types/generated/generatedTypes";
 import { constructApiSearchString } from "./helpers";
+import { HTTPExceptionModel } from "types/api";
```
JosephMarinier, line 3: If you let VSCode add the import for `HTTPExceptionModel` in alphabetical order, it should place it at the top of the file.
```suggestion
import { HTTPExceptionModel } from "types/api";
import { paths } from "types/generated/generatedTypes";
import { constructApiSearchString } from "./helpers";
```

webapp/src/utils/api.ts
```diff
@@ -88,7 +89,15 @@ export const fetchApi =
...
     if (!response.ok) {
-      throw Error(`${response.status} : ${response.statusText}`);
+      try {
+        const { detail: errorMessage } =
+          (await response.json()) as HTTPExceptionModel;
+        throw Error(errorMessage);
+      } catch {
+        throw Error(
+          `Failed to load resource: the server responded with a status of ${response.status} + ${response.statusText}. `
+        );
```
JosephMarinier, line 99: I think this is quite verbose, and the "+" is particularly weird to me.
image
What do you think of something that simple?
```suggestion
        throw Error(`${response.status} ${response.statusText}`);
```
image

webapp/src/services/api.ts
```diff
@@ -18,8 +18,8 @@ const responseToData =
     try {
       const response = await responsePromise(arg);
       return { data: await response.json() };
-    } catch {
-      return { error: { message } };
+    } catch (error) {
+      return { error: { message: `${message}\n${error}` } };
```
JosephMarinier, line 22: Oups! There was an issue with my suggestion. It should have been tested before committing it. The implicit conversion from error to string adds an `Error: ` in front of the error message. You were right to use the `.message` attribute!
```suggestion
      return { error: { message: `${message}\n${(error as Error).message}` } };
```

webapp/src/utils/api.ts
JosephMarinier: Oups! There was an issue with my suggestion. I'm sorry! It should have been tested before committing it. We `throw` inside the `try`, so it will always `catch` our error. 😅 
We can revert to your initial implementation with `httpExceptionModelCodes`. This only thing I would change would be to make `httpExceptionModelCodes` a constant at the top of the file. I suggest `HTTP_EXCEPTION_STATUS_CODES`.

webapp/src/utils/api.ts
nandhinibsn: Okay, I can modify it as you suggest.

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
```diff
@@ -139,7 +139,7 @@ describe("PerturbationTestingPreview with Failure response", () => {
       // expected error message
       expect(
         screen.getByText(
-          "Something went wrong fetching behavioral testing summary"
+          /Something went wrong fetching behavioral testing summary/i
```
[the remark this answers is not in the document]
JosephMarinier: Done in 357d66bf6df8ca89cf801a676f1bcfced9e2179d

JosephMarinier (approved):
Super! 👍

webapp/src/components/Analysis/PerturbationTestingPreview.test.tsx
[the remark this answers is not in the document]
nandhinibsn: Okay, tests were failing so I added regex but i is not required here. Thank you for removing it.

PR #525, opened 2023-03-31
Assign tasks to workers

### LINKED ISSUE #411: Keep GPU workload to one worker locally.
**Is your feature request related to a problem? Please describe.**
I have a very small GPU (6Gb), so even with a batch size of 1, I can't run Azimuth with saliency on my computer. Both Dask workers load the model on the GPU which creates a OOM.

**Describe the solution you'd like**
When using a local Dask cluster (the default), it would be great if we assigned resources to one of them so that it handles all GPU-related tasks.

**Describe alternatives you've considered**
I could run on CPU, but that takes a while.

**Additional context**

gabegma:
Part of #521
Fixes #411 

## Description:
* The tests used to be flacky on the CI, but I after rebasing on all the changes, I'm not sure if that is still the case. It seems it's no longer the case given the amount of back and forth. Yay!!
* This allows the user to manually assign the GPU to worker 0, because this is where the pipeline and encoder will be loaded.

Here is the memory before and after the change.
6  before
7  assign workers

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/artifact_manager.py
+++ b/azimuth/modules/base_classes/artifact_manager.py
@@ -6,6 +6,7 @@ from typing import Callable, Dict
 
 import structlog
 from datasets import DatasetDict
+from sentence_transformers import SentenceTransformer
 
 from azimuth.config import AzimuthConfig
 from azimuth.dataset_split_manager import DatasetSplitManager
@@ -79,6 +80,7 @@ class ArtifactManager:
         ] = defaultdict(dict)
         self.models_mapping: Dict[Hash, Dict[int, Callable]] = defaultdict(dict)
         self.metrics = {}
+        self.encoder = None
         log.debug(f"Creating new Artifact Manager {id(self)}.")
 
     def get_dataset_split_manager(
@@ -157,6 +159,11 @@ class ArtifactManager:
         if metric_hash not in self.metrics:
             self.metrics[metric_hash] = load_custom_object(config.metrics[name], **kwargs)
         return self.metrics[metric_hash]
+
+    def get_encoder(self, encoder_name_or_path):
+        if self.encoder is None:
+            self.encoder = SentenceTransformer(encoder_name_or_path)
+        return self.encoder
 
     @classmethod
     def instance(cls):
--- a/azimuth/modules/base_classes/artifact_manager.py
+++ b/azimuth/modules/base_classes/artifact_manager.py
@@ -6,7 +6,6 @@ from typing import Callable, Dict
 
 import structlog
 from datasets import DatasetDict
-from sentence_transformers import SentenceTransformer
 
 from azimuth.config import AzimuthConfig
 from azimuth.dataset_split_manager import DatasetSplitManager
@@ -80,7 +79,6 @@ class ArtifactManager:
         ] = defaultdict(dict)
         self.models_mapping: Dict[Hash, Dict[int, Callable]] = defaultdict(dict)
         self.metrics = {}
-        self.encoder = None
         log.debug(f"Creating new Artifact Manager {id(self)}.")
 
     def get_dataset_split_manager(
@@ -159,11 +157,6 @@ class ArtifactManager:
         if metric_hash not in self.metrics:
             self.metrics[metric_hash] = load_custom_object(config.metrics[name], **kwargs)
         return self.metrics[metric_hash]
-
-    def get_encoder(self, encoder_name_or_path):
-        if self.encoder is None:
-            self.encoder = SentenceTransformer(encoder_name_or_path)
-        return self.encoder
 
     @classmethod
     def instance(cls):
--- a/azimuth/modules/base_classes/dask_module.py
+++ b/azimuth/modules/base_classes/dask_module.py
@@ -156,7 +156,11 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
         log.info(f"Starting custom query {self.name}")
         # pure=false to be sure that everything is rerun.
         self.future = client.submit(
-            self.compute, custom_query, key=self.custom_query_task_id(custom_query), pure=False
+            self.compute,
+            custom_query,
+            key=self.custom_query_task_id(custom_query),
+            pure=False,
+            workers=self.worker,
         )
         # Tell that this future is for custom use only.
         self.future.is_custom = True
--- a/azimuth/modules/base_classes/dask_module.py
+++ b/azimuth/modules/base_classes/dask_module.py
@@ -28,7 +28,7 @@ ConfigScope = TypeVar("ConfigScope", bound=CommonFieldsConfig)
 
 class Worker(IntEnum):
     model = 0
-    encoder = 1
+    encoder = 0
 
 
 class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
--- a/azimuth/modules/base_classes/dask_module.py
+++ b/azimuth/modules/base_classes/dask_module.py
@@ -37,6 +37,7 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
     """
 
     allowed_splits = {DatasetSplitName.train, DatasetSplitName.eval}
+    can_load_model: bool = False
 
     def __init__(
         self,
@@ -118,6 +119,7 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
             pure=False,
             dependencies=deps,
             key=f"{self.task_id}_{uuid.uuid4()}",  # Unique identifier
+            workers=0 if self.can_load_model else None,
         )
         # Tell that this future is used on which indices.
         self.future.indices = self.get_caching_indices()
--- a/azimuth/modules/base_classes/dask_module.py
+++ b/azimuth/modules/base_classes/dask_module.py
@@ -7,6 +7,7 @@ import os
 import threading
 import time
 import uuid
+from enum import IntEnum
 from functools import partial
 from os.path import join as pjoin
 from typing import Any, Callable, Dict, Generic, List, Optional, TypeVar, cast
@@ -25,6 +26,11 @@ log = structlog.get_logger()
 ConfigScope = TypeVar("ConfigScope", bound=CommonFieldsConfig)
 
 
+class Worker(IntEnum):
+    model = 0
+    encoder = 1
+
+
 class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
     """Abstract class that define an item of work to be computed on the cluster.
 
@@ -37,8 +43,7 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
     """
 
     allowed_splits = {DatasetSplitName.train, DatasetSplitName.eval}
-    can_load_model: bool = False
-    can_load_encoder: bool = False
+    worker: Optional[Worker] = None
 
     def __init__(
         self,
@@ -120,7 +125,7 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
             pure=False,
             dependencies=deps,
             key=f"{self.task_id}_{uuid.uuid4()}",  # Unique identifier
-            workers=0 if self.can_load_model else 1 if self.can_load_encoder else None,
+            workers=self.worker,
         )
         # Tell that this future is used on which indices.
         self.future.indices = self.get_caching_indices()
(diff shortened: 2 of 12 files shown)


azimuth/modules/base_classes/dask_module.py
```diff
@@ -94,6 +94,10 @@ def _get_config_scope(self, config) -> ConfigScope:
             scoped_config = base.__args__[0]
         return cast(ConfigScope, scoped_config.parse_obj(config.dict(by_alias=True)))
 
+    @property
+    def can_load_model(self) -> bool:
+        raise NotImplementedError
```
gabegma, line 99: I had import circular issues if I implemented the property here.

azimuth/modules/base_classes/mixin.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class ModelMixin:
+    """Mixin to add to modules that can load the model.
+
+    This Mixin exists to assign tasks that can load the model to the same worker, to avoid
+    loading the model by 2 workers simultaneously.
+    """
+
+    config: ConfigScope  # type: ignore
```
gabegma, line 35: I couldn't figure out how to solve for the `mypy` issue.

Dref360: What was the error?

gabegma: `Type variable "azimuth.modules.base_classes.dask_module.ConfigScope" is unbound`

azimuth/modules/base_classes/dask_module.py
Dref360: Right, we define ConfigScope in `dask_module.py`, but it is used in the Mixin. Maybe move it?

gabegma: Ok, I needed to move another thing around but it now works.

azimuth/modules/base_classes/mixin.py
Dref360: At this moment, it "must" be a ModelContractConfig right?

gabegma: As discussed offline, I had a `mypy` error with ModelContractConfig too. I'll revert the Mixin and use a class variable instead.

azimuth/modules/base_classes/module.py
```diff
@@ -164,6 +164,8 @@ def get_model(self):
         Raises:
             ValueError if no valid pipeline exists.
         """
+        if not self.can_load_model:
+            raise ValueError("This module cannot load the model. Set can_load_model to True.")
```
JosephMarinier, line 168: Very good idea to catch it here! 👍 
Detail, but to be rigorous, it seems a [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) is not quite appropriate here:
> Raised when an operation or function receives an argument that has the right type but an inappropriate value

I think we can use a more generic [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError):
> Raised when an error is detected that doesn’t fall in any of the other categories.

```suggestion
            raise RuntimeError("This module cannot load the model. Set can_load_model to True.")
```

azimuth/modules/base_classes/dask_module.py
```diff
@@ -37,6 +37,7 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
     """
 
     allowed_splits = {DatasetSplitName.train, DatasetSplitName.eval}
+    can_load_model: bool = False
```
JosephMarinier, line 40: Regarding the "can" in the attribute name, is it because there are some modules that might or might not actually load the model? If so, I think it would help to leave a comment on why it's `can_load_model` and not something like `loads_model`, `needs_model`, or `needs_to_load_model`.
Asked differently, is there a reason to set `can_load_model = True` and not load the model?

azimuth/modules/base_classes/dask_module.py
```diff
@@ -118,6 +119,7 @@ def start_task_on_dataset_split(
             pure=False,
             dependencies=deps,
             key=f"{self.task_id}_{uuid.uuid4()}",  # Unique identifier
+            workers=0 if self.can_load_model else None,
```
JosephMarinier: > This allows the user to manually assign the GPU to worker 0

How does one do that? Should we do something in our code?

azimuth/modules/dataset_analysis/similarity_analysis.py
```diff
@@ -45,10 +44,7 @@ def get_encoder_name_or_path(self):
         return model_name_or_path
 
     def get_encoder(self):
```
JosephMarinier, line 46: Should we raise an exception if `not self.can_load_encoder`? I understand it is not possible now since `get_encoder()` is only defined in `FAISSModule`, which has `can_load_encoder = True`, but it would be an extra safety net for if someone tries to create, for some reason, a new module that inherits `FAISSModule` and overwrites `can_load_encoder = False`.

azimuth/modules/base_classes/artifact_manager.py
```diff
@@ -158,6 +160,11 @@ def get_metric(self, config, name: str, **kwargs):
             self.metrics[metric_hash] = load_custom_object(config.metrics[name], **kwargs)
         return self.metrics[metric_hash]
 
+    def get_encoder(self, encoder_name_or_path):
```
JosephMarinier, line 163: Why
> Move the encoder to ArtifactManager

No issue, I just want to follow. 😅

azimuth/modules/base_classes/dask_module.py
```diff
@@ -119,7 +120,7 @@ def start_task_on_dataset_split(
             pure=False,
             dependencies=deps,
             key=f"{self.task_id}_{uuid.uuid4()}",  # Unique identifier
-            workers=0 if self.can_load_model else None,
+            workers=0 if self.can_load_model else 1 if self.can_load_encoder else None,
```
JosephMarinier, line 123: So, we don't want a module to have both `can_load_model = True` and `can_load_encoder = True`, right?
I have an idea to have typing enforce that. 💡 I haven't tested, though.
We create a `Worker` enum to give workers meaningful names:
```py
class Worker(IntEnum):
    model = 0
    encoder = 1
```
We let modules set on which worker they should run:
```diff
 class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
 ...
-    can_load_model: bool = False
-    can_load_encoder: bool = False
+    worker: Optional[Worker] = None
```
```diff
 class FAISSModule(IndexableModule[SimilarityConfig]):
 ...
-    can_load_encoder = True
+    worker = Worker.encoder
```
We are still able to raise an exception if it's not set properly:
```diff
     def get_model(self):
     ...
-        if not self.can_load_model:
+        if self.worker != Worker.model:
            raise RuntimeError("This module cannot load the model. Set can_load_model to True.")
```
And here, we simply pass the worker:
```suggestion
            workers=self.worker
```
What do you think?

azimuth/modules/base_classes/module.py
gabegma: Perfect, I didn’t know what to put. 

It’s worth saying that this method (`can_load_model`) is not perfect, a module can call the prediction module with `.compute()` or `.compute_on_dataset_split()`, which will load the model without calling `get_model()`, the best example of this is perturbation testing. I’m not sure how we can make that more fail-proof. Perhaps I can add a TODO.

azimuth/modules/base_classes/dask_module.py
gabegma: Right, we could rename to `loads_model`, good point.

azimuth/modules/base_classes/dask_module.py
gabegma: Good question, I had a conversation about it with @Dref360 but didn’t ask for details. Do we need to add something in the code to allow the GPU to be assigned to worker 0?

azimuth/modules/base_classes/dask_module.py
JosephMarinier: Perfect, I think I'm following, then. If https://github.com/ServiceNow/azimuth/pull/525#discussion_r1171396147 works, we can ignore this.

azimuth/modules/base_classes/dask_module.py
gabegma: Yeah, I like it!

azimuth/modules/base_classes/artifact_manager.py
JosephMarinier: > This avoids loading it twice.

Because of eval and train?

gabegma: Yes exactly, sorry.

JosephMarinier: I get it now. Thanks!

gabegma: Yeah, I guess the same question applies to the pipelines and the dataset.

azimuth/modules/base_classes/dask_module.py
Dref360: No it's more of a convention we decided on. All Modules that require GPU **must** be assigned to worker 0.

azimuth/modules/base_classes/artifact_manager.py
Dref360: Loading the encoder is not a huge cost vs loading the model, but I think this is worthwhile so that in the future we can better control the GPU memory used by Azimuth.

If your model makes 100% of the GPU mem, we are currently not able to load the sentence transformer.

gabegma: Do you mean that it is worthwhile to assign tasks that can load the encoder to a different worker than the one that loads the pipeline, or that it is worthwhile to load it once in the artifact manager? Or both?

gabegma: Oh really! But without CPU, this overloads one worker no?

Also, I think I would revert my change of moving the encoder to the ArtifactManager, here is the difference in the memory used.
7  assign workers
10  encoder not in AM

azimuth/modules/base_classes/dask_module.py
```diff
@@ -118,6 +125,7 @@ def start_task_on_dataset_split(
             pure=False,
             dependencies=deps,
             key=f"{self.task_id}_{uuid.uuid4()}",  # Unique identifier
+            workers=self.worker,
```
JosephMarinier, line 128: Do we need to add that to the other `client.submit()` too, in `DaskModule.start_task()`?

gabegma: Good point!!

tests/test_modules/test_dataset_analysis/test_class_overlap.py
```diff
@@ -20,10 +20,10 @@ def encode(self, x, *args, **kwargs):
 
 
 def test_incomplete_class_set(clinc_text_config, dask_client, monkeypatch):
-    monkeypatch.setattr(faiss_mod, "SentenceTransformer", MockedTransformer)
+    monkeypatch.setattr(artifact_manager, "SentenceTransformer", MockedTransformer)
     # Mock SentenceTransformer on all workers.
     dask_client.run(
-        lambda: monkeypatch.setattr(faiss_mod, "SentenceTransformer", MockedTransformer)
+        lambda: monkeypatch.setattr(artifact_manager, "SentenceTransformer", MockedTransformer)
```
JosephMarinier, line 26: Should we revert the changes in this test file too?

gabegma: Yeah already done

gabegma: Ah oops no! Fixing :)

gabegma:
BTW - While investigating memory usage - I’m under the impression that we have some memory leakage issues with the saliency tasks. I added issue #549.

JosephMarinier (approved):
Awesome! Good job on the graphs and all the profiling. 👍

azimuth/modules/perturbation_testing/perturbation_testing.py
```diff
@@ -58,6 +59,9 @@ class PerturbationTestingModule(DatasetResultModule[PerturbationTestingConfig]):
     """
 
     required_mod_options = {"pipeline_index"}
+    # This module doesn't call self.get_model() but requires the model (predict_task.compute(batch))
```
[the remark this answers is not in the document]
gabegma: I caught it just in time :D

PR #527, opened 2023-04-03
Refactor DatasetSplitManager to reload latest cache.

gabegma:
Part of #521

## Description:
* The only problem I see with this refactor, is that modules/functions really need to call `mod.get_dataset_split()` first (or other dm methods which call it), for the DM to check if the dataset saved in memory is up-to-date. If they call some other methods before, I foresee it could cause an issue, but perhaps not? I haven't seen any issues with the tests not when starting the app though.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/dataset_split_manager.py
+++ b/azimuth/dataset_split_manager.py
@@ -30,6 +30,7 @@ log = structlog.get_logger("DatasetSplitManager")
 
 FEATURES = "features"
 FEATURE_FAISS = "features_faiss"
+Time = float
 
 
 @dataclass(eq=True, frozen=True)  # Generates __hash__
@@ -88,21 +89,29 @@ class DatasetSplitManager:
         self._index_path = pjoin(self._base_dataset_path, "index.faiss")
         self._features_path = pjoin(self._base_dataset_path, "features.faiss.npy")
         self._file_lock = pjoin(self._hf_path, f"{name}.lock")
-        self.last_update = -1
         # Load the dataset_split from disk.
-        cached_dataset_split = self._load_dataset_split()
-        if cached_dataset_split is None:
+        self._base_dataset_split_last_update: Time = -1
+        cached_base_dataset_split = self._load_latest_base_dataset_split()
+        if cached_base_dataset_split is None:
             if dataset_split is None:
                 raise ValueError("No dataset_split cached, can't initialize.")
             log.info("Initializing tags", tags=initial_tags)
-            self._base_dataset_split, self._malformed_dataset = self._load_base_dataset_split(
+            self._base_dataset_split, self._malformed_dataset = self._process_base_dataset_split(
                 dataset_split
             )
             self._save_base_dataset_split()
         else:
-            self._base_dataset_split, self._malformed_dataset = cached_dataset_split
+            self._base_dataset_split, self._malformed_dataset = cached_base_dataset_split
         self._prediction_tables: Dict[PredictionTableKey, Dataset] = {}
+        self._prediction_tables_last_update: Dict[PredictionTableKey, Time] = defaultdict(int)
         self._validate_columns()
+
+    @property
+    def last_update(self) -> Time:
+        return max(
+            [self._base_dataset_split_last_update]
+            + list(self._prediction_tables_last_update.values())
+        )
 
     def get_dataset_split(self, table_key: Optional[PredictionTableKey] = None) -> Dataset:
         """Return a dataset_split concatenated with the config predictions.
@@ -113,6 +122,11 @@ class DatasetSplitManager:
         Returns:
             Dataset with predictions if available.
         """
+        current_last_update = self._base_dataset_split_last_update
+        newest_base_ds, last_update = self.load_latest_cache(self._save_path, current_last_update)
+        if newest_base_ds:
+            self._base_dataset_split = newest_base_ds
+            self._base_dataset_split_last_update = last_update
         if table_key is None:
             return self._base_dataset_split
         return self.dataset_split_with_predictions(table_key=table_key)
@@ -144,16 +158,16 @@ class DatasetSplitManager:
             DatasetColumn.postprocessed_prediction,
         }
 
-    def _load_dataset_split(self) -> Optional[Tuple[Dataset, Dataset]]:
+    def _load_latest_base_dataset_split(self) -> Optional[Tuple[Dataset, Dataset]]:
         if os.path.exists(self._save_path):
-            log.debug("Reloading base dataset_split.", path=self._save_path)
-            with FileLock(self._file_lock):
-                ds = self.load_cache(self._save_path)
-                malformed = self.load_cache(self._malformed_path)
-            return ds, malformed
+            current_update = self._base_dataset_split_last_update
+            base_ds, last_update = self.load_latest_cache(self._save_path, current_update)
+            malformed, _ = self.load_latest_cache(self._malformed_path, current_update)
+            self._base_dataset_split_last_update = last_update
+            return assert_not_none(base_ds), assert_not_none(malformed)
         return None
 
-    def _load_base_dataset_split(self, dataset_split) -> Tuple[Dataset, Dataset]:
+    def _process_base_dataset_split(self, dataset_split) -> Tuple[Dataset, Dataset]:
         base_dataset_split, malformed_dataset = self._split_malformed(dataset_split)
 
         # Checking if a persistent id was provided.
@@ -171,10 +185,13 @@ class DatasetSplitManager:
     def _save_base_dataset_split(self):
         # NOTE: We should not have the Index in `self.dataset_split`.
         with FileLock(self._file_lock):
-            self._base_dataset_split.save_to_disk(self._get_new_version_path(self._save_path))
-            self._malformed_dataset.save_to_disk(self._get_new_version_path(self._malformed_path))
-        self.last_update = time.time()
-        log.debug("Base dataset split saved.", path=self._save_path)
+            version_path, last_update = self._get_new_version_path(self._save_path)
+            self._base_dataset_split.save_to_disk(version_path)
+            self._malformed_dataset.save_to_disk(
+                self._get_new_version_path(self._malformed_path)[0]
+            )
+        self._base_dataset_split_last_update = last_update
+        log.debug("Base dataset split saved.", path=version_path)
 
     def get_dataset_split_with_class_names(
         self, table_key: Optional[PredictionTableKey] = None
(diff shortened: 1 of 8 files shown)


azimuth/dataset_split_manager.py
```diff
@@ -88,22 +89,30 @@ def __init__(
...
         self._prediction_tables: Dict[PredictionTableKey, Dataset] = {}
+        self._prediction_tables_last_update: Dict[PredictionTableKey, Time] = defaultdict(int)
         self._validate_columns()
 
+    @property
+    def last_update(self) -> Time:
+        return max(
+            [self._base_dataset_split_last_update]
+            + list(self._prediction_tables_last_update.values())
+        )
```
JosephMarinier, line 114: There seems to be no need for lists here:
```suggestion
        return max(self._base_dataset_split_last_update, *self._prediction_tables_last_update.values())
```

gabegma: Yeah, I was waiting for feedback on that row - I couldn't figure out the best way to do it.
With your suggestion, I get `TypeError: 'float' object is not iterable`, when `self._prediction_tables_last_update` is an empty `dict`.

JosephMarinier: Ah! That makes sense. One list (or tuple) then!
```suggestion
        return max(
            (self._base_dataset_split_last_update, *self._prediction_tables_last_update.values())
        )
```

tests/test_dataset_manager.py
```diff
@@ -306,6 +307,8 @@ def test_caching(a_text_dataset, simple_text_config):
     assert (
         len(glob(pjoin(dm1._save_path, "version_*.arrow"))) == 2
     ), "Dataset not saved on disk when added a column"
+    modified_last_update = dm1.last_update
+    assert initial_last_update < modified_last_update, "Last update should be updated."
```
JosephMarinier, line 311: Good job on asserting the fix! 👍

azimuth/dataset_split_manager.py
```diff
@@ -588,13 +611,21 @@ def load_cache(self, folder: str) -> Dataset:
...
             None,
         )
-        if cache_file:
-            log.debug(f"Loading latest cache: {cache_file.split('/')[-1]}")
-            return Dataset.load_from_disk(cache_file)
-        raise FileNotFoundError(f"No previously saved dataset in {folder}")
+        if not cache_file:
+            raise FileNotFoundError(f"No previously saved dataset in {folder}")
+
+        last_update = float(cache_file.split("_")[-1][:-6])
```
JosephMarinier: I had a couple of small ideas here and I took the liberty of committing them in a749890c4d5c02f51655d5b1167fcb286b485279. Hop you like it!

azimuth/dataset_split_manager.py
```diff
@@ -473,37 +490,38 @@ def _get_prediction_table(self, table_key: PredictionTableKey) -> Dataset:
...
             self._prediction_tables[table_key] = self._init_dataset_split(
                 empty_ds, self._prediction_tags
             ).remove_columns([DatasetColumn.row_idx])
             self.save_prediction_table(table_key)
+        else:
+            current_last_update = self._prediction_tables_last_update[table_key]
+            newest_pred_ds, last_update = self.load_latest_cache(pred_path, current_last_update)
+            if newest_pred_ds:
+                self._prediction_tables[table_key] = newest_pred_ds
+                self._prediction_tables_last_update[table_key] = last_update
```
JosephMarinier, line 505: Detail, but we can avoid the `not` so there is less negation to understand.
```suggestion
        if os.path.exists(pred_path):
            current_last_update = self._prediction_tables_last_update[table_key]
            newest_pred_ds, last_update = self.load_latest_cache(pred_path, current_last_update)
            if newest_pred_ds:
                self._prediction_tables[table_key] = newest_pred_ds
                self._prediction_tables_last_update[table_key] = last_update
        else:
            empty_ds = Dataset.from_dict({"pred_row_idx": list(range(self.num_rows))})
            self._prediction_tables[table_key] = self._init_dataset_split(
                empty_ds, self._prediction_tags
            ).remove_columns([DatasetColumn.row_idx])
            self.save_prediction_table(table_key)
```

azimuth/dataset_split_manager.py
```diff
@@ -171,10 +185,13 @@ def _load_base_dataset_split(self, dataset_split) -> Tuple[Dataset, Dataset]:
...
         with FileLock(self._file_lock):
-            self._base_dataset_split.save_to_disk(self._get_new_version_path(self._save_path))
-            self._malformed_dataset.save_to_disk(self._get_new_version_path(self._malformed_path))
-        self.last_update = time.time()
-        log.debug("Base dataset split saved.", path=self._save_path)
+            version_path, last_update = self._get_new_version_path(self._save_path)
+            self._base_dataset_split.save_to_disk(version_path)
+            self._malformed_dataset.save_to_disk(
+                self._get_new_version_path(self._malformed_path)[0]
+            )
```
JosephMarinier, line 192: The `malformed, _ = ` you wrote above made it super clear! 👍 We could do the same here?
```suggestion
            malformed, _ = self._get_new_version_path(self._malformed_path)
            self._malformed_dataset.save_to_disk(malformed)
```

azimuth/dataset_split_manager.py
```diff
@@ -113,6 +122,11 @@ def get_dataset_split(self, table_key: Optional[PredictionTableKey] = None) -> D
         Returns:
             Dataset with predictions if available.
         """
+        current_last_update = self._base_dataset_split_last_update
+        newest_base_ds, last_update = self.load_latest_cache(self._save_path, current_last_update)
+        if newest_base_ds:
+            self._base_dataset_split = newest_base_ds
```
JosephMarinier, line 128: This is "newest" in the sense of "latest", tight? I say we keep it consistent.
```suggestion
        latest_base_ds, last_update = self.load_latest_cache(self._save_path, current_last_update)
        if latest_base_ds:
            self._base_dataset_split = latest_base_ds
```

azimuth/dataset_split_manager.py
```diff
@@ -88,22 +89,30 @@ def __init__(
...
-            self._base_dataset_split, self._malformed_dataset = self._load_base_dataset_split(
+            self._base_dataset_split, self._malformed_dataset = self._process_base_dataset_split(
                 dataset_split
             )
             self._save_base_dataset_split()
         else:
-            self._base_dataset_split, self._malformed_dataset = cached_dataset_split
+            self._base_dataset_split, self._malformed_dataset = cached_base_dataset_split
         self._prediction_tables: Dict[PredictionTableKey, Dataset] = {}
+        self._prediction_tables_last_update: Dict[PredictionTableKey, Time] = defaultdict(int)
```
JosephMarinier, line 106: `int` -> `float`?
```suggestion
        self._prediction_tables_last_update: Dict[PredictionTableKey, Time] = defaultdict(float)
```

azimuth/dataset_split_manager.py
```diff
@@ -473,37 +490,38 @@ def _get_prediction_table(self, table_key: PredictionTableKey) -> Dataset:
...
-        )
-        return pt
+        return pjoin(folder, f"{table_name}_cache_ds.arrow")
 
     def save_prediction_table(self, table_key: PredictionTableKey):
         """Save the prediction to disk."""
         with FileLock(self._file_lock):
-            pt = self._prediction_path(table_key=table_key)
-            self._prediction_tables[table_key].save_to_disk(self._get_new_version_path(pt))
-        self.last_update = int(time.time())
```
JosephMarinier, line 506: Just to be sure, that is why there was `int`s for the time everywhere before, right? We were flooring to the second.

azimuth/dataset_split_manager.py
gabegma: Perfect, that works! I never know when to use the `*`.

azimuth/dataset_split_manager.py
gabegma: Looks great!! Thanks!!

azimuth/dataset_split_manager.py
gabegma: Good point!

azimuth/dataset_split_manager.py
gabegma: Good point!

azimuth/dataset_split_manager.py
gabegma: Oh good catch! Yeah I think there was a mixture of `int` and `float` and I converted everything to float at some point.

azimuth/dataset_split_manager.py
gabegma: Yes, but I think it was `float` in a few places. Would you prefer we pick `int` everywhere? I was afraid is was not precise enough.

azimuth/dataset_split_manager.py
```diff
@@ -171,10 +185,13 @@ def _load_base_dataset_split(self, dataset_split) -> Tuple[Dataset, Dataset]:
     def _save_base_dataset_split(self):
         # NOTE: We should not have the Index in `self.dataset_split`.
         with FileLock(self._file_lock):
-            self._base_dataset_split.save_to_disk(self._get_new_version_path(self._save_path))
-            self._malformed_dataset.save_to_disk(self._get_new_version_path(self._malformed_path))
-        self.last_update = time.time()
```
gabegma, line 176: Like here it was `float`.

azimuth/dataset_split_manager.py
JosephMarinier: I don't see any advantage to getting rid of that precision. I like the float!

JosephMarinier: If ever there is a problem with floats, or if for example, we get issues with having the file names having a variable length or containing a `.` (like conflicting with the file extension), we can change for ints and still keep the precision, by using `time.time_ns()` instead of `time.time()`.

JosephMarinier (approved):
Cool cool cool! 👍

gabegma:
> Cool cool cool! 👍

Thank you for the quick review today and all of your contributions!!

azimuth/dataset_split_manager.py
gabegma: Oh cool! To keep in mind.

azimuth/dataset_split_manager.py
JosephMarinier: I like the `*` operator. I'll prepare a short presentation to the team so we all understand it well.

PR #529, opened 2023-04-03
Refactor ArtifactManager and restart TaskManager

### LINKED ISSUE #521: Tasks often get lost because a worker dies.

gabegma:
Fixes #521 

## Description:
* `ArtifactManager` is now a real Singleton (1 per process). 3 are created -> 1 per worker and 1 in the main thread.
   * By extension, fewer DMs are created. Different tasks will access the same DM.
* The client is now restarted after the startup tasks. This empties the memory, creating new ArtifactManagers. This avoids that the pipeline and the sentence encoder are kept in memory after the startup.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/app.py
+++ b/azimuth/app.py
@@ -25,14 +25,13 @@ from starlette.status import (
 
 from azimuth.config import AzimuthConfig, load_azimuth_config
 from azimuth.dataset_split_manager import DatasetSplitManager
-from azimuth.modules.base_classes import DaskModule
+from azimuth.modules.base_classes import ArtifactManager, DaskModule
 from azimuth.startup import startup_tasks
 from azimuth.task_manager import TaskManager
 from azimuth.types import DatasetSplitName, ModuleOptions, SupportedModule
 from azimuth.utils.cluster import default_cluster
 from azimuth.utils.conversion import JSONResponseIgnoreNan
 from azimuth.utils.logs import set_logger_config
-from azimuth.utils.project import load_dataset_split_managers_from_config
 from azimuth.utils.validation import assert_not_none
 
 _dataset_split_managers: Dict[DatasetSplitName, Optional[DatasetSplitManager]] = {}
@@ -296,6 +295,32 @@ def create_app() -> FastAPI:
     return app
 
 
+def load_dataset_split_managers_from_config(
+    azimuth_config: AzimuthConfig,
+) -> Dict[DatasetSplitName, Optional[DatasetSplitManager]]:
+    """
+    Load all dataset splits for the application.
+
+    Args:
+        azimuth_config: Azimuth Configuration.
+
+    Returns:
+        For all DatasetSplitName, None or a dataset_split manager.
+
+    """
+    artifact_manager = ArtifactManager.instance()
+    dataset = artifact_manager.get_dataset_dict(azimuth_config)
+
+    return {
+        dataset_split_name: None
+        if dataset_split_name not in dataset
+        else artifact_manager.get_dataset_split_manager(
+            azimuth_config, DatasetSplitName[dataset_split_name]
+        )
+        for dataset_split_name in [DatasetSplitName.eval, DatasetSplitName.train]
+    }
+
+
 def initialize_managers(azimuth_config: AzimuthConfig, cluster: SpecCluster):
     """Initialize DatasetSplitManagers and TaskManagers.
 
@@ -346,7 +371,6 @@ def run_validation(
     else:
         for pipeline_index in range(len(config.pipelines)):
             run_validation_module(pipeline_index)
-    task_manager.clear_worker_cache()
 
 
 def run_startup_tasks(azimuth_config: AzimuthConfig, cluster: SpecCluster):
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -358,6 +358,15 @@ class ModelContractConfig(CommonFieldsConfig):
             raise ValueError(f"Duplicated pipeline names {pipeline_names}.")
         return pipeline_definitions
 
+    def get_model_contract_hash(self):
+        return md5_hash(
+            self.dict(
+                include=ModelContractConfig.__fields__.keys()
+                - CommonFieldsConfig.__fields__.keys(),
+                by_alias=True,
+            )
+        )
+
 
 class MetricsConfig(ModelContractConfig):
     # Custom HuggingFace metrics
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -359,6 +359,7 @@ class ModelContractConfig(CommonFieldsConfig):
         return pipeline_definitions
 
     def get_model_contract_hash(self):
+        """Hash for fields related to model contract only (excluding fields from the parents)."""
         return md5_hash(
             self.dict(
                 include=ModelContractConfig.__fields__.keys()
--- a/azimuth/modules/base_classes/artifact_manager.py
+++ b/azimuth/modules/base_classes/artifact_manager.py
@@ -67,7 +67,8 @@ class ArtifactManager:
     """This class is a singleton which holds different artifacts.
 
     Artifacts include dataset_split_managers, datasets and models for each config, so they don't
-    need to be reloaded many times for a same module.
+    need to be reloaded many times for a same module. Inspired from
+    https://stackoverflow.com/questions/31875/is-there-a-simple-elegant-way-to-define-singletons.
     """
 
     def __init__(self):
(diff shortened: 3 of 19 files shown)


azimuth/app.py
```diff
@@ -297,6 +296,32 @@ def create_app() -> FastAPI:
     return app
 
 
+def load_dataset_split_managers_from_config(
```
gabegma, line 299: I needed to move this function because of circular import issues. 
One big change that I made here is to use an `ArtifactManager` in the main thread. I was afraid that multiple dms were created otherwise.

azimuth/modules/base_classes/artifact_manager.py
```diff
@@ -125,6 +159,6 @@ def get_metric(self, config, name: str, **kwargs):
         return self.metrics[hash]
 
     @classmethod
-    def clear_cache(cls) -> None:
-        with Lock():
-            cls.instance = None
+    def instance(cls):
```
gabegma, line 162: `mypy` would complain without that.

azimuth/routers/export.py
```diff
@@ -164,20 +167,23 @@ def get_export_perturbed_set(
 
 
 def make_utterance_level_result(
-    dm: DatasetSplitManager, results: List[List[PerturbedUtteranceResult]], pipeline_index: int
+    dm: DatasetSplitManager,
+    results: List[List[PerturbedUtteranceResult]],
+    pipeline_index: int,
+    config: AzimuthConfig,
```
gabegma, line 173: This was causing an error in a test. We shouldn't rely on the config of the dataset split manager, because it only depends on the `project_hash`.

azimuth/startup.py
```diff
@@ -251,12 +249,16 @@ def startup_tasks(
...
     mods = start_tasks_for_dms(config, dataset_split_managers, task_manager, start_up_tasks)
 
-    # Start a thread to monitor the status.
-    th = threading.Thread(
-        target=wait_for_startup, args=(mods, task_manager), name=START_UP_THREAD_NAME
-    )
-    th.setDaemon(True)
-    th.start()
+    startup_ready = all(m.done() for m in mods.values())
+    if startup_ready:
```
gabegma, line 253: This avoids calling the `task_manager.restart()` twice if the startup tasks are not launched.

tests/test_task_manager.py
```diff
@@ -63,29 +63,6 @@ def get_module_data(simple_text_config):
     )
 
 
-def test_clearing_cache(tiny_text_config):
```
gabegma, line 66: I'm not sure if we need a new test to replace this one - but this one didn't look relevant anymore. Replacing `task_manager.clear_worker_cache()` by `task_manager.restart()` would work, but then the test takes a lot of time (~50 seconds), and I'm not sure it is testing something relevant.

azimuth/task_manager.py
```diff
@@ -218,6 +218,7 @@ def clear_worker_cache(self):
         self.client.run(ArtifactManager.clear_cache)
 
     def restart(self):
-        # Clear futures to free memory.
+        log.info("Cluster restarted to free memory.")
```
JosephMarinier, line 217: SHould that be at the end of the method to be rigorous?

azimuth/modules/base_classes/artifact_manager.py
```diff
@@ -18,31 +19,66 @@
 Hash = int
 
 
+log = structlog.get_logger()
+
+
+class Singleton:
```
JosephMarinier, line 25: I guess you borrowed that from the Internet? Should we note the reference in a comment?

azimuth/config.py
```diff
@@ -358,6 +358,15 @@ def check_pipeline_names(cls, pipeline_definitions):
             raise ValueError(f"Duplicated pipeline names {pipeline_names}.")
         return pipeline_definitions
 
+    def get_model_contract_hash(self):
+        return md5_hash(
+            self.dict(
+                include=ModelContractConfig.__fields__.keys()
+                - CommonFieldsConfig.__fields__.keys(),
```
JosephMarinier, line 365: Is there a reason for excluding `CommonFieldsConfig.__fields__.keys()` and not `exclude_fields_from_cache(self)`? It's just simpler?
```suggestion
                exclude=exclude_fields_from_cache(self),
```

gabegma: I want to remove everything in `ProjectConfig` too here.

azimuth/modules/base_classes/artifact_manager.py
gabegma: I took it from StackOverflow and adapted it a bit. I don't think we need reference in this case?

azimuth/config.py
JosephMarinier: Oh right, because `ProjectConfig` in turn inherits from other classes, and we want to exclude those fields. I see. It was not obvious to me. Do you think we should add a comment justifying what we want here?

azimuth/modules/base_classes/artifact_manager.py
JosephMarinier: We don't _need_ a reference, but it could help if we get issues with this bit of code. 🤷‍♂️

azimuth/config.py
gabegma: I can add a small docstring!

azimuth/task_manager.py
gabegma: Good idea! Also, I'm also wondering if I still need to empty `self.current_tasks` or if the `self.client.restart()` is enough.

azimuth/startup.py
gabegma: Actually, I decided to no longer do the restart after the validation tasks. I don't think it is that useful given the pipeline will be loaded just after and it adds a delay. Also, the validation modules are cached now, so there is no need for a restart in that case.

azimuth/modules/base_classes/artifact_manager.py
gabegma: Ok good idea! I'll add it.

azimuth/startup.py
JosephMarinier: Right! And can you remind me why we need to `task_manager.restart()` after the startup tasks (if they run)?

azimuth/task_manager.py
JosephMarinier: You mean removing the `module.future = None` thing? I don't know. Is it a garbage collector thing that makes the `module.future` stay in memory until we explicitly remove our reference to it?

azimuth/modules/base_classes/artifact_manager.py
```diff
@@ -18,31 +19,66 @@
...
+    """
+    A non-thread-safe helper class to ease implementing singletons.
```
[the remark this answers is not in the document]
JosephMarinier: Is that issue new as of this PR? Or unrelated?

azimuth/task_manager.py
gabegma: I actually get an error when I don't delete the futures! So let's leave it as is.

azimuth/modules/base_classes/artifact_manager.py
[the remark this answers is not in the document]
gabegma: I see it on main too! Maybe it's also just on my computer?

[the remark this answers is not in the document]
gabegma: My comment earlier is not correct - it used to be visible in the logs because when I was developing, I added the logs as `log.info()`, but it's no longer showing since I converted to `log.debug()`. But with `log.info()` I could see one ArtifactManager created by process.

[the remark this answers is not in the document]
JosephMarinier: Sorry, I'm not following. Is the issue real? Is it also real on `main`?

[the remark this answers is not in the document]
gabegma: Yes sorry! The issue is real on my computer, even on main. I don't think we would have notice before.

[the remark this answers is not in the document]
JosephMarinier: Ok, thanks!

azimuth/task_manager.py
JosephMarinier: Fine by me!

azimuth/startup.py
JosephMarinier: Makes sense, thanks!

tests/test_task_manager.py
JosephMarinier: I don't really understand the assertions `any([m and d for m, d in cached.values()])`. I leave it up to you! Saving 50 seconds does sound tempting if it's useless.

JosephMarinier (approved):
LGTM! I don't understand all the details, but I guess we'll confirm that everything works by using it!

PR #538, opened 2023-04-12
Fix startup dependencies

gabegma:
## Description:
As I was debugging, I noticed a few problems related to dependencies.
- First `if dependencies is not None` would usually not work since `dependencies` is often an empty list. It would `secede()` and `rejoin()` potentially for nothing.
- The dependencies were working a bit by magic, because if I would be in `debug` mode and go through the startup task row per row (slowly), some modules would start without waiting for their dependencies. Somewhat it wasn't causing an issue when running at normal speed, but it's safer to define the tasks in `start_up_tasks` so that any dependencies are already defined in the list. I added a `TODO` so we can improve this in the future.
- If a startup name was a substring of another startup name, the module would wrongly identify its dependencies. I also added a `TODO`.
- I increased the occurrence at which we check that the startup is done, so we don't wait longer than necessary. This should also improve some tests speed.

## Checklist:

  our users.
* [ ] **DEV CHANGES.**


azimuth/startup.py
```diff
@@ -120,6 +120,9 @@ class Startup:
         dependency_names=["prediction"],
         run_on_all_pipelines=True,
     ),
+]
+
+LAST_TASKS = [
```
JosephMarinier, line 125: As discussed offline, your comment already makes it clear that the order is important, and we can't rename all those constants with names that indicate in which order they should be added, so I don't think it's necessary for this one either. We could go with something more similar to the other ones:
```suggestion
BY_FILTER_TASKS = [
```

gabegma: Sounds good!

PR #539, opened 2023-04-12
Avoid restarting task when pending

gabegma:
## Description:
* We had an issue where tasks would be restarted over and over again if the user would refresh the browser before the task was done. This PR fixes it.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/modules/base_classes/dask_module.py
+++ b/azimuth/modules/base_classes/dask_module.py
@@ -233,6 +233,9 @@ class DaskModule(HDF5CacheMixin, Generic[ConfigScope]):
         if self.done_event is not None:
             self.done_event.wait()
 
+    def should_be_started(self):
+        return self.status() not in {"pending", "finished"} and not self.done()
+
     def status(self):
         """Return the status of the future.
 
--- a/azimuth/modules/task_execution.py
+++ b/azimuth/modules/task_execution.py
@@ -26,8 +26,7 @@ def get_task_result(task_module: Module, result_type: Type[U]) -> U:
     Returns:
         The result of the Module.
     """
-    if not task_module.done():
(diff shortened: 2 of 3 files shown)


azimuth/modules/base_classes/dask_module.py
```diff
@@ -233,6 +233,9 @@ def wait(self):
         if self.done_event is not None:
             self.done_event.wait()
 
+    def is_pending(self):
+        return self.status() == "pending"
```
JosephMarinier, line 237: I think we have to check for the status `"finished"` too, since the task could be `"finished"` but not `done()` because some callbacks have yet to complete.

Also, I think this method should do both `not task.done() and not task.is_pending()`, to avoid repeating that (non so straightforward) logic in two places. For example, it could be:
```suggestion
    def started(self):
        return self.status() in {"pending", "finished"} or self.done()
```
and we call `if not task.started()`, then `task.start...`.

What do you think?

gabegma: I like putting everything in the same method, but I struggle with naming the method `started`, for the use case where it’s in the cache. Especially that the status will be "not started" in that case. 
- Maybe we make explicit, `done_or_pending`? 
- What about we reverse the logic and call the method `should_be_launched`, or something like that?

JosephMarinier: I like `should_be_launched`! 👍

JosephMarinier: Or let's keep `started` instead of `launched`, since the action that follows is `task.start_task...`.

JosephMarinier (approved):
👌

PR #542, opened 2023-04-12
Refactor validation modules

gabegma:
## Description:
* The validation modules will now leverage the caching. 
    * This will improve the speed when updating the config with fields that would not impact the validation module. 
    * It will also improve speed when developing. 
    * The only caveat is that if users modify their pipelines/data without changing the config itself, the validation modules won't catch the error. I think that we can live with it.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


JosephMarinier (approved):
Let's gooo! 👍

PR #543, opened 2023-04-12
Cleanup tasks in memory

gabegma:
## Description:
* Some tasks would be kept in memory forever. This PR makes sure to delete all futures that might be lingering.

## Checklist:

  our users.
* [x] **DEV CHANGES.**


azimuth/modules/base_classes/dask_module.py
```diff
@@ -217,6 +217,10 @@ def on_end(self, fut, module):
             self._store_data_in_cache(fut.result(), fut.indices)
             log.info(f"{self.name} completed and stored in cache", status=fut.status)
 
+    def delete_future(self):
```
JosephMarinier, line 220: The only question I have is whether the caller (from outside `dask_module.py`) needs to know HOW this works (setting `future` to `None`), or if it would be more user-friendly to simply say WHAT it does (not mentioning that it involves some `future` poutine), for example
```suggestion
    def clear(self):
```

JosephMarinier (approved):
I have one question/comment, and even if we decide to make a change, it could totally be in a later PR, since that LGTM as is.

azimuth/modules/base_classes/dask_module.py
gabegma: I would be ok with it! I'll make the change.

JosephMarinier: Cool! THat way, we keep the implementation for, well, the implementation. 👍

PR #544, opened 2023-04-12
Fix get_tags() method

gabegma:
## Description:
* Major issue that I introduced a while ago: `get_tags()` could take minutes now to compute on bigger datasets. This will fix a major bug where the app would no longer load on bigger datasets.
* You can see the profiling before and after.
[logs_before.txt](https://github.com/ServiceNow/azimuth/files/11215771/logs_before.txt)
[logs.txt](https://github.com/ServiceNow/azimuth/files/11215772/logs.txt)

## Checklist

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/dataset_split_manager.py
+++ b/azimuth/dataset_split_manager.py
@@ -326,10 +326,7 @@ class DatasetSplitManager:
             ds = ds.select(indices)
         available_tags = self._tags if table_key is None else self._tags + self._prediction_tags
 
-        return {
-            row_idx: {tag: ds[tag][idx] for tag in available_tags}
-            for idx, row_idx in enumerate(ds["row_idx"])
-        }
+        return {row[DatasetColumn.row_idx]: {tag: row[tag] for tag in available_tags} for row in ds}
 
     def class_distribution(self, labels_only=False):
         """Compute the class distribution for a dataset_split.


azimuth/dataset_split_manager.py
```diff
@@ -326,10 +326,7 @@ def get_tags(
             ds = ds.select(indices)
         available_tags = self._tags if table_key is None else self._tags + self._prediction_tags
 
-        return {
-            row_idx: {tag: ds[tag][idx] for tag in available_tags}
-            for idx, row_idx in enumerate(ds["row_idx"])
-        }
+        return {row[DatasetColumn.row_idx]: {tag: row[tag] for tag in available_tags} for row in ds}
```
JosephMarinier, line 329: Do you think that `ds.map()` could be faster than a `for` loop on `ds`? Something like this (not tested):
```suggestion
        return dict(
            ds.map(
                lambda row: (row[DatasetColumn.row_idx], {tag: row[tag] for tag in available_tags}),
                input_columns=[DatasetColumn.row_idx, *available_tags],
            )
        )
```

gabegma: I'll profile and get back to you!!

JosephMarinier: In fact, I just realized that in the two places that we call `get_tags()`, we then do `.values()`, so I think we can avoid the `dict` altogether!
```suggestion
        ds.map(
            lambda row: {tag: row[tag] for tag in available_tags},
            input_columns=[DatasetColumn.row_idx, *available_tags],
        )
```

gabegma: The goal was to be consistent with the `add_tags()` method - might not be needed though.

JosephMarinier: I see.

gabegma: I get `TypeError: <lambda>() takes 1 positional argument but 24 were given`. I'm not familiar enough with the `map() ` function to debug quickly.

JosephMarinier: Oh, right! Forget my `input_columns`. Without it, it should work, I think.

JosephMarinier: I am not familiar at all. I might be completely off. We can forget about it.

JosephMarinier (approved):
Wow! Good catch! Thank you very much!

PR #547, opened 2023-04-14
Don't block execution when waiting for startup tasks

JosephMarinier:
## Description:

* [Avoid checking `.done()` twice](https://github.com/ServiceNow/azimuth/pull/547/commits/1c7d7070d307971b8312889d130329d4ea028797)
* [**Check** for startup status a lot more frequently](https://github.com/ServiceNow/azimuth/pull/547/commits/41c081dfcbba15fbeee7d6896864e75c52806485) and **log** status updates only when there are some changes (to avoid spamming the user with logs).
* [Log progress in a separate thread](https://github.com/ServiceNow/azimuth/pull/547/commits/817524fb6ef0a99db4cb3e299c11d1e3915005ac)
* [Always log statuses in the same (logical) order](https://github.com/ServiceNow/azimuth/pull/547/commits/6d242fb5e8d071807164ab1056ad103db52b770a) and in one multiline log to make sure no logs from other processes get inserted in between.

image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/startup.py
+++ b/azimuth/startup.py
@@ -338,9 +338,15 @@ def wait_for_startup(startup_mods: Dict[str, DaskModule], task_manager: TaskMana
                 continue
             last_per_status = per_status
 
-            log.info(f"Startup tasks statuses: {len(per_status['finished'])}/{len(startup_mods)}")
-            for status, modules in per_status.items():
-                log.info(f"{status} ({len(modules)}): {', '.join(modules)}")
+            logs = [
+                f"Startup tasks statuses: {len(per_status['finished'])}/{len(startup_mods)}",
+                *(
+                    f"{status} ({len(per_status[status])}): {', '.join(per_status[status])}"
+                    for status in ("not_started", "pending", "lost", "error", "saving", "finished")
+                    if status in per_status
+                ),
+            ]
+            log.info("\n\t".join(logs))
 
     thread_log_progress = threading.Thread(target=log_progress)
     thread_log_progress.start()
--- a/azimuth/startup.py
+++ b/azimuth/startup.py
@@ -322,21 +322,32 @@ def wait_for_startup(startup_mods: Dict[str, DaskModule], task_manager: TaskMana
     """
     start_time = time.time()
     task_manager.lock()  # Lock the TaskManager to prevent new tasks.
-    last_per_status: Dict[str, List[str]] = {}
-    while not all(m.done() for m in startup_mods.values()):
-        time.sleep(0.1)  # to let the CPU do other stuff.
-        per_status = defaultdict(list)
-        for name, mod in startup_mods.items():
-            status = "saving" if mod.status() == "finished" and not mod.done() else mod.status()
-            per_status[status].append(name)
 
-        if per_status == last_per_status:
-            continue  # to avoid spamming the user with logs.
-        last_per_status = per_status
+    done = False
 
-        log.info(f"Startup tasks statuses: {len(per_status['finished'])}/{len(startup_mods)}")
-        for status, modules in per_status.items():
-            log.info(f"{status} ({len(modules)}): {', '.join(modules)}")
+    def log_progress():
+        last_per_status: Dict[str, List[str]] = {}
+        while not done:
+            time.sleep(5)  # to avoid spamming the user with logs.
+            per_status = defaultdict(list)
+            for name, mod in startup_mods.items():
+                status = "saving" if mod.status() == "finished" and not mod.done() else mod.status()
+                per_status[status].append(name)
+
+            if per_status == last_per_status:
+                continue
+            last_per_status = per_status
+
+            log.info(f"Startup tasks statuses: {len(per_status['finished'])}/{len(startup_mods)}")
+            for status, modules in per_status.items():
+                log.info(f"{status} ({len(modules)}): {', '.join(modules)}")
+
+    thread_log_progress = threading.Thread(target=log_progress)
+    thread_log_progress.start()
+    for mod in startup_mods.values():
+        mod.result()
+
+    done = True
 
     log.info("Startup task completed. The application should be accessible now.")
     log.debug(f"Startup took {time.time() - start_time}.")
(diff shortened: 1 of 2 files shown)


gabegma:
I really like the idea!! However, when I tried, I felt more spammed with logs than before; it prints so quickly that it can be hard to follow, as blocks of logs follow each other and it can be hard to see where it starts and ends. I tried a sleep of 5 seconds, and I think it is easier to follow what is going on. What do you think?

azimuth/startup.py
```diff
@@ -322,14 +322,19 @@ def wait_for_startup(startup_mods: Dict[str, DaskModule], task_manager: TaskMana
     """
     start_time = time.time()
     task_manager.lock()  # Lock the TaskManager to prevent new tasks.
+    last_per_status: Dict[str, List[str]] = {}
     while not all(m.done() for m in startup_mods.values()):
-        time.sleep(30)  # We wait to not spam the user with logs.
-        is_done = {k: v.done() for k, v in startup_mods.items()}
+        time.sleep(0.1)  # to let the CPU do other stuff.
```
JosephMarinier, line 327: @gabegma
> I really like the idea!! However, when I tried, I felt more spammed with logs than before; it prints so quickly that it can be hard to follow, as blocks of logs follow each other and it can be hard to see where it starts and ends. I tried a sleep of 5 seconds, and I think it is easier to follow what is going on. What do you think?

I really hate the idea of sleeping any significant amount of time and blocking the execution.

Other solutions:
* Logging progress only when `per_status['finished']` changes (ignoring the other statuses), but that might be too little logging for bid datasets.
* Logging progress only if 5 seconds have elapsed since the last log, without blocking execution. That could be done by counting the iterations and printing once every 50 iterations for example.
* Logging progress in another thread, which can sleep, leaving the main thread simply waiting for the module's `result()`s.

I'll see what we can do.

JosephMarinier: I went with 3, let me know what you think: [Log progress in a separate thread](https://github.com/ServiceNow/azimuth/pull/547/commits/817524fb6ef0a99db4cb3e299c11d1e3915005ac)

azimuth/startup.py
```diff
@@ -322,16 +322,38 @@ def wait_for_startup(startup_mods: Dict[str, DaskModule], task_manager: TaskMana
...
-            status = "saving" if mod.status() == "finished" and not mod.done() else mod.status()
-            per_status[status].append(name)
-        log.info(f"Startup tasks statuses: {sum(is_done.values())}/{len(is_done)}")
-        for status, modules in per_status.items():
-            log.info(f"{status} ({len(modules)}): {', '.join(modules)}")
+
+    done = False
+
+    def log_progress():
+        last_per_status: Dict[str, List[str]] = {}
```
gabegma, line 329: Should this be a `defaultdict()` too, for consistency with `per_status`?

azimuth/startup.py
```diff
@@ -322,16 +322,38 @@ def wait_for_startup(startup_mods: Dict[str, DaskModule], task_manager: TaskMana
...
+                f"Startup tasks statuses: {len(per_status['finished'])}/{len(startup_mods)}",
+                *(
+                    f"{status} ({len(per_status[status])}): {', '.join(per_status[status])}"
+                    for status in ("not_started", "pending", "lost", "error", "saving", "finished")
+                    if status in per_status
+                ),
+            ]
+            log.info("\n\t".join(logs))
+
+    thread_log_progress = threading.Thread(target=log_progress)
```
gabegma, line 351: I don't understand the consequences of threading - there is no problem starting a thread from a thread, right? It can’t cause performance issues?

azimuth/startup.py
```diff
@@ -349,3 +371,5 @@ def wait_for_startup(startup_mods: Dict[str, DaskModule], task_manager: TaskMana
     # After restarting, it is safe to unlock the task manager.
     task_manager.unlock()
     log.info("Cluster restarted to free memory.")
+
+    thread_log_progress.join()
```
gabegma, line 375: Should we do that earlier? `task_manager.restart()` empties the futures. Could it cause a problem with the logging?

gabegma (approved):
I love this - this is great!!! The logs are prettier too. Thanks a lot for tackling this! I learned about threading a bit more.

azimuth/startup.py
gabegma: I wonder if we still need the thread that is started in `startup_tasks` on line 255.

azimuth/startup.py
JosephMarinier: The point of creating this new thread is to not have to wait for it to finish before restarting and unlocking the `task_manager`.
To be honest, I added a `.join()` here when I was debugging, to make sure I undertook what was going on. It is not necessary here.
As for whether or not `log_progress()` might run into an issue if it runs after `task_manager.restart()` emptied the futures, I don't see a problem. Do you see a problem calling `mod.done()` or `mod.status()`?

azimuth/startup.py
JosephMarinier: I don't see any advantage. I went with that since a `dict` is a little bit quicker to instantiate than a `default_dict`. I can also initialize `last_per_status = None`, even quicker, and mypy is happy without any typing. Thank you for having me revisit this bit.

azimuth/startup.py
gabegma: Right probably! Let's leave it like that if a thread from a thread is no issue.

azimuth/startup.py
gabegma: Tasks might appear as being not started again? I thought the `join()` was to kill the thread - does it naturally ends when done is set to `True`?

JosephMarinier: Yes. `join()` never "kills" a thread, it just waits (hanging) for it to die.

gabegma: Ok good, so we shouldn't see any issue then.

JosephMarinier:
Cool! Thank you! I learned it too!

PR #554, opened 2023-04-21
Return plain text errors

JosephMarinier:
## Description:

Don't worry! 199 of those line changes are in the `generatedTypes.ts`.

We used to expect error responses to have a JSON body like `{"detail": string}`. This was dangerous as the front end would fail if it encountered invalid JSON or plain text. And, for some obscure reason, we were still getting some errors in plain text, including some 500 Internal Server Error even though we had a handler for `HTTP_500_INTERNAL_SERVER_ERROR: handle_internal_error` that was returning some JSON. Also, the front end could receive network errors that don't come from the back end, for example if it is unable to reach the back end.

So, the easiest solution is to expect plain text from all error responses. To our knowledge, by default, only `HTTPException`s and `ValidationError`s used to return JSON, so I propose changing those handlers to plain text.

Worst case, if an error turns out to be JSON, it will still be valid text. The other way around is not true.

Also, this results in 21 fewer lines of code. 🤷‍♂️ 

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/app.py
+++ b/azimuth/app.py
@@ -9,7 +9,7 @@ import structlog
 from distributed import SpecCluster
 from fastapi import APIRouter, Depends, FastAPI, HTTPException
 from fastapi.exceptions import RequestValidationError
-from pydantic import BaseModel, ValidationError
+from pydantic import ValidationError
 from starlette.middleware.cors import CORSMiddleware
 from starlette.status import (
     HTTP_400_BAD_REQUEST,
@@ -30,7 +30,7 @@ from azimuth.types import DatasetSplitName, ModuleOptions, SupportedModule
 from azimuth.utils.cluster import default_cluster
 from azimuth.utils.conversion import JSONResponseIgnoreNan
 from azimuth.utils.exception_handlers import (
-    handle_internal_error,
+    handle_http_exception,
     handle_validation_error,
 )
 from azimuth.utils.logs import set_logger_config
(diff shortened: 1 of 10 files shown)


gabegma:
Thank you for tackling the bug! Is it possible to hide the error message for unexpected errors? I don't think that we should be showing them in the UI in this case. 
Screen Shot 2023-04-24 at 18 54 05

gabegma:
It also shows up in plots. 
Screen Shot 2023-04-24 at 19 01 43

azimuth/app.py
```diff
@@ -167,12 +163,15 @@ def create_app() -> FastAPI:
...
+        responses={
+            code: {"content": {"text/plain": {"schema": {"type": "string"}}}}
+            for code in COMMON_HTTP_ERROR_CODES
+        },
         exception_handlers={
+            HTTPException: handle_http_exception,
             ValidationError: handle_validation_error,  # for PATCH "/config",
             # where we call old_config.copy(update=partial_config, deep=True) ourselves.
             RequestValidationError: handle_validation_error,
-            HTTP_500_INTERNAL_SERVER_ERROR: handle_internal_error,
```
JosephMarinier, line 175: @gabegma
> Is it possible to hide the error message for unexpected errors? I don't think that we should be showing them in the UI in this case.

I would like that too, but I don't know how. For the examples you shared, the API response didn't change in this PR. All that changed is that on top of that error, the front end used to crash when trying to parse the error, as it was expecting some JSON, and showing something like this instead of the API error:
> Uncaught SyntaxError: Unexpected token 'S', "Something "... is not valid JSON

Even when I tried to explicitly handle the 500 errors with this `handle_internal_error` here and to return a simpler message ("Internal server error"), it was not working in some cases, like in the example you shared.

I think we should discuss prioritizing another task for surveying what errors to expect and try to catch them or handle them differently if necessary.

Does that make sense?

JosephMarinier:
Closing in favor of #556

PR #558, opened 2023-05-01
Support null dataset

JosephMarinier:
## Description:

Hello! I know there are a lot of commits, but they are relatively small and straightforward, so I started with one PR. I recommend going through them one by one. Here is a list of who I think should review which commit. Some commits could be independent, so if some changes need more attention than others and it's possible, I'll be happy to move them to separate PRs. Let me know.

@gabegma
* [Clean up unused `datasetInfo.classNames`](https://github.com/ServiceNow/azimuth/pull/558/commits/ddd93b073bb14332299eb09169fe7e5eb47aa12d)
* [Clean up `datasetInfo`](https://github.com/ServiceNow/azimuth/pull/558/commits/7aaa2aba3fa637e1d10d7f4ec40eece3d9197923) 
  - Replace `datasetInfo.evalClassDistribution` and `datasetInfo.trainClassDistribution` with `datasetInfo.utteranceCountPerDatasetSplit`
* [Clean up `object_loader.py`](https://github.com/ServiceNow/azimuth/pull/558/commits/c254c59ecea9103eccbfd0f7c39557244765d4e2)
* [Handle more `AzimuthValidationErrors` when loading a `CustomObject`](https://github.com/ServiceNow/azimuth/pull/558/commits/ea24ad5d87801e64ba328ff5e0cb60a14ccd1940) instead of unexpected internal server errors.

@christyler3030
* [Fix nullable `StringField`](https://github.com/ServiceNow/azimuth/pull/558/commits/f20aef1c78cd943f5e2c153ded89c686b89ea489)
* [Fix dataset columns repeated labels](https://github.com/ServiceNow/azimuth/pull/558/commits/ebc306058bfe5d0394af64eac2589f2781bfb5c3)
* [Reuse `handleDiscard()`](https://github.com/ServiceNow/azimuth/pull/558/commits/231cd8d958be7c05c7e28e871f28fbb7ffd61d6f)
* [Render config modal only if it is open (optimization)](https://github.com/ServiceNow/azimuth/pull/558/commits/0002960087c0343b7bf13b56b7fff9940f13f734)
* [Expand the Project Configuration accordion by default](https://github.com/ServiceNow/azimuth/pull/558/commits/af3d1b8266c85bfb5c5a343ffad45d6222c06d66)

@gabegma and @christyler3030
* [Support null dataset](https://github.com/ServiceNow/azimuth/pull/558/commits/8f75a93805299f6080e72ffbbc3a058518e92900)

@christyler3030
[...]

PR #560, opened 2023-05-02
Create an API route to get the config history

JosephMarinier:
## Description:

Create an API route to get the config history along with a hash of each config to differentiate them easily.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, List, Literal, Optional, TypeVar, Union
 
 import structlog
 from jsonlines import jsonlines
-from pydantic import Extra, Field, root_validator, validator
+from pydantic import Extra, Field, ValidationError, root_validator, validator
 
 from azimuth.types import AliasModel, DatasetColumn, SupportedModelContract
 from azimuth.utils.conversion import md5_hash
@@ -474,6 +474,19 @@ class AzimuthConfig(
         else:
             return AzimuthConfigHistory.parse_obj(last_config).config
 
+    def get_config_history(self) -> List["AzimuthConfigHistoryWithHash"]:
+        config_history = []
+        try:
+            with jsonlines.open(self.get_config_history_path(), mode="r") as reader:
+                for item in reader:
+                    try:
+                        config_history.append(AzimuthConfigHistoryWithHash.parse_obj(item))
+                    except ValidationError:
+                        pass
+        except (FileNotFoundError, ValueError):
+            pass
+        return config_history
+
     def log_info(self):
         log.info(f"Config loaded for {self.name} with {self.model_contract} as a model contract.")
 
@@ -517,6 +530,14 @@ class AzimuthConfig(
 class AzimuthConfigHistory(AzimuthBaseSettings):
     config: AzimuthConfig
     created_on: str = Field(default_factory=lambda: str(datetime.now(timezone.utc)))
+
+
+class AzimuthConfigHistoryWithHash(AzimuthConfigHistory):
+    hash: str = ""
+
+    @root_validator(skip_on_failure=True)
+    def _set_hash(cls, values):
+        return {**values, "hash": md5_hash(values["config"].dict())}
 
 
 def load_azimuth_config(config_path: Optional[str], load_config_history: bool) -> AzimuthConfig:
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -120,8 +120,8 @@ class TemperatureScaling(CustomObject):
     ] = "azimuth.utils.ml.postprocessing.TemperatureScaling"
     temperature: float = Field(1, ge=0, env="TEMP")
 
-    @root_validator()
-    def check_temps(cls, values):
+    @root_validator
+    def _check_temps(cls, values):
         kwargs = values.get("kwargs", {})
         if "temperature" not in kwargs:
             kwargs["temperature"] = values.get("temperature", 1)
@@ -136,8 +136,8 @@ class ThresholdConfig(CustomObject):
     ] = "azimuth.utils.ml.postprocessing.Thresholding"
     threshold: float = Field(0.5, ge=0, le=1, env="TH")
 
-    @root_validator()
-    def check_threshold(cls, values):
+    @root_validator
+    def _check_threshold(cls, values):
         kwargs = values.get("kwargs", {})
         if "threshold" not in kwargs:
             kwargs["threshold"] = values.get("threshold", 0.5)
@@ -333,7 +333,7 @@ class ModelContractConfig(CommonFieldsConfig):
     saliency_layer: Optional[str] = Field(None, nullable=True)
 
     @validator("pipelines", pre=True)
-    def check_pipeline_names(cls, pipeline_definitions):
+    def _check_pipeline_names(cls, pipeline_definitions):
         # We support both [] and None (null in JSON), and we standardize it to None.
         if not pipeline_definitions:
             return None
@@ -433,8 +433,8 @@ class AzimuthConfig(
     # Reminder: If a module depends on an attribute in AzimuthConfig, the module will be forced to
     # include all other configs in its scope.
 
-    @root_validator()
-    def dynamic_language_config_values(cls, values):
+    @root_validator
+    def _dynamic_language_config_values(cls, values):
         defaults = config_defaults_per_language[values["language"]]
         if behavioral_testing := values.get("behavioral_testing"):
             neutral_token = behavioral_testing.neutral_token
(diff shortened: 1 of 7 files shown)


azimuth/routers/config.py
```diff
@@ -38,7 +39,7 @@
     description="Get the default configuration",
     response_model=AzimuthConfig,
 )
-def get_default_config_def(
+def get_config_default(
```
JosephMarinier, line 42: Not an important change, but I thought `get_config_default()` follows more the URL path for `GET` `/config/default`.

azimuth/config.py
```diff
@@ -519,6 +532,14 @@ class AzimuthConfigHistory(AzimuthBaseSettings):
     created_on: str = Field(default_factory=lambda: str(datetime.now(timezone.utc)))
 
 
+class AzimuthConfigHistoryWithHash(AzimuthConfigHistory):
+    hash: str = ""
+
+    @root_validator(skip_on_failure=True)
+    def _set_hash(cls, values):
+        return {**values, "hash": md5_hash(values["config"].dict(by_alias=False))}
```
gabegma, line 540: We use `by_alias=True` in other hash, was there a reason here to use `False`?

azimuth/config.py
```diff
@@ -474,6 +474,19 @@ def load_last_from_config_history(cls, config_history_path: str) -> Optional["Az
...
+    def get_config_history(self) -> List["AzimuthConfigHistoryWithHash"]:
+        config_history = []
+        try:
+            with jsonlines.open(self.get_config_history_path(), mode="r") as reader:
+                for item in reader:
+                    try:
+                        config_history.append(AzimuthConfigHistoryWithHash.parse_obj(item))
+                    except ValidationError:
+                        pass
+        except (FileNotFoundError, ValueError):
```
gabegma, line 486: When can we get a `ValueError`?

tests/test_routers/test_config.py
```diff
@@ -233,9 +244,7 @@ def test_update_config(app: FastAPI, wait_for_startup_after):
     initial_config = client.get("/config").json()
     initial_contract = initial_config["model_contract"]
     initial_pipelines = initial_config["pipelines"]
-    jsonl_file_path = f"{initial_config['artifact_path']}/config_history.jsonl"
-    with jsonlines.open(jsonl_file_path, "r") as reader:
-        initial_config_count = len(list(reader))
+    initial_config_count = len(client.get("/config/history").json())
```
gabegma, line 247: Nice re-use!

gabegma (approved):
LGTM!! Awesome work! Have you considered refactoring the code in `load_last_from_config_history` to leverage the new method?

azimuth/config.py
JosephMarinier: With `class_name`, when the module exists but the class/function cannot be found in the module.

azimuth/config.py
```diff
@@ -474,6 +474,19 @@ def load_last_from_config_history(cls, config_history_path: str) -> Optional["Az
         else:
             return AzimuthConfigHistory.parse_obj(last_config).config
 
+    def get_config_history(self) -> List["AzimuthConfigHistoryWithHash"]:
+        config_history = []
+        try:
+            with jsonlines.open(self.get_config_history_path(), mode="r") as reader:
+                for item in reader:
```
JosephMarinier, line 481: @gabegma
> Have you considered refactoring the code in `load_last_from_config_history` to leverage the new method?

I don't want to use `get_config_history()` in `load_last_from_config_history()` since the latter only needs to parse the last config, not every single item, which could raise unnecessary validation errors.

Does that make sense?

tests/test_routers/test_config.py
JosephMarinier: 🙂

azimuth/config.py
gabegma: Yes, thanks!!

azimuth/config.py
JosephMarinier: Oh yeah, good question. No reason. As long as it uses the same hash for all the items, it serves the purpose of comparing them.

Anyway, I think `by_alias=True` has no effect on the config since it doesn't have any aliases. For example, our `AliasModel` has an alias generator to make all the field names camelCase for the JSON API. However, `AzimuthConfig` doesn't inherit from `AliasModel`.

Or am I missing something?

If not, I'll remove it.
```suggestion
        return {**values, "hash": md5_hash(values["config"].dict())}
```

azimuth/config.py
JosephMarinier: Wow! I was drunk! That answer applies to [this line](https://github.com/ServiceNow/azimuth/pull/558/files#diff-9d3b3c07d2e92acdf617803569e0a5270200eb4791d0bc8355e65ffb3727c164R115), not here. 😅 

Here, the `ValidationError` happens if a line of the file is not valid JSON. That would happen if someone messed with the file. In this case, I am treating it as if the file didn't exist and I'm returning an empty history. That behavior is up for debate.
1. As is, it's like expected or normal to have a missing/broken config history, in which case I'm just expecting the front end to not show history options.
1. I could instead return a 404 Not found, in which case we would get a red toast in the front end.

I don't think we are resilient to saving the config history when the existing file is not valid, so that might deserve more testing/love.

azimuth/config.py
gabegma: Yeah, I was wondering too if it was making a difference. Perhaps we should remove it in the 2 other places where we generate the hash then, for consistency.

JosephMarinier: I removed them in 19c6780 and I restarted Azimuth, which read from cache as expected, so the hashes seem identical.

gabegma: Good test! Thanks!

JosephMarinier: I committed 3d3e552 too, to clean things up while it's fresh.

JosephMarinier: Quick check if it makes sense to you?

gabegma (commented):
Sorry I forgot to "submit review", my comment was pending.

azimuth/types/general/alias_model.py
```diff
@@ -32,11 +32,9 @@ class AliasModel(BaseModel):
...
     """
 
-    def dict(self, *, by_alias: bool = False, **kwargs: Any) -> Dict[str, Any]:
-        return super().dict(by_alias=True, **kwargs)
-
-    def no_alias_dict(self, **kwargs: Any) -> Dict[str, Any]:
-        return super().dict(by_alias=False, **kwargs)
+    def dict(self, *, by_alias: bool = True, **kwargs: Any) -> Dict[str, Any]:
+        """Simply calling the super, but with by_alias that defaults to True instead of False."""
+        return super().dict(by_alias=by_alias, **kwargs)
```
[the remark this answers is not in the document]
JosephMarinier: I don't know.

My guess was that we are overwriting `dict()` to call `dict(by_alias=True)` so that when FastAPI calls `dict()` to produce the API responses, it uses `by_alias=True` so we get camelCase instead of snake_case JSON. This doesn't happen in our code, it happens in FastAPI's code, so instead of trying to change that behavior, we changed the `dict()` that we control. However, upon testing and debugging, it doesn't seem to be the case! FastAPI to call `dict(by_alias=True)` [here](https://github.com/tiangolo/fastapi/blob/60ef67a66b1e0cadc2f4b4d6ea9f84baec8768f3/fastapi/routing.py#L80) already. So, I don't know why we overload `dict()` like that. 😅 It predates the initial commit. Maybe @Dref360 has an idea why?

I was hesitant to change that behavior as I was afraid to break something subtle without noticing. If add a breakpoint here and I run a random test, it reaches my breakpoint like >100 times, and I can't pretend I understand all the calls made to it.

That said, I tried removing this `dict()` overload completely and running the app and I didn't see any issues. We can try it and see if we get any issues in the coming weeks. What do you think?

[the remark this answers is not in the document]
JosephMarinier: The only difference I can find is that now the perturbation testing summary CSV export has snake_case column headers, just like our other CSV exports. 👍 See 4655539dfe3b6151470f013b3900abe57e60c39f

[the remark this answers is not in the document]
gabegma: Ok! Let's do it! Thanks for trying to understand it further!! Should we make those commit in a separate PR so it's easy to revert? I'll let you decide.

tests/test_modules/test_model_contracts/test_model_contract_module.py
```diff
@@ -58,7 +58,7 @@ def test_pipeline_steps(simple_text_config):
     )
     for order, class_name in enumerate(("TemperatureScaling", "Thresholding")):
         assert (
-            ds[DatasetColumn.pipeline_steps][0]["postprocessing_steps"][order]["className"]
+            ds[DatasetColumn.pipeline_steps][0]["postprocessing_steps"][order]["class_name"]
```
gabegma, line 61: Yeah, here too it's actually better to have it in snake case.

JosephMarinier: Yesss! 💪

PR #561, opened 2023-05-02
Support adding/removing pipelines/post-processors

nandhinibsn:
Resolve #

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -96,16 +96,6 @@ const SUPPORTED_SPACY_MODELS: SupportedSpacyModels[] = [
   "fr_core_news_md",
 ];
 const USE_CUDA_OPTIONS = ["auto", "true", "false"] as const;
-
-type Postprocessor = TemperatureScaling | ThresholdConfig;
-
-type PostprocessorClassName = Postprocessor["class_name"];
-
-const POSTPROCESSORS_CLASS_NAMES: PostprocessorClassName[] = [
-  "azimuth.utils.ml.postprocessing.TemperatureScaling",
-  "azimuth.utils.ml.postprocessing.Thresholding",
-];
-
 type UseCUDAOption = typeof USE_CUDA_OPTIONS[number];
 
 const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof AzimuthConfig)[] = [
@@ -117,6 +107,15 @@ const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof AzimuthConfig)[] = [
   "uncertainty",
   "metrics",
 ];
+
+type KnownPostprocessor = TemperatureScaling | ThresholdConfig;
+
+const KNOWN_POSTPROCESSORS: {
+  [T in KnownPostprocessor as T["class_name"]]: Partial<T>;
+} = {
+  "azimuth.utils.ml.postprocessing.TemperatureScaling": { temperature: 1 },
+  "azimuth.utils.ml.postprocessing.Thresholding": { threshold: 0.5 },
+};
 
 const Columns: React.FC<{ columns?: number }> = ({ columns = 1, children }) => (
   <Box display="grid" gap={4} gridTemplateColumns={`repeat(${columns}, 1fr)`}>
@@ -158,9 +157,6 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
   >({});
 
   const isEmptyPartialConfig = Object.keys(partialConfig).length === 0;
-
-  const isCustomPostprocessorClassName = (class_name: string) =>
-    !POSTPROCESSORS_CLASS_NAMES.includes(class_name as PostprocessorClassName);
 
   const handleDiscard = () => {
     setPartialConfig({});
@@ -337,13 +333,6 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
       </Box>
     );
   }
-
-  const parseClassName = (class_name: string) =>
-    class_name && class_name.toLowerCase().includes("temperature")
-      ? "temperature"
-      : class_name.toLowerCase().includes("threshold")
-      ? "threshold"
-      : undefined;
 
   const updateSubConfig = <Key extends SubConfigKeys>(
     key: Key,
@@ -653,49 +642,35 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
                     <Columns columns={2}>
                       <AutocompleteStringField
                         label="class_name"
-                        options={POSTPROCESSORS_CLASS_NAMES}
+                        options={Object.keys(KNOWN_POSTPROCESSORS)}
                         value={postprocessor.class_name}
                         autoFocus
                         disabled={
-                          resultingConfig.pipelines![pipelineIndex]
-                            .postprocessors === null || isUpdatingConfig
+                          isUpdatingConfig || pipeline.postprocessors === null
                         }
-                        onChange={(class_name) => {
-                          const previousClassName = parseClassName(
-                            postprocessor.class_name
-                          );
-                          const updatedClassName = parseClassName(class_name);
-                          const defaultPostProcessorValue =
-                            (updatedClassName === "temperature" && 1) ||
-                            (updatedClassName === "threshold" && 0.5);
-                          updatePostprocessor(pipelineIndex, index, {
-                            class_name,
-                            ...(!isCustomPostprocessorClassName(class_name) &&
-                              updatedClassName && {
-                                [updatedClassName]: defaultPostProcessorValue,
-                                kwargs: {
-                                  [updatedClassName]: defaultPostProcessorValue,
-                                },
-                              }),
-                            ...(isCustomPostprocessorClassName(class_name) &&
-                              previousClassName && {
-                                kwargs: {
-                                  [previousClassName]: undefined,
-                                },
-                                [previousClassName]: undefined,
-                              }),
-                            ...(!isCustomPostprocessorClassName(
-                              postprocessor.class_name
-                            ) &&
-                              previousClassName && {
-                                [previousClassName]: undefined,
-                              }),
-                          });
-                        }}
+                        onChange={(class_name) =>
+                          updatePipeline(pipelineIndex, {
+                            postprocessors: splicedArray(
+                              pipeline.postprocessors!,
+                              index,
+                              1,
+                              {
+                                args: [],
+                                kwargs: {},
+                                remote: null,
+                                ...(KNOWN_POSTPROCESSORS[
+                                  class_name as keyof typeof KNOWN_POSTPROCESSORS
+                                ] ||
+                                  postprocessor.class_name in
+                                    KNOWN_POSTPROCESSORS || // true spreads nothing
+                                  postprocessor),
+                                class_name,
+                              }
+                            ),
+                          })
+                        }
                       />
-                      {isCustomPostprocessorClassName(
-                        postprocessor.class_name
-                      ) ? (
+                      {!(postprocessor.class_name in KNOWN_POSTPROCESSORS) && (
                         <>
                           <StringField
                             label="remote"
@@ -730,49 +705,38 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
                             }
                           />
                         </>
-                      ) : (
-                        (postprocessor.class_name?.includes("Temperature") && (
-                          <NumberField
-                            label="temperature"
-                            value={
-                              "temperature" in postprocessor
-                                ? postprocessor.temperature
-                                : 1
-                            }
-                            disabled={
-                              resultingConfig.pipelines![pipelineIndex]
-                                .postprocessors === null || isUpdatingConfig
-                            }
-                            onChange={(temperature) =>
-                              updatePostprocessor(pipelineIndex, index, {
-                                temperature,
-                                kwargs: { temperature },
-                              })
-                            }
-                            {...FLOAT}
-                          />
-                        )) ||
-                        (postprocessor.class_name?.includes("Threshold") && (
-                          <NumberField
-                            label="threshold"
-                            value={
-                              "threshold" in postprocessor
-                                ? postprocessor.threshold
-                                : 0.5
-                            }
-                            disabled={
-                              resultingConfig.pipelines![pipelineIndex]
-                                .postprocessors === null || isUpdatingConfig
-                            }
-                            onChange={(threshold) =>
-                              updatePostprocessor(pipelineIndex, index, {
-                                threshold,
-                                kwargs: { threshold },
-                              })
-                            }
-                            {...PERCENTAGE}
-                          />
-                        ))
+                      )}
+                      {"temperature" in postprocessor && (
+                        <NumberField
+                          label="temperature"
+                          value={postprocessor.temperature}
+                          disabled={
+                            isUpdatingConfig || pipeline.postprocessors === null
+                          }
+                          onChange={(temperature) =>
+                            updatePostprocessor(pipelineIndex, index, {
+                              temperature,
+                              kwargs: { temperature },
+                            })
+                          }
+                          {...FLOAT}
+                        />
+                      )}
+                      {"threshold" in postprocessor && (
+                        <NumberField
+                          label="threshold"
+                          value={postprocessor.threshold}
+                          disabled={
+                            isUpdatingConfig || pipeline.postprocessors === null
+                          }
+                          onChange={(threshold) =>
+                            updatePostprocessor(pipelineIndex, index, {
+                              threshold,
+                              kwargs: { threshold },
+                            })
+                          }
+                          {...PERCENTAGE}
+                        />
                       )}
                     </Columns>
                   </FormGroup>
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -640,7 +640,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
                 onChange={(postprocessors) =>
                   updatePipeline(pipelineIndex, { postprocessors })
                 }
-                renderItem={(postprocessor, index) => (
+                renderItem={(postprocessor, index, postprocessors) => (
                   <FormGroup sx={{ marginTop: 2 }}>
                     <Columns columns={2}>
                       <AutocompleteStringField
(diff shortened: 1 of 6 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -110,6 +125,21 @@ const Columns: React.FC<{ columns?: number }> = ({ columns = 1, children }) => (
   </Box>
 );
 
+const CustomValueField: React.FC<
```
JosephMarinier, line 128: Two things:
1. The name. Most fields are for "custom values", maybe with the only exception of the select menus. What is special about this one is quite the opposite: it suggests/autocompletes some predefined values (the opposite of custom). How about `AutocompleteStringField`
2. Could you please move this component to its own file, like `JSONField`, `NumberField`, `StringArrayField` and `StringField`?

webapp/src/pages/Settings.tsx
```diff
@@ -92,6 +102,11 @@ const SUPPORTED_SPACY_MODELS: SupportedSpacyModels[] = [
   "fr_core_news_md",
 ];
 const USE_CUDA_OPTIONS = ["auto", "true", "false"] as const;
+const POSTPROCESSORS_CLASS_NAMES: string[] = [
```
JosephMarinier, line 105: I would suggest making the type more robust in case the values need to change:
```suggestion
type PostprocessorClassName = (TemperatureScaling | ThresholdConfig)["class_name"];
const POSTPROCESSORS_CLASS_NAMES: PostprocessorClassName[] = [
```
with the types in `src/types/api.ts` of course:
```ts
type TemperatureScaling = components["schemas"]["TemperatureScaling"];
type ThresholdConfig = components["schemas"]["ThresholdConfig"];
```

webapp/src/pages/Settings.tsx
```diff
@@ -1,5 +1,11 @@
-import { Close, Warning } from "@mui/icons-material";
 import {
+  AddCircleOutline,
```
JosephMarinier, line 2: We normally use the filled variants of the icons. I think it would be best, especially to align with the `DeleteForever` icon which is filled.
```suggestion
  AddCircle,
```

webapp/src/pages/Settings.tsx
```diff
@@ -362,6 +395,64 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+    postprocessor: PipelineDefinition["postprocessors"]
+  ) => (
+    <Divider textAlign="left" sx={{ position: "relative", height: 2 }}>
+      <Tooltip title="Add postprocessor">
+        <IconButton
+          sx={{
+            bottom: 25,
+            marginTop: 2,
+            padding: 0,
+          }}
```
JosephMarinier, line 437: I notice that if you remove the `position` and the `height` on the `Divider`, as well as the `bottom`, the `marginTop`, and the `padding` on the `IconButton`, the button centers properly. I think it looks good with the `size="small"`.
```suggestion
    <Divider textAlign="left">
      <Tooltip title="Add postprocessor">
        <IconButton
          size="small"
```

webapp/src/pages/Settings.tsx
```diff
@@ -619,69 +735,192 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
                           )}
-                        </Columns>
-                      </Paper>
-                    ))}
-                  </FormGroup>
+                        {(
+                          pipeline.postprocessors ??
+                          defaultConfig.pipelines![0].postprocessors
+                        )?.map((postprocessor, index) => (
+                          <>
```
JosephMarinier, line 764: > Warning: Each child in a list should have a unique "key" prop.

webapp/src/pages/Settings.tsx
nandhinibsn: Yes, the suggested name is suitable for the scenario. I can move it to a separate file.

webapp/src/pages/Settings.tsx
```diff
@@ -619,69 +735,192 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
                       />
                     </Columns>
                   </FormGroup>
-                </FormControl>
```
JosephMarinier, line 622: I think you accidentally moved the following `FormControl` inside this `FormControl`. It should go next to it instead.

webapp/src/pages/Settings.tsx
JosephMarinier: I prefer reducing the gap between the divider and the button by styling the divider component, with `[`& .${dividerClasses.wrapper}`]: { padding: 0 }`. We can also better align the button to the left (instead of the `textAlign="left"` that looked weird):
```suggestion
    <Divider
      sx={{
        "&::before": { width: (theme) => theme.spacing(2) },
        [`& .${dividerClasses.wrapper}`]: { padding: 0 },
      }}
    >
      <Tooltip title="Add postprocessor">
        <IconButton
```

webapp/src/pages/Settings.tsx
```diff
@@ -558,14 +649,39 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
               >
                 <FormControl>
-                  {displaySectionTitle("General")}
+                  <Box
+                    display="flex"
+                    justifyContent="space-between"
+                    alignItems="center"
+                  >
+                    {displaySectionTitle("General")}
+                    <IconButton
```
JosephMarinier, line 665: For the new icon buttons, I like the color `sx={{ color: (theme) => theme.palette.grey[400] }}`. What do you think?

webapp/src/pages/Settings.tsx
JosephMarinier: Note that I said
> instead of the `textAlign="left"` that looked weird

You can remove the `textAlign="left"` now.

webapp/src/pages/Settings.tsx
JosephMarinier: Instead of a `<Box key={index}>`, you can use a `<React.Fragment key={index}>` to avoid the unnecessary div once rendered.

webapp/src/pages/Settings.tsx
JosephMarinier: Don't forget the Icon buttons to delete a pipeline/postprocessor.

webapp/src/pages/Settings.tsx
JosephMarinier: Oh, no, even simpler: https://github.com/ServiceNow/azimuth/pull/561/files#r1185347695

webapp/src/pages/Settings.tsx
```diff
@@ -557,133 +637,292 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+            variant="outlined"
+            sx={{ display: "flex", flexDirection: "column", padding: 2 }}
+          >
+            <FormGroup sx={{ gap: 2 }}>
+              {displayPipelineAddSection(-1)}
+              {resultingConfig.pipelines.map((pipeline, pipelineIndex) => (
+                <Box
+                  key={pipelineIndex}
+                  sx={{ display: "flex", flexDirection: "column" }}
+                >
```
JosephMarinier, line 650:
```suggestion
          <Paper variant="outlined">
            {displayPipelineAddSection(-1)}
            {resultingConfig.pipelines.map((pipeline, pipelineIndex) => (
              <React.Fragment key={pipelineIndex}>
                <FormGroup>
```

webapp/src/pages/Settings.tsx
```diff
@@ -557,133 +637,292 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
-                </FormControl>
-              </Paper>
-            ))}
-          </FormGroup>
+                    </FormControl>
+                  </FormControl>
+                  {displayPipelineAddSection(pipelineIndex)}
+                </Box>
+              ))}
+            </FormGroup>
```
JosephMarinier, line 924:
```suggestion
                </FormGroup>
                {displayPipelineAddSection(pipelineIndex)}
              </React.Fragment>
            ))}
```

JosephMarinier (commented):
I suggest moving the `<FormGroup>` inside the `pipelines.map()` so the dividers can span all the way to the `Paper` borders. The same thing would apply to the postprocessors.

webapp/src/pages/Settings.tsx
```diff
@@ -557,133 +637,292 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                            displayPostprocessorAddSection(
+                              pipelineIndex,
+                              -1,
+                              pipeline.postprocessors
+                            )}
+                          {(
+                            pipeline.postprocessors ??
+                            defaultConfig.pipelines![0].postprocessors
+                          )?.map((postprocessor, index) => (
+                            <Box key={index}>
```
JosephMarinier, line 760:
```suggestion
                      <Paper variant="outlined">
                        {pipeline.postprocessors &&
                          displayPostprocessorAddSection(
                            pipelineIndex,
                            -1,
                            pipeline.postprocessors
                          )}
                        {(
                          pipeline.postprocessors ??
                          defaultConfig.pipelines![0].postprocessors
                        )?.map((postprocessor, index) => (
                          <React.Fragment key={index}>
                            <FormGroup sx={{ marginTop: 2 }}>
```

webapp/src/pages/Settings.tsx
```diff
@@ -557,133 +637,292 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                              </Columns>
+                              {pipeline.postprocessors &&
+                                displayPostprocessorAddSection(
+                                  pipelineIndex,
+                                  index,
+                                  pipeline.postprocessors
+                                )}
+                            </Box>
+                          ))}
+                        </FormGroup>
```
JosephMarinier, line 917:
```suggestion
                            </FormGroup>
                            {pipeline.postprocessors &&
                              displayPostprocessorAddSection(
                                pipelineIndex,
                                index,
                                pipeline.postprocessors
                              )}
                          </React.Fragment>
                        ))}
```

JosephMarinier (commented):
Same thing for postprocessors.

webapp/src/pages/Settings.tsx
```diff
@@ -557,133 +637,292 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                                ) : (
+                                  (postprocessor.class_name?.includes(
+                                    "Temperature"
+                                  ) && (
+                                    <NumberField
+                                      label="temperature"
+                                      value={
+                                        "temperature" in postprocessor
+                                          ? postprocessor.temperature
+                                          : 0.5
```
JosephMarinier, line 835: 3 things:
1. That should be equivalent to
```suggestion
                                        postprocessor.temperature ?? 0.5
```
2. I think 1 would be a better default for temperature, as it is equivalent to not applying any temperature scaling. Starting from there, the user would choose if it should go up or down.
3. If you change from a "known" postprocessor to another, the `temperature` and `threshold` don't get cleaned up, so you end up with an invalid config that fails:
```json
{
    "class_name": "azimuth.utils.ml.postprocessing.TemperatureScaling",
    "args": [],
    "kwargs": {
        "temperature": 0.6
    },
    "remote": null,
    "threshold": 0.5,
    "temperature": 0.6
}
```

webapp/src/pages/Settings.tsx
```diff
@@ -362,6 +381,67 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+      <Tooltip title="Add postprocessor">
+        <IconButton
+          sx={{ color: (theme) => theme.palette.grey[400] }}
+          aria-label="add-postprocessor"
+          disabled={isUpdatingConfig}
+          onClick={() =>
+            updatePipeline(pipelineIndex, {
+              postprocessors: [
+                ...postprocessor!.slice(0, postprocessorIndex + 1),
+                ...defaultConfig.pipelines![0].postprocessors!,
```
JosephMarinier, line 433: It doesn't make sense to add multiple times the same postprocessor class (in this case `Threshold`), so I would suggest adding an empty one and letting the user choose the class.
```suggestion
                { class_name: "", args: [], kwargs: {}, remote: null },
```

webapp/src/pages/Settings.tsx
```diff
@@ -362,6 +381,67 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+      <Tooltip title="Add pipeline">
+        <IconButton
+          sx={{ color: (theme) => theme.palette.grey[400] }}
+          aria-label="add-pipeline"
+          disabled={isUpdatingConfig}
+          onClick={() =>
+            updatePartialConfig({
+              pipelines: [
+                ...resultingConfig.pipelines!.slice(0, pipelineIndex + 1),
+                ...defaultConfig.pipelines!,
```
JosephMarinier, line 400: This seems dangerous. If one day we change the `defaultConfig` to have two pipelines, the button with start adding pairs if pipelines instead of just one. I suggest:
```suggestion
                defaultConfig.pipelines![0],
```

webapp/src/pages/Settings.tsx
```diff
@@ -557,133 +637,292 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                                    />
+                                  ))
+                                )}
+                                <IconButton
+                                  sx={{
+                                    position: "absolute",
+                                    right: 20,
+                                  }}
+                                  aria-label="delete"
+                                  disabled={isUpdatingConfig}
```
JosephMarinier, line 892: The delete button should be disabled if `pipeline.postprocessors` in `null` (and we only see the default values). Right now, if we click on it, we get an
> Uncaught TypeError: Cannot read properties of null (reading 'slice')

webapp/src/pages/Settings.tsx
```diff
@@ -324,6 +343,13 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     );
   }
 
+  const parseClassName = (class_name: string) =>
+    class_name.toLowerCase().includes("temperature")
```
JosephMarinier, line 347: I see some
```
caught TypeError: Cannot read properties of null (reading 'toLowerCase')
    at parseClassName (Settings.tsx:347:1)
```

webapp/src/pages/Settings.tsx
JosephMarinier: Your solution doesn't seem to work. I am not able to define a custom `class_name` anymore (with `args`, `kwargs`, etc.).

nandhinibsn: Can you give please some examples to define the class name with a custom value? I can figure out the issue behind it.

JosephMarinier: I don't have a concrete example, but if I type "potato" as the `class_name`, I expect to get the `args`, `kwargs` and `remote` fields.

JosephMarinier: Oh! I see that it works, I just have to hit enter on the `class_name` field. That was not obvious to me. Let's think of a solution...

nandhinibsn: Okay, Yes! Thank you for confirming it. I am figuring out to skip enter key.

JosephMarinier: I found that by adding `autoSelect`, we don't need to hit enter, just to focus out of the field. 👌 
With that, I found that setting `isOptionEqualToValue={() => false}` helps make sure there isn't an option that is highlighted by accident (which would get unexpectedly selected when focusing out).
Also, I would add `disableClearable` to remove the `x` button as it doesn't seem necessary.

nandhinibsn: Thank you for the suggestion :)

webapp/src/pages/Settings.tsx
```diff
@@ -560,133 +646,307 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                          pipelineIndex,
+                          -1,
+                          pipeline.postprocessors
+                        )}
+                      {(
+                        pipeline.postprocessors ??
+                        defaultConfig.pipelines![0].postprocessors
+                      )?.map((postprocessor, index) => (
+                        <React.Fragment key={index}>
+                          <IconButton
```
JosephMarinier, line 748: Now that you moved the delete button definition up (and use `position: "absolute"`), if you define a custom metric, the remote field gets in front of the button, making clicking it difficult. The delete button should be defined after (bellow) the fields.

webapp/src/pages/Settings.tsx
```diff
@@ -560,133 +646,307 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+              <React.Fragment key={pipelineIndex}>
+                <FormGroup>
+                  <FormControl>
+                    <Box
+                      display="flex"
+                      justifyContent="space-between"
+                      alignItems="center"
+                    >
+                      {displaySectionTitle("General")}
+                      <IconButton
```
JosephMarinier, line 661: Would it be possible to position the delete button the same way for the pipelines as it is for the postprocessors? So it's uniform.

nandhinibsn:
I did some styling on the delete button as per your comment. Let me know what do you think of it.

image

webapp/src/pages/Settings.tsx
JosephMarinier: > I did some styling on the delete button as per your comment.

I still see a bigger gap around the pipeline delete button than there is around the postprocessor delete button.
239018517-6ff8fc5b-20d1-4609-b664-ee78f6fcc0e3

webapp/src/pages/Settings.tsx
```diff
@@ -560,133 +646,312 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                      {(
+                        pipeline.postprocessors ??
+                        defaultConfig.pipelines![0].postprocessors
+                      )?.map((postprocessor, index) => (
+                        <React.Fragment key={index}>
+                          <Box
+                            display="flex"
+                            justifyContent="flex-end"
+                            alignItems="center"
+                            padding={1}
```
JosephMarinier, line 752: Because of this new padding, the spacings in blue are no longer equal.

239018517-6ff8fc5b-20d1-4609-b664-ee78f6fcc0e3

nandhinibsn: Space and button made it sync now. Let me know your feedback from below.

image

webapp/src/pages/Settings.tsx
```diff
@@ -560,133 +646,313 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+                              <DeleteForever fontSize="large" />
+                            </IconButton>
+                          </Box>
+                          <FormGroup sx={{ marginTop: 2 }}>
+                            <Columns columns={2}>
+                              <AutocompleteStringField
+                                label="class_name"
+                                options={POSTPROCESSORS_CLASS_NAMES}
+                                value={postprocessor.class_name}
+                                autoFocus
```
JosephMarinier, line 787: @nandhinibsn I liked your idea from your other PR of adding `autoFocus`, so I added it here too. 👍

JosephMarinier (approved):

PR #569, opened 2023-05-31
Support custom metrics

JosephMarinier:
## Description:

This PR adds support for customizing, adding, and removing some metrics.

This is also introducing the functionality of disabling the `Apply and close` button if the names of the metrics are duplicated or empty. It will be possible to extend this functionality to other field validations, but for the sake of this PR, that's it.

One question remains: should we move the metrics to a separate section to keep the Model section small? The cons I see with the separate section are:
1. In the new section, say "Metrics", the "box" would need a sub-title to be uniform with the rest, and repeating "Metrics" looks bad. We would have to come up with a solution.
2. If I have no pipeline, I'm not interested in the Metrics sections, so making them look separate/independent might be more misleading.

Screenshot 2023-05-31 at 11 25 46 AM

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -51,6 +51,10 @@ import {
 import { PickByValue } from "types/models";
 import { UNKNOWN_ERROR } from "utils/const";
 
+type MetricState = MetricDefinition & { name: string };
+
+type ConfigState = Omit<AzimuthConfig, "metrics"> & { metrics: MetricState[] };
+
 const CONFIG_UPDATE_MESSAGE =
   "Please wait while the config changes are validated.";
 const PERCENTAGE = { scale: 100, units: "%", inputProps: { min: 0, max: 100 } };
@@ -79,7 +83,7 @@ const FIELDS: Record<
 };
 
 type SubConfigKeys = keyof PickByValue<
-  AzimuthConfig,
+  ConfigState,
   { [key: string]: unknown } | null
 >;
 
@@ -97,7 +101,7 @@ const SUPPORTED_SPACY_MODELS: SupportedSpacyModels[] = [
 const USE_CUDA_OPTIONS = ["auto", "true", "false"] as const;
 type UseCUDAOption = typeof USE_CUDA_OPTIONS[number];
 
-const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof AzimuthConfig)[] = [
+const FIELDS_TRIGGERING_STARTUP_TASKS: (keyof ConfigState)[] = [
   "behavioral_testing",
   "similarity",
   "dataset_warnings",
@@ -115,8 +119,6 @@ const KNOWN_POSTPROCESSORS: {
   "azimuth.utils.ml.postprocessing.TemperatureScaling": { temperature: 1 },
   "azimuth.utils.ml.postprocessing.Thresholding": { threshold: 0.5 },
 };
-
-type Metric = MetricDefinition & { name: string };
 
 const isErrorMetrics = (metrics: Metric[]) => {
   const names = new Set(metrics.map(({ name }) => name));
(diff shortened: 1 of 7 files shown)


gabegma (commented):
LGTM for the UX!! Awesome work :)

webapp/src/pages/Settings.tsx
```diff
@@ -138,6 +145,44 @@ const KeyValuePairs: React.FC = ({ children }) => (
 const updateArrayAt = <T,>(array: T[], index: number, update: Partial<T>) =>
   splicedArray(array, index, 1, { ...array[index], ...update });
 
+const CustomObjectFields: React.FC<{
```
JosephMarinier, line 148: @christyler3030

> Are there any opportunities for factoring out some sub components, utility functions, or constants?  The new `CustomObjectFields` component is an obvious candidate. Are there others? Thoughts?

For sure.

This `CustomObjectFields` is the only component outside of `Settings` that is more than 3 lines long, so I don't see other trivial moves. It would require some refactoring to move more code out of `Settings`. I kept it here because I can't imagine any case to reuse it, but I'm happy to move it anyway. The `components/Settings/` folder is perfect for that. 👍

christyler3030: Up to you. Just an observation.

webapp/src/pages/Settings.tsx
```diff
@@ -168,7 +213,10 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     onClose();
   };
 
-  const resultingConfig = Object.assign({}, config, partialConfig);
+  const resultingConfig = React.useMemo(
+    () => Object.assign({}, config, partialConfig),
+    [config, partialConfig]
```
christyler3030, line 218: Given that config and partialConfig are objects, are you sure useMemo is doing anything? (I can't quite remember how this works but my intuition is that this will change every time)

webapp/src/pages/Settings.tsx
```diff
@@ -117,6 +116,14 @@ const KNOWN_POSTPROCESSORS: {
   "azimuth.utils.ml.postprocessing.Thresholding": { threshold: 0.5 },
 };
 
+type Metric = MetricDefinition & { name: string };
+
+const isErrorMetrics = (metrics: Metric[]) => {
```
christyler3030, line 121: It makes sense once you read the code but I was initially confused by the name of this function. What about something like `metricsAreValid` or something if there are plans to extend the validation later or something like `metricNamesAreValid` if not?

webapp/src/pages/Settings.tsx
JosephMarinier: So, `config` and `partialConfig` are objects, but they are memoized themselves, so if they don't "change", the same object reference is received and it is `===`. For example, `config` comes from RTK Query and is constant once it is loaded, so it stays `===` on every render.
In practice, this doesn't make an appreciable difference because `partialConfig` basically holds the state for everything in the page, so it does change (almost?) all the time. I thought it didn't hurt, and it's ready for if other things start holding state.
Does that make sense?

christyler3030: Makes sense.

webapp/src/pages/Settings.tsx
JosephMarinier: Agree!
Yes, it already handles both metrics' names and metrics' `class_name`s. I would like it to be generic to any errors one day, so I'll call it `hasErrors`. I'll add a check for if `dataset.class_name` is empty to make the point, which is useful if you open an "empty" Azimuth for the first time, as we discussed in #558.
Does that make sense? I think it looks straightforward:
```ts
disabled={isEmptyPartialConfig || isUpdatingConfig || hasErrors}
```

christyler3030 (approved):

PR #579, opened 2023-06-16
Add a route to validate a config

JosephMarinier:
## Description:

* [Clean up](https://github.com/ServiceNow/azimuth/pull/579/commits/d56cb1bcc8c558d3a1bad416cbca559ea50728c7)
* [Fix updating config with equivalent `artifact_path`](https://github.com/ServiceNow/azimuth/pull/579/commits/9e40e507612a39d2eaf75635e7221f3963902a19) (for example relative vs absolute)
  - Have Pydantic parse `artifact_path` as an absolute path.
  - Have only Pydantic deal with the `partial_config: Dict`, and check permission to update the config on the resulting updated config (instead of on the `partial_config: Dict`).
  - As a result, I can update from a relative `artifact_path="cache"` to an absolute `artifact_path="/Users/joseph/azimuth/cache"` for example, or even weird stuff like `artifact_path="/Users/joseph/azimuth/cache/../cache"`.
* [Add a route to validate a config](https://github.com/ServiceNow/azimuth/pull/579/commits/9019fd24d2decf4c835654b8624f3d67f6082503) without applying it

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/azimuth/config.py
+++ b/azimuth/config.py
@@ -284,10 +284,14 @@ class ProjectConfig(AzimuthBaseSettings):
 
 class ArtifactsConfig(AzimuthBaseSettings, extra=Extra.ignore):
     artifact_path: str = Field(
-        "cache",
+        default_factory=lambda: os.path.abspath("cache"),
         description="Where to store artifacts (Azimuth config history, HDF5 files, HF datasets).",
         exclude_from_cache=True,
     )
+
+    @validator("artifact_path")
+    def validate_artifact_path(cls, artifact_path):
+        return os.path.abspath(artifact_path)
 
     def get_config_history_path(self):
         return f"{self.artifact_path}/config_history.jsonl"
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -8,6 +8,7 @@ from jsonlines import jsonlines
 from pydantic import ValidationError
 
 from azimuth.config import (
+    ArtifactsConfig,
     AzimuthConfig,
     AzimuthConfigHistoryWithHash,
     PipelineDefinition,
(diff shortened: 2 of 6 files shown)


azimuth/config.py
```diff
@@ -284,11 +284,15 @@ def get_project_hash(self):
 
 class ArtifactsConfig(AzimuthBaseSettings, extra=Extra.ignore):
     artifact_path: str = Field(
-        "cache",
+        default_factory=lambda: os.path.abspath("cache"),
```
gabegma, line 287: I think we need to edit some of our documentation here, at the very least, the default value in the reference.

azimuth/config.py
```diff
@@ -284,11 +284,15 @@ def get_project_hash(self):
...
         description="Where to store artifacts (Azimuth config history, HDF5 files, HF datasets).",
         exclude_from_cache=True,
     )
 
+    @validator("artifact_path")
+    def validate_artifact_path(cls, artifact_path):
+        return os.path.abspath(artifact_path)
```
gabegma, line 294: Does that mean that our user still needs to provide the relative path? Maybe we don't need to edit the documentation then!

tests/test_config.py
```diff
@@ -289,3 +290,10 @@ def test_config_history_with_hash():
         ValidationError, match="1 validation error for AzimuthConfigHistoryWithHash\nconfig -> name"
     ):
         AzimuthConfigHistoryWithHash(config={"name": None})
+
+
+def test_artifact_path_equality():
+    # This is important since we forbid updating the config if the artifact_path differs.
+    default = ArtifactsConfig()
+    assert ArtifactsConfig(artifact_path="cache/../cache") == default
```
gabegma, line 298: What does `cache/../cache` mean?

gabegma (approved):
LGTM! :)

tests/test_config.py
JosephMarinier: The same as `cache`. You know how `..` means the parent directory, like in `cd ..`? So `cache/../cache` is just to test that it gets correctly resolved as the same as `cache`.

azimuth/config.py
JosephMarinier: Users can specify any relative or absolute path, like before. The only difference is that we resolve it as an absolute path to save it, so the representation is unique. I don't think we have to change anything in the documentation.

gabegma: Yes, ok! Thanks!

JosephMarinier: Thank you! 👍

PR #580, opened 2023-06-16
Fix importing the same proposed actions CSV file twice

### LINKED ISSUE #479: Fix support for importing the same CSV file twice
Bug: When importing a CSV file of proposed actions, importing the same file path twice fails silently. The file browser closes normally but the `fileReader.onload` is never called.
Use case: One might have edited the file and wants to re-import it.
Workaround: Refresh the app.

JosephMarinier:
Resolve #479

## Description:

* [Create reusable component FileInputButton](https://github.com/ServiceNow/azimuth/pull/580/commits/42509e900e1dceaa4dcd891d0be72f5b2859303b) (I will be reusing it in a future PR)
* [Fix uploading same file twice](https://github.com/ServiceNow/azimuth/pull/580/commits/bb26895760eab31a7c1ad6003d74d3ccdd0c9d01)
* [Log changes](https://github.com/ServiceNow/azimuth/pull/580/commits/9dcdc5c0710a8ce4eb95b666984b9eb9576fdd92)

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Analysis/UtterancesTable.tsx
+++ b/webapp/src/components/Analysis/UtterancesTable.tsx
@@ -33,6 +33,7 @@ import CopyButton from "components/CopyButton";
 import Description from "components/Description";
 import OutcomeIcon from "components/Icons/OutcomeIcon";
 import TargetIcon from "components/Icons/Target";
+import FileInputButton from "components/FileInputButton";
 import Loading from "components/Loading";
 import SmartTagFamilyBadge from "components/SmartTagFamilyBadge";
 import { Column, RowProps, Table } from "components/Table";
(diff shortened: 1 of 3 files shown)


webapp/src/components/Analysis/UtterancesTable.tsx
```diff
@@ -436,35 +437,28 @@ const UtterancesTable: React.FC<Props> = ({
...
-        updateDataAction({
-          ignoreNotFound: true,
-          body,
-          ...getUtterancesQueryState,
-        });
-      }
-    };
-    fileReader.readAsText(file);
+    const body = rows.map((row) => {
+      const [persistentId, dataAction] = row.split(",");
```
christyler3030, line 454: Is there any other validations necessary here?

christyler3030 (approved):

webapp/src/components/Analysis/UtterancesTable.tsx
JosephMarinier: (Just to be sure we're on the same page, I did not change this in this PR.)
Here in the front end, we validate that the CSV headers are exactly as expected, otherwise, we wouldn't know how to parse the file and what to send to the back end. But it stops there.
I think it's fine to let the back end validate the actual values.
For example, we could validate in the front end that the `dataAction` is one of the valid values, but the back end already does it. 🤷
Does that make sense?

PR #581, opened 2023-06-16
Add buttons to upload and download a config file

JosephMarinier:
## Description:

* [Create reusable conversion functions between `AzimuthConfig` and `ConfigState`](https://github.com/ServiceNow/azimuth/pull/581/commits/a15eadd0241b897c013acd50ba8f3ca508f202a4) 
* [Add missing config fields that may trigger time-consuming computations](https://github.com/ServiceNow/azimuth/pull/581/commits/6347aaf97099c97348a1030e5b7270ad12a73df7)
* [Fix download icon semantic](https://github.com/ServiceNow/azimuth/pull/581/commits/7d4a082a8e54cbfcefd6abcca911275b4d330d52) by using the identical `Download` icon instead of `GetApp`.
* [Add button to upload a config file](https://github.com/ServiceNow/azimuth/pull/581/commits/e860569a80d523dd05b0914ca41765da22da5404)
* [Add button to download a config file](https://github.com/ServiceNow/azimuth/pull/581/commits/ad37eab037f4a452011bea7cbb4bc702091357b5)

image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -278,7 +278,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
       validateConfig({ jobId, body })
         .unwrap()
         .then((config) => setPartialConfig(azimuthConfigToConfigState(config)))
-        .catch(() => {}); // Avoid the uncaught error log.
+        .catch(() => {}); // Avoid the uncaught error log. Toast already raised by `rtkQueryErrorInterceptor` middleware.
     } catch (error) {
       raiseErrorToast(
         `Something went wrong parsing JSON file\n${
@@ -404,7 +404,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
             })
               .unwrap()
               .then(handleClose)
-              .catch(() => {}); // Avoid the uncaught error log.
+              .catch(() => {}); // Avoid the uncaught error log. Toast already raised by `rtkQueryErrorInterceptor` middleware.
           }}
         >
           Apply and close
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -55,6 +55,24 @@ type MetricState = MetricDefinition & { name: string };
 
 type ConfigState = Omit<AzimuthConfig, "metrics"> & { metrics: MetricState[] };
 
+const azimuthConfigToConfigState = ({
+  metrics,
+  ...rest
+}: AzimuthConfig): ConfigState => ({
+  ...rest,
+  metrics: Object.entries(metrics).map(([name, m]) => ({ name, ...m })),
+});
+
+const configStateToAzimuthConfig = ({
+  metrics,
+  ...rest
+}: Partial<ConfigState>): Partial<AzimuthConfig> => ({
+  ...rest,
+  ...(metrics && {
+    metrics: Object.fromEntries(metrics.map(({ name, ...m }) => [name, m])),
+  }),
+});
+
 const CONFIG_UPDATE_MESSAGE =
   "Please wait while the config changes are validated.";
 const PERCENTAGE = { scale: 100, units: "%", inputProps: { min: 0, max: 100 } };
@@ -152,14 +170,10 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     SupportedLanguage | undefined
   >();
   const { data: azimuthConfig } = getConfigEndpoint.useQuery({ jobId });
-  const config = React.useMemo(() => {
-    if (azimuthConfig === undefined) return undefined;
-    const { metrics, ...rest } = azimuthConfig;
-    return {
-      metrics: Object.entries(metrics).map(([name, m]) => ({ name, ...m })),
-      ...rest,
-    };
-  }, [azimuthConfig]);
+  const config = React.useMemo(
+    () => azimuthConfig && azimuthConfigToConfigState(azimuthConfig),
+    [azimuthConfig]
+  );
 
   const [updateConfig, { isLoading: isUpdatingConfig }] =
     updateConfigEndpoint.useMutation();
@@ -335,12 +349,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
           onClick={() => {
             updateConfig({
               jobId,
-              body: {
-                ...partialConfig,
-                metrics: Object.fromEntries(
-                  resultingConfig.metrics.map(({ name, ...m }) => [name, m])
-                ),
-              },
+              body: configStateToAzimuthConfig(partialConfig),
             })
               .unwrap()
               .then(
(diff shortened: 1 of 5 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -253,14 +275,38 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
       <DialogTitle id="config-dialog-title">
         <Box display="flex" justifyContent="space-between" alignItems="center">
           <Typography variant="subtitle1">
-            View and edit certain fields from your config file. Once your
-            changes are saved, expect some delays for recomputing the affected
-            tasks.
+            View and edit your config. Once your changes are saved, expect some
+            delays for recomputing the affected tasks.
```
JosephMarinier, line 279: I shortened that since it's not only "certain fields" anymore, but it's still very long for a title.
image

I think the second sentence is redundant with the "These changes may trigger some time-consuming computations." that we get next to the Apply button when we make changes that trigger recompute. Can I remove it?
```suggestion
            View and edit your config.
```
Or even a simple quick title?
```suggestion
            Config
```

gabegma: Given that now it's also there the first time you launch Azimuth without a config, I would just go with `Configuration'.

webapp/src/pages/Settings.tsx
```diff
@@ -102,12 +123,17 @@ const USE_CUDA_OPTIONS = ["auto", "true", "false"] as const;
...
   "behavioral_testing",
   "similarity",
   "dataset_warnings",
   "syntax",
+  "model_contract",
   "pipelines",
   "uncertainty",
+  "saliency_layer",
   "metrics",
 ];
```
JosephMarinier, line 138: @gabegma, could you double-check that with me? I added some config fields that were missing from this list of fields that may trigger the startup tasks.

webapp/src/pages/Settings.tsx
```diff
@@ -253,14 +280,37 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
+              } catch (error) {
+                raiseErrorToast(
+                  `Something went wrong parsing JSON file\n${
+                    (error as SyntaxError).message
+                  }`
+                );
+              }
+            }}
+          >
+            Import JSON config file
```
JosephMarinier, line 308: @christyler3030, @gabegma

I have a hard time deciding on the text (and icon) of the button. Let's consider that button as well as the download button that will come next to it to download the config as a file on your laptop...

The rest of our UI uses "import" and "export". I find that quite confusing.

This button here loads a file in the UI, so you can edit it before `Apply`ing the changes. So it's importing it to the browser, but it's exporting it from your laptop. 😨 

Here, I intuitively wrote "import", but the similar button that reads the proposed actions from a file is called "export". 😵‍💫

Also, the download button that will come next to this one could also be called "import". 😅

Even "uploading" and "downloading" doesn't seem super accurate as the config just moves from a file (locally) to the browser (locally), and vice versa. Not a lot of "up" and "down".

So, is it more like a `Load` and `Save`? I'm afraid `Save` could be confused with the `Apply` button. Would it work if it's more explicit like `Load JSON file` and `Save JSON file`?

I feel like I'm not thinking of something obvious. Do you guys have ideas?

gabegma: Screenshot 2023-06-16 at 12 52 25

> but the similar button that reads the proposed actions from a file is called "export". 😵‍💫

I don't understand this comment - it's called import no?

webapp/src/pages/Settings.tsx
gabegma: This list exists for the purpose of the message that it might take a while? The list seems good - but I would need to see what is left to make sure that nothing is missing. "language" maybe?

webapp/src/pages/Settings.tsx
JosephMarinier: You are totally right! See, I find import/export very confusing.

gabegma: Haha, I don't find it confusing personally. import/export options in a software almost always is from locally in the laptop to the software. I would keep the same terminology.

webapp/src/pages/Settings.tsx
JosephMarinier: Yes.
I hesitated about `language`. It is not necessary to add it because changing the language already changes `syntax`, `similarity`, and `behavioral_testing`. It wouldn't hurt to add it, though. 🤷‍♂️ 
Other than that, the fields that are not here (that are considered safe to edit without triggering the startup tasks) are
- name
- artifact_path
- batch_size
- use_cuda
- large_dask_cluster
- read_only_config

webapp/src/pages/Settings.tsx
JosephMarinier: Right, so the word import/export is in reference to Azimuth. However, in the icons, the rectangle under the arrow seems to be in reference to my laptop. We say "import" while the arrow goes **out** to the rectangle, and "export" while the arrow goes **in**. 😵  I think that's what throws me off.

webapp/src/pages/Settings.tsx
gabegma: Looks good!! Thank you!!

webapp/src/pages/Settings.tsx
gabegma: Right, I see what you mean. Although, I'm so used to the down arrow for export/download and up arrow for import that it's not shocking to me.

JosephMarinier: I totally agree with the arrows. 👍 I'm more used to seeing them for upload/download, though. For example, Google Drive:
image
image
Or Teams:
image
I can't think of an example of a Web app using the import/export vocabulary. Can you?

Anyway, I'll stick with import/export for now, as well as the arrows, and we can come back to this conversation later if we notice confusion with other people.

webapp/src/pages/Settings.tsx
JosephMarinier: Thank you! :)

webapp/src/pages/Settings.tsx
JosephMarinier: Here is what it can look like with the new simpler and slightly bigger `Configuration` title, and the two buttons in the same fashion as the similar buttons in our app:
image

gabegma: > I can't think of an example of a Web app using the import/export vocabulary. Can you?

I believe that tools like Sketchup use it for example.

The new look is perfect!!!

webapp/src/pages/Settings.tsx
```diff
@@ -243,6 +272,29 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     metricsNames.has("") ||
     resultingConfig.metrics.some(({ class_name }) => class_name.trim() === "");
 
+  const handleFileRead = (text: string) => {
+    try {
+      const body = JSON.parse(text);
+      validateConfig({ jobId, body })
+        .unwrap()
+        .then((config) => setPartialConfig(azimuthConfigToConfigState(config)))
+        .catch(() => {}); // Avoid the uncaught error log.
```
christyler3030, line 281: This stops error propagation here right? Do you not want to show a toast?

christyler3030 (approved):

PR #583, opened 2023-06-20
Detect saliency layer

### LINKED ISSUE #582: Auto selection of the saliency layer
Hello!

I recently come around [`model.base_model.get_input_embeddings()`](https://huggingface.co/docs/transformers/main/en/main_classes/model#transformers.PreTrainedModel.get_input_embeddings), a method implemented by Huggingface models to get the `Embedding` layer. I was wondering if we could add an "auto" setting to the saliency layer so that I don't have to look for this layer in my architecture.

### Proposed changes:

```python
# my_config.json
{
...
"saliency_layer": "auto"
...
}
```

```python
class ModelContractConfig:
    saliency_layer: Literal['auto'] | None | str = None
```

gabegma:
Resolve #582 

## Description:
* This sets the default to "auto", so that users no longer need to think about it. LMK what you think @Dref360.

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Released changes are shown in the
 
 ### Added
 - Two buttons in the config UI to import/export a JSON config file.
+- For HF pipelines, the saliency layer is now automatically detected and no longer needs to be specified in the config. To disable saliency maps, the user can still specify `null`.
 
 ### Changed
 


docs/docs/reference/configuration/model_contract.md
```diff
@@ -208,10 +209,11 @@ in [:material-link: Uncertainty Estimation](../../key-concepts/uncertainty.md).
 
 ## Saliency Layer
 
-🟡 **Default value**: `None`
+🟡 **Default value**: `auto`
 
 If using a Pytorch model, [:material-link: Saliency Maps](../../key-concepts/saliency.md) can be
-available. Specify the name of the embedding layer on which to compute them.
+available. Specify the name of the embedding layer on which to compute them. The default is set to "auto".
+In that case, the embedding layer will be found using `model.base_model.get_input_embeddings()`.
```
JosephMarinier, line 216: Detail, but I would suggest
```suggestion
available. Specify the name of the embedding layer on which to compute them. The default is set to "auto",
in which case the embedding layer will be detected automatically using `model.base_model.get_input_embeddings()`.
```

docs/docs/reference/configuration/model_contract.md
```diff
@@ -40,7 +40,8 @@ Fields from this scope defines how Azimuth interacts with the ML pipelines and t
     2. List of pipelines. Can also be set to `null` to launch Azimuth with a dataset only.
     3. Enable uncertainty quantification.
     4. Layer name where to calculate the gradients, normally the word embeddings layer.
-       Only available for Pytorch models.
+       Only available for Pytorch models. Defaults to "auto"; in that case the layer will be found using
+       `model.base_model.get_input_embeddings()`.
```
JosephMarinier, line 44: If that makes sense
```suggestion
       Only available for Pytorch models. Defaults to "auto", in which case the embedding layer will be detected automatically using `model.base_model.get_input_embeddings()`.
```

docs/docs/reference/configuration/model_contract.md
JosephMarinier: So, if I don't have the necessary layer in my model, I now have to specify `"saliency_layer": null`, whereas that was the default before. Is it worth noting the breaking change in the change log? Or would there be a way to also support that use case with the `"auto"`?

gabegma: "auto" will also work when your model is not a HF pipeline (and will be equivalent to "null"). @Dref360, do you foresee a use case where the model would be a HF pipeline, but the `input_embedding` layer wouldn't be appropriate for saliency_maps?

Dref360: Maybe in a bizarre use case with multiple inputs, but this is not a supported use case.

gabegma: Thanks. Then I don't think it's a breaking change, but I added it to the changelog as a new feature. LMK what you think.

JosephMarinier (approved):
Cool! Thank you, and thank you @Dref360 for the idea!

PR #586, opened 2023-06-27
Add drop-down menu to load a previous config

JosephMarinier:
## Description:

I show a hash of each config in a colored chip to easily tell if some configs are duplicates (for example when going back and forth between two configs). The color of the chip is based on the hash. Special cases:
* Don't show the hash (hashSize = null) if no two different configs have the same name. ❓ Maybe that's not desirable?
* Show a 6-char hash if there is a collision in the first 3 chars.❓ Do you think that is overkill? See the probabilities below.
* Otherwise, show a 3-char hash.

I calculated the probability of a hash collision (two different configs producing the same hash)...
* with the 3-char hash:
  - 10 different configs: 1 %
  - 30 different configs: 10 %
  - 76 different configs: 50 %
* with the 6-char hash:
  - 581 different configs: 1 %

image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -282,6 +282,24 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
     metricsNames.has("") ||
     resultingConfig.metrics.some(({ class_name }) => class_name.trim() === "");
 
+  const fullHashCount = new Set(configHistory?.map(({ hash }) => hash)).size;
+  const hashCount = new Set(configHistory?.map(({ hash }) => hash.slice(0, 3)))
+    .size;
+  const nameCount = new Set(configHistory?.map(({ config }) => config.name))
+    .size;
+
+  // Don't show the hash (hashSize = null) if no two different configs have the same name.
+  // Show a 6-char hash if there is a collision in the first 3 chars.
+  // Otherwise, show a 3-char hash.
+  // Probability of a hash collision with the 3-char hash:
+  // 10 different configs: 1 %
+  // 30 different configs: 10 %
+  // 76 different configs: 50 %
+  // With the 6-char hash:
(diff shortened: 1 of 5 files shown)


gabegma:
WHAT!! How cool is that!!!!

christyler3030 (approved):
Neat!  Approved. Question: is it worth keeping multiple copies of configs that have the same hash?  I can see why you would do it but just posing the question.

webapp/src/pages/Settings.tsx
```diff
@@ -272,6 +282,24 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
...
 
+  const fullHashCount = new Set(configHistory?.map(({ hash }) => hash)).size;
+  const hashCount = new Set(configHistory?.map(({ hash }) => hash.slice(0, 3)))
+    .size;
+  const nameCount = new Set(configHistory?.map(({ config }) => config.name))
+    .size;
+
+  // Don't show the hash (hashSize = null) if no two different configs have the same name.
+  // Show a 6-char hash if there is a collision in the first 3 chars.
+  // Otherwise, show a 3-char hash.
```
JosephMarinier, line 293: @christyler3030
> Question: is it worth keeping multiple copies of configs that have the same hash?

Good question! I hesitated. I think showing the full history could help the user figure out what is what. One might remember "I first changed this, then I reverted, then that" and expect a certain number of configs and dates to show up. With some missing, one might get confused.

One solution could be to gray out configs that are repeated later in the history, or something like that, but I couldn't think of something clearly intuitive. The list would begin to get complex.

Other ideas?

JosephMarinier: Cool! Thank you for your review! 🙏

PR #587, opened 2023-06-28
Fix disabled look

JosephMarinier:
## Description:

Three steps:
* [Make `onChange()` required](https://github.com/ServiceNow/azimuth/pull/587/commits/b44fa1edb840c86cdd36b96edddcff4d4fc09990) (unlerated)
  - This ensures we never forget to define `onChange`.
  - It allows for simplifying the handlers in our custom field components.
  - Only one field is read-only and requires adding a `onChange={() => {}}`. I think it's good that it becomes explicit, showing it's not because we forgot `onChange`.
* [Refactor `KeyValuePairs`](https://github.com/ServiceNow/azimuth/pull/587/commits/6ed04dc4be56ca07f6f5e2b893cd81906cf2aea1)
  - Removing duplicated code like
```tsx
          <Box display="flex" flexDirection="column">
            <Typography variant="caption">...</Typography>
            <KeyValuePairs>
              {... => (
                <React.Fragment key={...}>
                  <Typography variant="body2">{...}:</Typography>
                  ...
                </React.Fragment>
              ))}
            </KeyValuePairs>
          </Box>
```
* [Fix disabled look](https://github.com/ServiceNow/azimuth/pull/587/commits/c5430cb7a314ea972f8f26a2b9d5f20e3af9f98f)

Before:
Screenshot 2023-06-27 at 11 27 23 PM
After:
Screenshot 2023-06-27 at 11 26 33 PM

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -1000,6 +1000,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
           label="artifact_path"
           value={resultingConfig.artifact_path}
           InputProps={{ readOnly: true, disableUnderline: true }}
+          onChange={() => {}}
         />
         <NumberField
           label="batch_size"
--- a/webapp/src/pages/Settings.tsx
+++ b/webapp/src/pages/Settings.tsx
@@ -21,6 +21,7 @@ import {
   inputLabelClasses,
   Menu,
   MenuItem,
+  Theme,
   Typography,
 } from "@mui/material";
 import noData from "assets/void.svg";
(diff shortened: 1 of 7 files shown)


webapp/src/pages/Settings.tsx
```diff
@@ -1000,6 +1000,7 @@ const Settings: React.FC<Props> = ({ open, onClose }) => {
           label="artifact_path"
           value={resultingConfig.artifact_path}
           InputProps={{ readOnly: true, disableUnderline: true }}
+          onChange={() => {}}
```
christyler3030, line 1018: Idea: how about defining the props such that there is an option boolean `readonly` and the if it is set `onChange` is not required. 

This is fine but it would be clearer if it were all made explicit. This change is not required, however. What you have is fine.

christyler3030 (approved):

webapp/src/pages/Settings.tsx
JosephMarinier: Yeah, I considered it. The problem is that the type of `onChange` comes from `FieldProps`, which is used by our 5 field components. If we do something like `{readOnly?: false; onChange: (newValue: T) => void} | {readOnly: true; onChange?: never}`, then all our 5 field components need to support this and do some `if`s to check if `onChange` exists and to add the read-only stuff. Since we have only one read-only field in the whole config (and I don't foresee adding more), I felt like this was not worth the trouble. Does that make sense?

christyler3030: Yeah fine by me. Maybe if we were making a framework or something we would make the effort but probably overkill here.

JosephMarinier: Exactly. And MUI's `TextField` didn't even bother. Thanks for your input. 👍

PR #588, opened 2023-06-28
Add discard buttons to individual config fields

JosephMarinier:
## Description:

Here is a screenshot.
image
And here is another one showing the tooltips. They are shown for demo purposes. They usually only pop up when hovering.
image

## Checklist:

  our users.
* [x] **DEV CHANGES.**

### DIFF
--- a/webapp/src/components/Settings/CustomObjectFields.tsx
+++ b/webapp/src/components/Settings/CustomObjectFields.tsx
@@ -7,13 +7,15 @@ const CustomObjectFields: React.FC<{
   excludeClassName?: boolean;
   disabled: boolean;
   value: CustomObject;
+  originalValue: CustomObject | undefined;
   onChange: (update: Partial<CustomObject>) => void;
-}> = ({ excludeClassName, disabled, value, onChange }) => (
+}> = ({ excludeClassName, disabled, value, originalValue, onChange }) => (
   <>
     {!excludeClassName && (
       <StringField
         label="class_name"
         value={value.class_name}
+        originalValue={originalValue?.class_name}
         disabled={disabled}
         onChange={(class_name) => onChange({ class_name })}
       />
@@ -22,6 +24,7 @@ const CustomObjectFields: React.FC<{
       label="remote"
       nullable
       value={value.remote}
+      originalValue={originalValue?.remote}
       disabled={disabled}
       onChange={(remote) => onChange({ remote })}
     />
@@ -29,12 +32,14 @@ const CustomObjectFields: React.FC<{
       array
       label="args"
       value={value.args}
+      originalValue={originalValue?.args}
       disabled={disabled}
       onChange={(args) => onChange({ args })}
     />
     <JSONField
       label="kwargs"
       value={value.kwargs}
+      originalValue={originalValue?.kwargs}
       disabled={disabled}
       onChange={(kwargs) => onChange({ kwargs })}
     />
--- a/webapp/src/components/Settings/DiscardButton.tsx
+++ b/webapp/src/components/Settings/DiscardButton.tsx
@@ -1,0 +1,19 @@
+import { Replay } from "@mui/icons-material";
+import { InputAdornment, Tooltip, IconButton } from "@mui/material";
+
+export const DiscardButton: React.FC<{
+  title: string;
+  disabled: boolean;
+  onClick: () => void;
+}> = ({ title, ...props }) => (
+  <InputAdornment position="start">
+    <Tooltip
+      placement="bottom-start"
+      title={props.disabled ? "" : `Discard change, reset to ${title}`}
+    >
+      <IconButton color="secondary" sx={{ padding: 0 }} {...props}>
+        <Replay />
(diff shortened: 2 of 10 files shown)


christyler3030 (approved):
Have you considered simply hiding the button if there is nothing to undo rather than disabling it? This would reduce the clutter of the screen.

webapp/src/components/Settings/CustomObjectFields.tsx
```diff
@@ -7,13 +7,15 @@ const CustomObjectFields: React.FC<{
   excludeClassName?: boolean;
   disabled: boolean;
```
JosephMarinier, line 8: @christyler3030
> Have you considered simply hiding the button if there is nothing to undo rather than disabling it?

Great question!

The problem with `hidden` (which `{display: none}`) or simply not rendering the button is that when typing in a field, the text value would jump to the right as the button appears. I didn't like that. Here are two solutions I explored, with their challenges:
1. Placing the reset button at the end of the field. This is great for basic text fields, but it's problematic with all the fields that already have something at the end, like a unit or the number pickers' plus/minus arrows. Then, they jump.
2. Reserve the place, but with `visibility: hidden`. I found that weird, considering the button is at the start of the field. It leaves an awkward gap.
3. 2 but with a twist. I tried to mitigate that by showing the disabled button if you hover the field, but I still think it looked weird.
```ts
[`.${textFieldClasses.root}:not(:hover) .${iconButtonClasses.root}[disabled]`]: {
  visibility: "hidden",
},
```

I stopped there to make a PR quickly and get feedback. I could try to combine ideas 1 and 2. I think the button would then need to be outside of the field (not underlined) so the units and number pickers' arrows are at the end of the underline.

Finally, I saw some (mild) interest in differentiating the fields that didn't change (disabled button) from fields for which I am not tracking changes (no button at all). For example, if you add a new pipeline, I don't have old values to compare with, so there is no discard button at all. I could walk you through some special cases like that, if you are interested.

Any thoughts?

christyler3030: The last point about didn't-change vs. no-original-value is most convincing. As for implementation you could also put it at the end *outside* the field. But this is also fine as is. As always you can iterate based on feedback!

webapp/src/components/Settings/DiscardButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const DiscardButton: React.FC<{
+  title: string;
+  disabled: boolean;
+  onClick: () => void;
+}> = ({ title, ...props }) => (
+  <InputAdornment position="start">
+    <Tooltip
+      placement="bottom-start"
+      title={props.disabled ? "" : `Discard change, reset to ${title}`}
```
JosephMarinier, line 12: I'm wondering if I should add more, like
```suggestion
      title={props.disabled ? "" : `The blue arrow indicates a change. Click to reset to ${title}`}
```

christyler3030: Personally I like the tooltip as is

webapp/src/components/Settings/CustomObjectFields.tsx
JosephMarinier: I remembered something: I think I initially tried placing the button outside at the end of the field, but I thought it was super confusing when long fields like that are next to each other.
Screenshot 2023-06-28 at 1 58 50 PM

christyler3030: Yeah maybe not clear who it belongs to... aight good enough as is. Ship it!

JosephMarinier: It's not too bad with the proper spacing.
image
Here is a the same screenshot as before
image
compared to
image
Maybe the cleanness is worth it. What do you think? @gabegma?

gabegma: Yeah!! I prefer it!! Great job!

christyler3030: Works for me!

christyler3030 (approved):

webapp/src/components/Settings/StringArrayField.tsx
```diff
@@ -42,19 +42,17 @@ const StringArrayField: React.FC<
...
-                  onClick={() => onChange(originalValue)}
-                />
-              )}
-              {params.InputProps.startAdornment}
+              {params.InputProps.endAdornment}
+              {originalValue !== undefined &&
+                JSON.stringify(originalValue) !== JSON.stringify(value) && (
+                  <DiscardButton
+                    title={originalValue.join(", ")}
+                    disabled={disabled}
```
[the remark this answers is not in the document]
JosephMarinier: All inputs, including those reset buttons, are disabled while the config is being applied and we wait for the validation by the back end. This ensures that if there is an error in the back end, the form is enabled again, exactly where you left off, and you can fix the errors.

